PR #1, opened 2024-01-17
ci: Set up CI for the project

denisonbarbosa:
Adds the CI skeleton for the project: Set up issue templates, CODEOWNERS, QA checks, tests and so on.

UDENG-2037


tools/go.mod
```diff
@@ -0,0 +1,189 @@
+module github.com/ubuntu/oidc-broker/tools
+
+go 1.21.5
```
jibel, line 3: shouldn't this be ?

```
go 1.21.0
toolchain go1.21.5 
```

jibel (changes requested):

tools/go.mod
denisonbarbosa: I followed what we [have in authd](https://github.com/ubuntu/authd/blob/main/tools/go.mod), but I can update to this new format.

jibel (approved):
x/crypto dependency and tests will have to  be fixed independently of this PR

PR #2, opened 2024-01-22
fix: Fix some issues with existing code

denisonbarbosa:
Some tests that were not marked to fail were failing, so I took a look to understand why. Details for each fix are explained in the commit messages.

### DIFF
--- a/cmd/oidc-broker/daemon/daemon.go
+++ b/cmd/oidc-broker/daemon/daemon.go
@@ -31,8 +31,8 @@ type App struct {
 
 // only overriable for tests.
 type systemPaths struct {
-	BrokersConf string
-	Cache       string
+	BrokerConf string
+	Cache      string
 }
 
 // daemonConfig defines configuration parameters of the daemon.
@@ -60,8 +60,8 @@ func New(name string) *App {
 			}
 			a.config = daemonConfig{
 				Paths: systemPaths{
-					BrokersConf: filepath.Join(consts.DefaultBrokersConfPath, name),
-					Cache:       systemCache,
+					BrokerConf: filepath.Join(consts.DefaultBrokersConfPath, name),
+					Cache:      systemCache,
 				},
 			}
 
--- a/cmd/oidc-broker/daemon/daemon.go
+++ b/cmd/oidc-broker/daemon/daemon.go
@@ -17,9 +17,6 @@ import (
 	"github.com/ubuntu/oidc-broker/internal/daemon"
 )
 
-// cmdName is the binary name for the agent.
-var cmdName = filepath.Base(os.Args[0])
-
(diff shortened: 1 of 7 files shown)


cmd/oidc-broker/daemon/daemon.go
```diff
@@ -44,31 +45,35 @@ type daemonConfig struct {
 }
 
 // New registers commands and return a new App.
-func New() *App {
-	a := App{ready: make(chan struct{})}
+func New(name string) *App {
+	if name == "" {
```
didrocks, line 49: I would directly change to not make it optional then, and remove the global name altogether (to ensure we don’t use it by accident)

internal/brokerservice/brokerservice.go
```diff
@@ -107,7 +108,12 @@ func (s *Service) Serve() error {
 
 // Stop stop the service and do all the necessary cleanup operation.
 func (s *Service) Stop() error {
-	close(s.serve)
-	s.disconnect()
+	// Check if already stopped.
+	select {
+	case <-s.serve:
+	case <-time.After(10 * time.Millisecond):
+		close(s.serve)
```
didrocks, line 115: unsure how this fixes the race/double close case? was it something we were doing in adsys or authd?

didrocks (changes requested):
One change requested and one question!

internal/brokerservice/brokerservice.go
denisonbarbosa: Receiving from a closed channel is a non-blocking no-op. If the channel is not closed, it will hang there the same way as the `serve()` function does and then we close it after a little delay.

TLDR: If the channel is already closed, Stop() is a no-op.

didrocks: instead of a small delay (which introduce a bigger race window if we don’t use mutex), why not use `default:`?

denisonbarbosa: You are very smart, mister. I didn't think of it at all 🤣

cmd/oidc-broker/main.go
```diff
@@ -20,7 +20,7 @@ import (
 
 func main() {
 	i18n.InitI18nDomain(consts.TEXTDOMAIN, po.Files)
-	a := daemon.New("")
+	a := daemon.New("oidc-broker")
```
didrocks, line 23: No, because we will not be able to rename it. This is why (see the diff, it’s `filepath.Base(os.Args[0])`) and not hardcoded.

denisonbarbosa: oops, you're right! I'll fix it.

didrocks (approved):
All good!

PR #3, opened 2024-01-24
feat(config): ship generic oidc configuration

didrocks:
This configuration should be functional for the default oidc broker. It will then be transformed when doing the snap build.

UDENG-2046


denisonbarbosa (approved):
I'm glad we are finally done with this huge PR! :sunglasses: 
Well done!

didrocks:
I thought I would never be able to see the end if it! :p

PR #21, opened 2024-06-04
build: Add default (generic OIDC) snap package for the broker

denisonbarbosa:
Add snap packaging for the default version of the broker.

UDENG-2044

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -1,0 +1,40 @@
+name: oidc-broker
+summary: OIDC Broker for authd
+description: |
+  Broker that enables OIDC authentication for authd.
+version: git
+grade: stable
+base: core24
+confinement: strict
+license: GPL-3.0+
+
+apps:
+  oidc-broker:
+    command: bin/oidc-broker
+    daemon: simple
+    slots:
+      - dbus-oidc
+    plugs:
+      - network
+      - config-files
+    restart-condition: always
+
+slots:
+  dbus-oidc:
+    interface: dbus
+    bus: system
+    name: com.ubuntu.authd.OidcBroker
+
+plugs:
+  config-files:
+    interface: system-files
+    read:
+      - /etc/authd/brokers.d/oidc-broker
+
+parts:
+  oidc-broker:
+    source: .
+    source-type: local
+    plugin: go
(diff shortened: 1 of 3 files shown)


snap/snapcraft.yaml
```diff
@@ -0,0 +1,41 @@
...
+plugs:
+  config-files:
+    interface: system-files
+    read:
+      - /etc/authd/brokers.d/oidc-broker
+
+parts:
+  oidc-broker:
+    source: https://github.com/ubuntu/oidc-broker
+    source-branch: main
```
didrocks, line 37: can’t that be . to take the current branch? (and potentially, build snaps from other branches?)

snap/snapcraft.yaml
```diff
@@ -0,0 +1,41 @@
...
+    source-type: git
```
didrocks, line 38: ditto here, can’t we just use what’s in the current directory? I can see a world when we want to do a change, then, rebuild the snap and test it from the local modification (which may not be committed) :p

didrocks (changes requested):
A small couple of questions before getting that in! Nice work on the generic broker name :)

snap/snapcraft.yaml
denisonbarbosa: I'll answer both into the same comment, since they are related: yeah, we could leave the source-branch empty or use the local source to build the snap. I didn't do it for two reasons:
1) I think it's good to be as explicit as possible here to avoid confusion or misuse later;
2) When building the snaps through CI, we probably want to use the main branch instead of local files, no? (Especially since we might want to run the build in a separate branch)

snap/snapcraft.yaml
denisonbarbosa: {discussion in the comment above}

snap/snapcraft.yaml
didrocks: well, I see 1. being rather a source of confusion when hacking locally. And having worked as a dev advocate for snaps, I can ensure you this will happen :)

On 2. Let me check, but I think the snapstore is working by targetting a branch and checking it out, let me find back some old snaps config I have to confirm

didrocks (approved):

PR #23, opened 2024-06-07
build: MS EntraID snap packaging

denisonbarbosa:
Add MS EntraID snap packaging for the broker


didrocks (approved):
Nice work! We will refine the version and build override later! :)

PR #30, opened 2024-06-10
fix: Improve messages on auth modes labels and contents

denisonbarbosa:
No more "gimme your password" \o/

UDENG-3114
UDENG-3115


didrocks (approved):
Gimme me a tissue for those tears in my eyes :p

PR #32, opened 2024-06-12
ci: Add action to autorebase Snap branches

denisonbarbosa:
To avoid having to manually rebase every snap branch every time we want a new release, we created this action to do it for us or open a pull request when it can't auto-merge due to conflicts.

UDENG-3171


didrocks (approved):
\o/ that was a good pairing session :)

PR #35, opened 2024-06-18
feat: Prevent empty local password

denisonbarbosa:
Empty strings (i.e. "") still worked as local passwords for the broker users. We now prevent the user from setting an empty password as its local one.

UDENG-3119


didrocks (approved):
Ok, let’s see how this goes. I feel we may have special cases in the future, but let’s not worry about this for now and force this to potentially relax it after :)

PR #36, opened 2024-06-25
broker: Add support for qrcode code field

3v1n0:
Authd supports a code field now, so expose the value as it is if the UI supports it.

Related to https://github.com/ubuntu/authd/pull/387


denisonbarbosa:
I'll close this as this is already integrated into #37.

PR #37, opened 2024-06-26
fix(broker): Improvements to the authentication modes flow

denisonbarbosa:
This contains a couple of adjustments to our AuthenticationModes-related functions: GetAuthenticationModes and SelectAuthenticationMode (very minor). Details are explained in more detail in the commit messages, but the TLDR is:
1- Simplify a bit of the logic in the functions;
2- Add support for the new `code` layout field introduced in [authd#387](https://github.com/ubuntu/authd/pull/387);
3- Use a more comprehensive name for the authentication mode related to the device authentication;
4- Only show device authentication as a valid authentication mode if the provider is reachable and supports this mode;

UDENG-3121


internal/broker/broker_test.go
```diff
@@ -111,15 +111,21 @@ func TestGetAuthenticationModes(t *testing.T) {
...
 		// Auth Session
-		"Get qrcode if there is no token":                      {},
-		"Get newpassword if already authenticated with QRCode": {secondAuthStep: true},
-		"Get password and qrcode if token exists":              {tokenExists: true},
+		"Get device_auth if there is no token":                      {},
+		"Get newpassword if already authenticated with device_auth": {secondAuthStep: true},
+		"Get password and device_auth if token exists":              {tokenExists: true},
+
+		"Get only password if token exists and provider is not available":             {tokenExists: true, providerAddress: "127.0.0.1:31310", unavailableProvider: true},
+		"Get only password if token exists and provider does not support device_auth": {tokenExists: true, providerAddress: "127.0.0.1:31311", deviceAuthUnsupported: true},
```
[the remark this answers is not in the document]
denisonbarbosa: Yeah, I'm not that happy with how it turned out either. I could use `t.Name()` to generate the ports, but it seems very unnecessary and too much work for little improvement.

internal/broker/broker.go
```diff
@@ -204,10 +204,24 @@ func (b *Broker) GetAuthenticationModes(sessionID string, supportedUILayouts []m
 	_, err = os.Stat(session.cachePath)
 	tokenExists := err == nil
 
+	// Checks if the provider is accessible and if device authentication is supported
+	providerReachable := true
+	r, err := http.Get(strings.TrimSuffix(b.auth.providerURL, "/") + "/.well-known/openid-configuration")
```
didrocks, line 209: double // doesn’t work or this is just for principles? :)

internal/broker/broker.go
```diff
@@ -302,13 +316,13 @@ func (b *Broker) generateUILayout(session *sessionInfo, authModeID string) (map[
...
 			"label": fmt.Sprintf(
-				"Scan the QR code or access %q and use the code %q",
+				"Scan the QR code or access %q and use the provided code",
 				response.VerificationURI,
-				response.UserCode,
 			),
 			"wait":    "true",
 			"button":  "regenerate QR code",
 			"content": response.VerificationURI,
+			"code":    response.UserCode,
```
3v1n0, line 325: This good, but if you test this with the current version of gdm where the `code` is not yet supported (I mean the version already in the repository, since the code for supporting it won't land very early in distro and may not be fully synchronized with authd), will it work anyways?

As in the other PR I supported both cases, so both if the UI supports it or not, but I'm not sure if that would fail anywhere. I guess no, since it's optional, but asking for confirmation here.

didrocks (approved):
Nothing is burning my eyes apart from my small comment. It reads better. If you are happy with it, feel free to merge! Thanks :)

internal/broker/broker.go
denisonbarbosa: It will work, but the code won't be shown if GDM doesn't support it. The thing is: there's no way to know this on the broker side since what matters here is the supported layouts that authd is sending us (and right now it is sending code as optional). I can attach the code back to the label if you think it's better.

internal/broker/broker.go
denisonbarbosa: To be honest, I just used the same code that the `oidc` package uses :joy: but I'm pretty sure that double slashes invalidate the URI.

PR #41, opened 2024-06-28
broker: Use native JSON marshalling to handle messages and user infos

3v1n0:
During the test phase we got errors messages whose encoding wasn't correct JSON, causing unexpected errors.

To prevent this, just use native go JSON encoding to prevent us to have to manage special cases that may break the assumptions, and be always safer.

UDENG-3428

### DIFF
--- a/internal/broker/broker.go
+++ b/internal/broker/broker.go
@@ -17,7 +17,7 @@ import (
 	"slices"
 	"strings"
 	"sync"
-	"text/template"
+	"testing"
 	"time"
 
 	"github.com/coreos/go-oidc/v3/oidc"
@@ -368,22 +368,28 @@ func (b *Broker) IsAuthenticated(sessionID, authenticationData string) (string, 
 	authDone := make(chan struct{})
 	var access, data string
 	go func() {
-		access, data = b.handleIsAuthenticated(ctx, &session, authData)
+		access, data = b.handleIsAuthenticated(ctx, &session, authData).result()
 		close(authDone)
 	}()
 
 	select {
 	case <-authDone:
 	case <-ctx.Done():
-		return AuthCancelled, `{"message": "authentication request cancelled"}`, ctx.Err()
+		access, data := isAuthenticatedResult{
+			access:  AuthCancelled,
+			Message: "authentication request cancelled",
+		}.result()
+		return access, data, ctx.Err()
 	}
 
 	switch access {
 	case AuthRetry:
 		session.attemptsPerMode[session.selectedMode]++
 		if session.attemptsPerMode[session.selectedMode] == maxAuthAttempts {
-			access = AuthDenied
-			data = `{"message": "maximum number of attempts reached"}`
+			access, data = isAuthenticatedResult{
+				access:  AuthDenied,
+				Message: "maximum number of attempts reached",
+			}.result()
 		}
 
 	case AuthNext:
@@ -396,11 +402,49 @@ func (b *Broker) IsAuthenticated(sessionID, authenticationData string) (string, 
 	return access, data, nil
 }
 
-func (b *Broker) handleIsAuthenticated(ctx context.Context, session *sessionInfo, authData map[string]string) (access, data string) {
+type isAuthenticatedResult struct {
+	access   string          `json:"-"`
+	UserInfo json.RawMessage `json:"userinfo,omitempty"`
+	Message  string          `json:"message,omitempty"`
+}
+
+func (iar *isAuthenticatedResult) encode() (string, error) {
+	if iar.UserInfo == nil && iar.Message == "" {
+		return "", nil
+	}
+	b, err := json.Marshal(iar)
+	if err != nil {
+		return "", err
+	}
+	if testing.Testing() {
+		// Indent in testing, for improve readability of golden files
+		var indented bytes.Buffer
+		err = json.Indent(&indented, b, "", "  ")
+		if err != nil {
+			return "", err
+		}
+		b = indented.Bytes()
+	}
+	return string(b), err
+}
+
+func (iar isAuthenticatedResult) result() (string, string) {
+	ret, err := iar.encode()
+	if err != nil {
+		slog.Debug(fmt.Sprintf("Error when encoding %#v: %v", iar, err))
+		return iar.access, ""
+	}
+	return iar.access, ret
+}
+
+func (b *Broker) handleIsAuthenticated(ctx context.Context, session *sessionInfo, authData map[string]string) isAuthenticatedResult {
 	// Decrypt challenge if present.
 	challenge, err := decodeRawChallenge(b.privateKey, authData["challenge"])
 	if err != nil {
-		return AuthRetry, fmt.Sprintf(`{"message": "could not decode challenge: %v"}`, err)
+		return isAuthenticatedResult{
+			access:  AuthRetry,
+			Message: fmt.Sprintf("could not decode challenge: %v", err),
+		}
 	}
 
 	var authInfo authCachedInfo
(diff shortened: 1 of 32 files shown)


internal/broker/broker.go
```diff
@@ -669,38 +749,22 @@ type claims struct {
...
-		"dir": "{{.Home}}",
-		"shell": "{{.Shell}}",
-		"groups": [ {{range $index, $g := .Groups}}
-			{{- if $index}}, {{end -}}
-			{"name": "{{.Name}}", "ugid": "{{.UGID}}"}
-		{{- end}} ]
-}`)).Execute(&buf, user)
-	if err != nil {
-		return "", err
-	}
```
denisonbarbosa, line 703: IIRC, the templating was necessary due to the way `json.Marshal` formats go slices, but I don't recall exactly the reason why. Can you give some more input here, @didrocks?

internal/broker/export_test.go
```diff
@@ -105,5 +107,12 @@ func (b *Broker) FetchUserInfo(sessionID string, cachedInfo *authCachedInfo) (st
...
 	}
 
-	return b.userInfoFromClaims(uInfo, groups)
+	userInfo := b.userInfoFromClaims(uInfo, groups)
+	data, err := json.Marshal(userInfo)
+	if err != nil {
+		return "", err
+	}
+	var indented bytes.Buffer
+	err = json.Indent(&indented, data, "", "\t\t")
```
denisonbarbosa, line 116: Do we really need two tabs?

internal/broker/broker.go
```diff
@@ -396,11 +402,49 @@ func (b *Broker) IsAuthenticated(sessionID, authenticationData string) (string,
...
+}
+
+func (iar isAuthenticatedResult) result() (string, string) {
+	ret, err := iar.encode()
+	if err != nil {
+		slog.Debug(fmt.Sprintf("Error when encoding %#v: %v", iar, err))
+		return iar.access, ""
+	}
+	return iar.access, ret
+}
```
denisonbarbosa, line 438: This wrapping is not necessary. Since the `access` and `data` only matter when going out of `IsAuthenticated`, I would replace the places where we need to unwrap this to use `isAuthenticatedResult` instead and then handle this logic at the end of the `IsAuthenticated` function (especially since, if we fail to encode the data, we need to return either `Retry` or `Denied`).

internal/broker/broker.go
```diff
@@ -368,22 +368,28 @@ func (b *Broker) IsAuthenticated(sessionID, authenticationData string) (string,
 	authDone := make(chan struct{})
 	var access, data string
 	go func() {
-		access, data = b.handleIsAuthenticated(ctx, &session, authData)
+		access, data = b.handleIsAuthenticated(ctx, &session, authData).result()
```
denisonbarbosa, line 371: Following my comment on the wrapping, here would become something like: 
`result := b.handleIsAuthenticated(ctx, &session, authData)`

denisonbarbosa (changes requested):
Looks good, but I do have some suggestions/requests :wink:

internal/broker/broker.go
```diff
@@ -17,7 +17,7 @@ import (
 	"slices"
 	"strings"
 	"sync"
-	"text/template"
+	"testing"
```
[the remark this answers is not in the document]
3v1n0: mh I guessed that, but I feel that's why `testing.Testing()` existed no?

But maybe I can just replicate the same using a different implementation of the same function both in a test-code and no-test code I guess

internal/broker/broker.go
```diff
@@ -396,11 +402,49 @@ func (b *Broker) IsAuthenticated(sessionID, authenticationData string) (string,
...
+	}
+	if testing.Testing() {
+		// Indent in testing, for improve readability of golden files
+		var indented bytes.Buffer
+		err = json.Indent(&indented, b, "", "  ")
+		if err != nil {
+			return "", err
+		}
+		b = indented.Bytes()
+	}
```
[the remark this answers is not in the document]
3v1n0: More than testability, is also making it easier for diffing, the more the lines, the better.

And as you may see in this PR the no-content-changes to the golden files are appreciable and easy to notice, so I guess that having that in also future PRs can be convenient. So maybe we can do something similar to what we did with `validateJSONFunc` in authd?

internal/broker/export_test.go
3v1n0: No, I did it to be consistent with what we had in the current golden files, so to reduce the diffing for making the review easier.

internal/broker/broker.go
3v1n0: It was  not to repeat the same in 3 places, but ok...

internal/broker/broker.go
```diff
@@ -366,24 +364,27 @@ func (b *Broker) IsAuthenticated(sessionID, authenticationData string) (string,
...
+		result = b.handleIsAuthenticated(ctx, &session, authData)
 		close(authDone)
 	}()
 
 	select {
 	case <-authDone:
 	case <-ctx.Done():
-		return AuthCancelled, `{"message": "authentication request cancelled"}`, ctx.Err()
+		result := isAuthenticatedResult{Message: "authentication request cancelled"}
+		data, _ := result.toJSON()
```
denisonbarbosa, line 377: I know the message is just above, but it's a good practice to add a comment explaining why we are ignoring the error.

internal/broker/broker.go
```diff
@@ -366,24 +364,27 @@ func (b *Broker) IsAuthenticated(sessionID, authenticationData string) (string,
...
-	switch access {
+	switch result.access {
 	case AuthRetry:
 		session.attemptsPerMode[session.selectedMode]++
 		if session.attemptsPerMode[session.selectedMode] == maxAuthAttempts {
-			access = AuthDenied
-			data = `{"message": "maximum number of attempts reached"}`
+			result := isAuthenticatedResult{Message: "maximum number of attempts reached"}
+			data, _ := result.toJSON()
+			return data, AuthDenied, nil
```
denisonbarbosa, line 387: Ditto

internal/broker/broker.go
```diff
@@ -393,14 +394,42 @@ func (b *Broker) IsAuthenticated(sessionID, authenticationData string) (string,
...
+	data, err := result.toJSON()
+	if err == nil {
+		return result.access, data, nil
+	}
+
+	slog.Debug(fmt.Sprintf("Impossible to convert to JSON %#v: %v", result, err))
+	if result.access == AuthDenied {
+		return result.access, data, nil
+	}
+	return AuthRetry, `{"message": "authentication data encoding failure"}`, nil
```
denisonbarbosa, line 407: Remember the happy path concept...

internal/broker/broker.go
```diff
@@ -366,24 +364,27 @@ func (b *Broker) IsAuthenticated(sessionID, authenticationData string) (string,
...
 		close(authDone)
 	}()
 
 	select {
 	case <-authDone:
 	case <-ctx.Done():
-		return AuthCancelled, `{"message": "authentication request cancelled"}`, ctx.Err()
+		result := isAuthenticatedResult{Message: "authentication request cancelled"}
+		data, _ := result.toJSON()
+		return data, AuthCancelled, ctx.Err()
```
denisonbarbosa, line 378: Returned values are out of order...

internal/broker/export_test.go
```diff
@@ -105,5 +107,25 @@ func (b *Broker) FetchUserInfo(sessionID string, cachedInfo *authCachedInfo) (st
...
+	defaultStringifyJSON := stringifyJSON
+	stringifyJSON = func(b []byte, err error) (string, error) {
+		if err != nil {
+			return defaultStringifyJSON(b, err)
+		}
+		// Indent in testing, to improve readability of golden files
+		var indented bytes.Buffer
+		err = json.Indent(&indented, b, "", "  ")
+		return defaultStringifyJSON(indented.Bytes(), err)
+	}
```
denisonbarbosa, line 130: I'm sorry, @3v1n0, but I still don't like doing this just for readability. I think our golden files should match the actual output of the function as close as we can.

internal/broker/broker.go
3v1n0: I did, the reason why I went this way, it was to avoid nesting two times later, as initially I definitely errored earlier, but since that included further checks inside an if I went this way.

internal/broker/broker.go
3v1n0: Uuups, yeah... Reason why I was using a single function, to avoid this kind of errors that are quite easy to get since the current code, that is not strictly typed using strings, is fragile to this kind of issues

internal/broker/export_test.go
3v1n0: Ok, we can also use this in production code, but if you don't care about readability (I did it only because I thought it was a project goal, and to help you and others reviewing the PR), I'll just drop this.

internal/broker/broker.go
didrocks: The issue was with 0 length or nil slices IIRC, but to be confirmed.

internal/broker/broker.go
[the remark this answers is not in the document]
didrocks: The import itself is not an issue, but I don’t want too that we have test specific code in production code. The test specific code is only to add helpers changing interfaces or values that are not exported and they should be really limited.

If you import testing for anything else, it means that the abstraction is wrong.

internal/broker/testdata/TestIsAuthenticated/golden/authenticating_with_password_refreshes_expired_token/first_call
```diff
@@ -1,11 +1,3 @@
...
-data: |-
-    {"userinfo": {
-    		"name": "test-user@email.com",
-    		"uuid": "test-user-id",
-    		"gecos": "test-user",
-    		"dir": "/home/test-user@email.com",
-    		"shell": "/usr/bin/bash",
-    		"groups": [ {"name": "remote-group", "ugid": "12345"}, {"name": "linux-local-group", "ugid": ""} ]
-    }}
+data: '{"userinfo":{"name":"test-user@email.com","uuid":"test-user-id","dir":"/home/test-user@email.com","shell":"/usr/bin/bash","gecos":"test-user","groups":[{"name":"remote-group","ugid":"12345"},{"name":"linux-local-group","ugid":""}]}}'
```
didrocks, line 2: I think you get a better formatting when generating the golden files and formatting the data there to have ident, and easier to read, as it was before.

internal/broker/export_test.go
```diff
@@ -105,5 +108,26 @@ func (b *Broker) FetchUserInfo(sessionID string, cachedInfo *authCachedInfo) (st
...
+	data, err := json.Marshal(userInfo)
+	if err != nil {
+		return "", err
+	}
+	var indented bytes.Buffer
+	err = json.Indent(&indented, data, "", "\t\t")
+	return indented.String(), err
+}
+
+var ErrTestChallengeDecodeError error
```
didrocks, line 121: I don’t think the approach is correct that way, Only integration tests should change the default, not for default tests where it should be some tests parameters.

internal/broker/broker.go
```diff
@@ -412,67 +442,103 @@ func (b *Broker) handleIsAuthenticated(ctx context.Context, session *sessionInfo
...
+			return isAuthenticatedResult{
+				access:  AuthRetry,
+				Message: fmt.Sprintf("could not authenticate user: %v", err),
+			}
 		}
 
 		rawIDToken, ok := t.Extra("id_token").(string)
 		if !ok {
-			return AuthDenied, `{"message": "could not get id_token"}`
+			return isAuthenticatedResult{
```
didrocks, line 461: This is not C or other languages. Go allows multiple returned values, with last one being an error by convention. Don’t create and wrap this to return a simple struct.

internal/broker/broker.go
```diff
@@ -393,14 +396,41 @@ func (b *Broker) IsAuthenticated(sessionID, authenticationData string) (string,
...
+}
+
+type isAuthenticatedResult struct {
+	access   string          `json:"-"`
+	UserInfo json.RawMessage `json:"userinfo,omitempty"`
+	Message  string          `json:"message,omitempty"`
 }
 
-func (b *Broker) handleIsAuthenticated(ctx context.Context, session *sessionInfo, authData map[string]string) (access, data string) {
+func (iar *isAuthenticatedResult) toJSON() (string, error) {
```
didrocks, line 418: You should implement `MarshalJSON` interface rather so that you don’t have a call an annex function when you want to marshall and this is more idiomatic.

didrocks (changes requested):
To me, this PR should be reworked from the ground up. We keep Go idiomatic code outside of `pam/`. I don’t want this to change and this PR adds pattern I don’t want to maintain long term.

I agree we should use the native marshalling interface, but not at any cost. Also, returning and reassembling this to json should be at the edge (we don’t create subtype randomly), not in the middle of the logic.

I’m happy to work on it if you don’t mind. That way, you can focus on the other kinds of bugs.

internal/broker/export_test.go
```diff
@@ -105,5 +108,26 @@ func (b *Broker) FetchUserInfo(sessionID string, cachedInfo *authCachedInfo) (st
...
+
+func init() {
+	defaultDecodeRawChallenge := decodeRawChallenge
+	decodeRawChallenge = func(priv *rsa.PrivateKey, rawChallenge string) (decoded string, err error) {
```
didrocks, line 125: `decodeRawChallenge` is a function attached to the broker struct with a default, and ways to override it with `WithDecodeRawChallengeFunc()`.

That way, we keep parallelized tests and don’t change global state.

internal/broker/broker.go
```diff
@@ -393,14 +396,41 @@ func (b *Broker) IsAuthenticated(sessionID, authenticationData string) (string,
...
+	return result.access, data, nil
+}
+
+type isAuthenticatedResult struct {
+	access   string          `json:"-"`
+	UserInfo json.RawMessage `json:"userinfo,omitempty"`
+	Message  string          `json:"message,omitempty"`
 }
 
-func (b *Broker) handleIsAuthenticated(ctx context.Context, session *sessionInfo, authData map[string]string) (access, data string) {
```
didrocks, line 399: So, I think here, we want to change data itself to a struct that we can marshall to json and keep the 2 returned values.

didrocks:
After a second look, I think we can rework that PR to get it closer to the original go idiomatic intent. Maybe this is the opportunity for a good pairing session to fix those issues?

internal/broker/testdata/TestIsAuthenticated/golden/authenticating_with_password_refreshes_expired_token/first_call
3v1n0: Indeed I do agree, in fact initially the code was changed to use indented results in tests, but I was told in another comment that it [wasn't needed](https://github.com/ubuntu/oidc-broker/pull/41#discussion_r1659300492). While imho having the results indented is quite important for helping reviewers and spot changes.

3v1n0:
I'm closing this since it's going to be handled differently.

PR #42, opened 2024-07-01
fix(broker): Improve newpassword label for passwd sessions

denisonbarbosa:
We used to show the same label for both sessions types (auth and passwd), but this could create confusion for users when, in passwd sessions, they were asked to "Create local password" instead of changing it. This commit should make things clearer and easier to navigate.

UDENG-3423


didrocks (approved):
Easy one but nice improvements for our users :)

PR #43, opened 2024-07-01
fix/refactor(broker): Fix JSON generation of returned authentication data (userinfo and message)

denisonbarbosa:
Some errors were causing our previous strategy of generating the JSON on the go to fail, as some error messages could have double quotes. To avoid this, it's better to use the native marshaling of strings that the `encode/json` package provides.

UDENG-3428


internal/broker/broker.go
```diff
@@ -393,86 +393,83 @@ func (b *Broker) IsAuthenticated(sessionID, authenticationData string) (string,
 	if err = b.updateSession(sessionID, session); err != nil {
 		return AuthDenied, "", err
 	}
-	return access, data, nil
+
+	encoded, err := data.MarshalJSON()
```
denisonbarbosa, line 397: Since the type is unknown here, we can't use `json.Marshal` directly otherwise it would defer to reflection + recursion for each of the structs fields and override our current implementation for them.

denisonbarbosa: Now that I reviewed myself, I'll also address my own comment and add it as a comment to the code :joy:

internal/broker/broker.go
```diff
@@ -380,15 +378,17 @@ func (b *Broker) IsAuthenticated(sessionID, authenticationData string) (string,
 	select {
 	case <-authDone:
 	case <-ctx.Done():
-		return AuthCancelled, `{"message": "authentication request cancelled"}`, ctx.Err()
+		// We can ignore the error here since the message is constant.
```
didrocks, line 381: proud of you :)

internal/broker/broker.go
didrocks: Hum, I’m unsure to understand you here.

I was going to suggest `json.Marshal` here too, because it will, for each type, assert if you have the interface and call it, as in https://pkg.go.dev/encoding/json#Marshal: "Marshal traverses the value v recursively. If an encountered value implements [Marshaler](https://pkg.go.dev/encoding/json#Marshaler) and is not a nil pointer, Marshal calls [Marshaler.MarshalJSON] to produce JSON"

So, from what I gues `json.Marshall(data)` should then call `data.MarshalJSON` under the hood?

internal/broker/broker.go
```diff
@@ -398,86 +398,85 @@ func (b *Broker) IsAuthenticated(sessionID, authenticationData string) (string,
 	if err = b.updateSession(sessionID, session); err != nil {
 		return AuthDenied, "", err
 	}
-	return access, data, nil
+
+	// The data type is unknown, so we need to directly call its marshal function rather than relying
+	// on json.Marshal, which would reflect the struct fields and use what is implemented for the basic types.
+	encoded, err := data.MarshalJSON()
+	if err != nil {
+		return AuthDenied, "", fmt.Errorf("could not marshal data: %v", err)
```
didrocks, line 406: marshal is a little bit generic, I would that could not convert data to json or something like this?

internal/broker/data.go
```diff
@@ -0,0 +1,58 @@
+package broker
```
didrocks, line 1: Can you rename the file to something less generic than `data`? this file seems to only deal with userinfo and errors, so maybe `authresponse.go`?

internal/broker/broker.go
```diff
@@ -564,16 +563,14 @@ func (b *Broker) updateSession(sessionID string, session sessionInfo) error {
 // authCachedInfo represents the token that will be saved on disk for offline authentication.
 type authCachedInfo struct {
 	Token      *oauth2.Token
-	AcquiredAt time.Time
```
didrocks, line 567: that wasn’t used anymore and you hide some cleanup here? tsssss tsssss :p

internal/broker/data.go
```diff
@@ -0,0 +1,58 @@
...
+	"bytes"
+	"encoding/json"
+	"text/template"
+
+	"github.com/ubuntu/oidc-broker/internal/providers/group"
+)
+
+// userInfo represents the user information that is returned to authd.
+type userInfo struct {
+	Name   string
```
didrocks, line 13: As we don’t use json reflection here and do serialization ourself, there is no need for those fields to be public, isn’t it?

internal/broker/data.go
```diff
@@ -0,0 +1,58 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return buf.Bytes(), nil
+}
+
+// errorMessage represents the error message that is returned to authd.
+type errorMessage struct {
+	Message string
```
didrocks, line 44: ditto on public field!

internal/broker/data.go
```diff
@@ -0,0 +1,58 @@
...
+func (msg errorMessage) MarshalJSON() ([]byte, error) {
+	if msg.Message == "" {
+		return nil, nil
+	}
+
+	b, err := json.Marshal(msg.Message)
+	if err != nil {
+		return nil, err
+	}
+	return []byte(`{"message": ` + string(b) + `}`), nil
```
didrocks, line 57: nitpick: I would still use `fmt.Sprintf()` as explained yesterday for readability

internal/broker/data.go
didrocks: I see you are using the private fields in tests, but let’s not tamper production code for tests. In that case, it’s better to transcribed to another struct which has the mapping with public fields in `export_test.go` or just dump it to json in golden IMHO

internal/broker/testdata/TestFetchUserInfo/golden/successfully_fetch_user_info_with_default_home_when_not_provided
```diff
@@ -1,8 +1,10 @@
...
+name: test-user@email.com
+uuid: saved-user-id
+home: /home/test-user@email.com
+shell: /usr/bin/bash
+gecos: saved-user
+groups:
+    - name: remote-group
+      ugid: "12345"
+    - name: linux-local-group
+      ugid: ""
```
didrocks, line 10: True that it’s more readable as structs here.

internal/broker/testdata/TestIsAuthenticated/golden/authenticating_with_password_still_allowed_if_token_is_expired_and_server_is_unreachable/first_call
```diff
@@ -6,6 +6,6 @@ data: |-
     		"gecos": "saved-user",
     		"dir": "/home/test-user@email.com",
     		"shell": "/usr/bin/bash",
-    		"groups": [{"name": "saved-remote-group", "gid": "12345"}, {"name": "saved-local-group", "gid": ""}]
+    		"groups": [ {"name": "remote-group", "ugid": "12345"}, {"name": "linux-local-group", "ugid": ""} ]
```
didrocks, line 9: hum, is the order swap expected? (groups is still a slice, no? So the order is deterministic supposively?)

didrocks (changes requested):
Nice work!

I think this is the right approach with the existing code. Thanks for tackling this!
As you saw, I have some improvements requests in the PR, but nothing changing the structure of the overall PR.

Also, shouldn’t we add a test where the error message was not producing a valid json previously on purpose? Because this is what is supposively fixed in the PR :D

internal/broker/data.go
denisonbarbosa: Here you are completely ignoring the profiling talks from the sprint...

internal/broker/broker.go
denisonbarbosa: Well, it doesn't seem to apply to the actual full type. Here, when calling `json.Marshal(data)` it would just do the default struct marshaling instead (very weird since both types implement the `json.Marshaler` interface, but that's what happens), so my guess is that it can't assert the type of data by itself.

internal/broker/data.go
3v1n0: Not sure we are in a performance-sensible place here :)

internal/broker/broker.go
denisonbarbosa: This never happened, you never saw it, it was never here... 
[Jedi mind erasing movements]

internal/broker/data.go
didrocks: Maintainance > profiling :)

internal/broker/data.go
```diff
@@ -0,0 +1,58 @@
...
+			{{- if $index}}, {{end -}}
+			{"name": "{{.Name}}", "ugid": "{{.UGID}}"}
+		{{- end}} ]
+}}`)).Execute(&buf, u)
+	if err != nil {
+		return nil, err
+	}
+
+	return buf.Bytes(), nil
+}
```
3v1n0, line 40: To understand, what is the rationale for having this (which is is still technically still potentially affected by the root issue), compared to native marshalling?

3v1n0:
I think it would be worth adding a test to ensure we properly do the encoding (something [maybe inspired to this](https://github.com/ubuntu/authd-oidc-brokers/pull/41/commits/ce6875e3edbd96924b6f7cfb73e3a9fcc8d1c5d1), but done with the proper abstraction which at the time I didn't do since it was mostly a PoC).

Indeed `internal/broker/testdata/TestIsAuthenticated/golden/error_when_mode_is_password_and_can_not_fetch_user_info/first_call` does already some check, but it's not very close to the issue we had in the past IMHO.

Maybe also making the tests to-remarshall the output to a native go map, just to ensure that's valid?

internal/broker/data.go
denisonbarbosa: Well, the struct itself is already private. IMHO, changing its fields to private creates more nuisance than it helps prevent, no? I was trying to find some guidelines for idiomatic struct declarations and so on, but did not find any. Do you have some links?

Also, we do unmarshal this type quite a lot inside our code (when loading the cached info for the user and so on), so I'm not 100% sure we can make these private anyway.

internal/broker/data.go
3v1n0: To be fair, I don't like too much either the fact we're hardcoding `{"message": ...}` as it is.

As I've a strong allergy with hardcoded code anyways and especially strings, so I'd rather like it more to have `errorMessage` to be fully marshallable. Not sure what was the problem with it though.

internal/broker/data.go
denisonbarbosa: Honestly, I don't like it either. There was a problem with it, but I think it used to happen before we defined a type for the user groups. I'm down to strip it and run some checks to see if everything still works as expected.

internal/broker/data.go
denisonbarbosa: The issue here is the `json.Marshaler` implementation. We can't call `json.Marshal` directly on the `errorMessage` (otherwise it would recurse infinitely). I feel like the best way here (especially taking the other comment about the userInfo marshaling into consideration), is to define our own interface (something like `dataMarshaler`) to avoid this kind of recursion. Wdyt?

3v1n0: Can't we use `json.RawMessage` to stop the recursion?

internal/broker/broker.go
```diff
@@ -369,9 +369,9 @@ func (b *Broker) IsAuthenticated(sessionID, authenticationData string) (string,
 
 	authDone := make(chan struct{})
 	var access string
-	var data json.Marshaler
+	var iaResponse isAuthenticatedDataResponse
```
3v1n0, line 372: nit: `iadResponse` ?

internal/broker/broker.go
```diff
@@ -399,16 +399,19 @@ func (b *Broker) IsAuthenticated(sessionID, authenticationData string) (string,
...
+	encoded, err := json.Marshal(iaResponse)
 	if err != nil {
 		return AuthDenied, "", fmt.Errorf("could not marshal data: %v", err)
 	}
-	return access, string(encoded), nil
+
+	data := string(encoded)
+	if data == "null" {
+		data = ""
+	}
```
3v1n0, line 410: Can't we prevent this earlier? Like checking zero'd `encoded` or implementing a special encoder?

internal/broker/testdata/TestIsAuthenticated/golden/authenticating_with_password_refreshes_expired_token/first_call
3v1n0, line 1: You know I'd like these to be easily parsed by using indentation, but if you can parse the potential future diffs here (I can't easily) well, I'm fine.

internal/broker/broker_test.go
3v1n0, line 1: Since you're here, can you also check that the `data` returned by `access, data, err := b.IsAuthenticated(sessionID, authData)` can be marshalled to a native object again?

As it would ensure that what we return can be parsed by the caller no? Or is there already something that does it?

3v1n0 (commented):
Indeed nicer now, thanks.

Some further nits :)

internal/broker/broker_test.go
denisonbarbosa: It's already done multiple times in the broker flow, by marshaling and unmarshaling the `userinfo` from the cache, so we are safe here, IMHO.

internal/broker/testdata/TestIsAuthenticated/golden/authenticating_with_password_refreshes_expired_token/first_call
denisonbarbosa: GitHub is a bit annoying in the diff department, I agree. However, it's very easy to look at it locally by using `git diff --word-dif` (which is also the default diff also shown by VSCode, IIRC).

internal/broker/broker.go
denisonbarbosa: Didier and I thought about it, but it's not worth implementing an `emptyMessage` type as this only happens on a single occasion, IIRC.  As this is a very simple check, I guess it's okay as it is.

internal/broker/broker_test.go
```diff
@@ -365,6 +367,7 @@ func TestIsAuthenticated(t *testing.T) {
 				"token": (&oauth2.Token{}).WithExtra(map[string]interface{}{"id_token": "invalid"}),
 			},
 		},
+		"Error when selected username does not match the provider one": {username: "not-matching", firstChallenge: "-", wantSecondCall: true},
```
didrocks, line 370: I won’t describe the arguments here, I would rather say "Error message returns valid json when error contains quotes" or something like that. As this is really what we want to test here and won’t remember this is the case in the future.

The issue is that the previous code could make the test pass too:
```            	            	data: '{"message": "could not get user info: could not fetch user info: returned user "test-user@email.com" does not match the selected one "not-matching""}'
```

I suggest we had, before creating the golden file, some validation in the tests that this is valid json. That way, the old code will really "fail" (and we don’t count on the developer to notice that the golden file can be or can’t be marshall to json, which is error-prone)

internal/broker/data.go
didrocks: As told, this was relevant when we used templating, but not anymore with the new approach.

didrocks (changes requested):
It’s readable and easier to parse now!

So, I only have one remaining requests to ensure that we don’t end up with having invalid json sent back to the caller without noticing on the testing side, otherwise good!

internal/broker/broker_test.go
denisonbarbosa: I don't like the idea of having a test case that doesn't exactly match the function contract, but I agree with having JSON validation in the tests. I'll add checks for it on the returned values, wdyt?

denisonbarbosa: Pushed a commit that will fail the tests so you can see the outcome better: https://github.com/ubuntu/authd-oidc-brokers/pull/43/commits/d54e61d267338e124e8a6e1f14950654f0118ce4

3v1n0 (approved):
Nice to see validation was actually needed :)

internal/broker/testdata/TestIsAuthenticated/golden/authenticating_with_password_refreshes_expired_token/first_call
3v1n0: Sure, I know about word diffing, but I feel still it can be more readable and easier to catch when on different lines, since when adding/removing fields order can be hard to catch.

But not blocking on this if Didier and you are fine with parsing those :)

denisonbarbosa:
> Nice to see validation was actually needed :)

Yeah, it wouldn't break anything (as we don't care about the message in the cases that the validation was failing, and unmarshaling an empty string is also fine). However, it's still good to keep things as standardized as possible and ensure we always have a valid JSON-encoded message in non-error cases.

internal/broker/broker.go
```diff
@@ -398,86 +398,88 @@ func (b *Broker) IsAuthenticated(sessionID, authenticationData string) (string,
...
 	}
+
+	encoded, err := json.Marshal(iaResponse)
+	if err != nil {
+		return AuthDenied, "", fmt.Errorf("could not parse data to JSON: %v", err)
+	}
+
+	data := string(encoded)
+	if data == "null" {
+		data = `""`
```
3v1n0, line 409: **One thing, in authd's `brokers/broker.go`  we have:

```go
if data == "" {
		data = "{}"
	}
```

Isn't this breaking such assumption? And we may return a string instead of an object?**

denisonbarbosa: Good point, actually. I think it's then better to keep it as an empty string then and only run `json.Valid` in cases where `data != ""`, wdyt? 

I was thinking about formatting all the empty strings as well, but it's not very idiomatic Go to not return empty values when you have an error.

3v1n0: Imho we should just rerun an empty object and that's it... So it's valid and we don't have to use the brokers check. As ideally it should assume that the returned json is valid

didrocks (approved):
QA failed due Tuesday go release vulnerability notice. Thanks for doing the change!

3v1n0:
All good here after the final commit :)

PR #44, opened 2024-07-02
Align project names

didrocks:
To align all namings, we rename:
- the project itself, contaings code for multiple authd brokers to authd-oidc-brokers.
- oidc-broker implementation to authd-oidc.

This way, all projects relates to authd. We rename the snap names, URLs, imports, configuration files to prevent stuttering.
Remove the config static file.
Update gitignore.

UDENG-3431

Once this is merged, we will need to reshape the msentraid branch too to follow that convention.


jibel (changes requested):

jibel (approved):
Great renaming! Thanks!

PR #47, opened 2024-07-02
broker: Use better wording for the qrcode button

3v1n0:
The button is selecting the authentication again, requesting a new login code, not the qrcode itself.

So improve wording here a bit

UDENG-3416


didrocks (approved):
That’s a better wording

PR #49, opened 2024-07-03
feat: Implement UserPreCheck

denisonbarbosa:
This adds the implementation for the UserPreCheck method on the broker. Without it, users would never be able to authenticate through SSH without first authenticating locally in the machine, which is not ideal.

UDENG-3415

### DIFF
--- a/internal/broker/broker.go
+++ b/internal/broker/broker.go
@@ -30,17 +30,19 @@ const maxAuthAttempts = 3
 
 // Config is the configuration for the broker.
 type Config struct {
-	IssuerURL   string
-	ClientID    string
-	CachePath   string
-	HomeBaseDir string
+	IssuerURL          string
+	ClientID           string
+	CachePath          string
+	HomeBaseDir        string
+	AllowedSSHSuffixes []string
 }
 
 // Broker is the real implementation of the broker to track sessions and process oidc calls.
 type Broker struct {
-	providerInfo providers.ProviderInfoer
-	auth         authConfig
-	homeDirPath  string
+	providerInfo       providers.ProviderInfoer
+	auth               authConfig
+	homeDirPath        string
+	allowedSSHSuffixes []string
 
 	currentSessions   map[string]sessionInfo
 	currentSessionsMu sync.RWMutex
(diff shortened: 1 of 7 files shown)


internal/broker/broker.go
```diff
@@ -540,6 +543,16 @@ func (b *Broker) CancelIsAuthenticated(sessionID string) {
...
 }
 
+// UserPreCheck checks if the user is valid and can be allowed to authenticate.
+func (b *Broker) UserPreCheck(username string) error {
+	for _, suffix := range b.allowedSSHSuffixes {
+		if strings.HasSuffix(username, suffix) {
+			return nil
+		}
+	}
+	return errors.New("username does not match the allowed suffixes")
```
didrocks, line 553: How is then this error transcribed to ssh?

internal/broker/broker_test.go
```diff
@@ -661,6 +669,21 @@ func TestEndSession(t *testing.T) {
 	require.NoError(t, err, "EndSession should not have returned an error when ending an existent session")
 }
 
+func TestUserPreCheck(t *testing.T) {
```
didrocks, line 672: I would really have a more comprehensive tests here.

Some with different brokers, like one not having the `ssh_allowed_suffixes` key, this is the main use case after all :)

Also, you can have one where the broker has more than one suffix as we only test one here.

So, I think if this, you end up with 5 test cases, which then justify table testing so that we don’t have to add it after we grew up in complexity as in other parts of authd where I needed to turn some tests to table later on :).

didrocks (changes requested):
The production code itself looks good to me! I have some issues with the test themselves that are not testing the main default use case (nothing configured) or multiple suffixes in the config.

internal/broker/broker.go
denisonbarbosa: It's not. This error goes to NSS, but we don't show it anywhere else besides the system logs.

didrocks: ok, sounds good to me then!

internal/broker/broker_test.go
```diff
@@ -661,6 +669,50 @@ func TestEndSession(t *testing.T) {
...
+	t.Parallel()
+
+	tests := map[string]struct {
+		username        string
+		allowedSuffixes []string
+
+		wantErr bool
+	}{
+		"Successfully allow user": {},
+		"Successfully allow user through multiple suffixes": {
```
didrocks, line 682: nitpick: I find "through multiple suffixes" a little bit confusing, why not "with multiple suffixes permitted" or something like this?

internal/broker/broker_test.go
```diff
@@ -661,6 +669,50 @@ func TestEndSession(t *testing.T) {
...
+func TestUserPreCheck(t *testing.T) {
+	t.Parallel()
+
+	tests := map[string]struct {
+		username        string
+		allowedSuffixes []string
+
+		wantErr bool
+	}{
+		"Successfully allow user": {},
```
didrocks, line 681: I find the default case here a little bit confusing as we don’t have many of them, and I would expect the default config to be the default shipped to the user, so no-one permitted.

We don’t have yet a lot of defaults using the default username, maybe we should just skip the "empty" trick, wdyt?

didrocks (changes requested):
I would clarify the tests before we land

PR #50, opened 2024-07-03
refactor: Fail authentication sooner if failed to fetch user info

denisonbarbosa:
When first authenticating with QRCode, the user would have to also create its local password before knowing if the authentication failed. Now, the user info is fetched before the password creation prompt, and authentication will fail before asking for password creation. 

UDENG-3414


didrocks (approved):
Approving but! I really think the test case should have a comment above telling that it’s testing two things:
- immediate denial after token reach despite multiple mfas
- quotes in error string.

That way, we illustrate those use cases in the test.

PR #52, opened 2024-07-08
fix: split config files and drop write-files plug

didrocks:
To be able to drop the write files permissions, let's split the config file in 2 parts.

That will also allow us to not depend on the file on /etc by making that built-in.

UDENG-3485


jibel (approved):
Thanks. Looks good.

PR #58, opened 2024-07-15
fix(broker): Update UserPreCheck with API changes in authd

denisonbarbosa:
Refer to [authd#430](https://github.com/ubuntu/authd/pull/430) for the rationale of the changes.

UDENG-3528

### DIFF
--- a/internal/broker/broker.go
+++ b/internal/broker/broker.go
@@ -550,13 +550,24 @@ func (b *Broker) CancelIsAuthenticated(sessionID string) {
 }
 
 // UserPreCheck checks if the user is valid and can be allowed to authenticate.
-func (b *Broker) UserPreCheck(username string) error {
+func (b *Broker) UserPreCheck(username string) (string, error) {
+	found := false
 	for _, suffix := range b.allowedSSHSuffixes {
 		if strings.HasSuffix(username, suffix) {
-			return nil
+			found = true
+			break
 		}
 	}
-	return errors.New("username does not match the allowed suffixes")
+
+	if !found {
+		return "", errors.New("username does not match the allowed suffixes")
+	}
+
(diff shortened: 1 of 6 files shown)


internal/broker/broker.go
```diff
@@ -550,13 +550,30 @@ func (b *Broker) CancelIsAuthenticated(sessionID string) {
...
+			break
 		}
 	}
-	return errors.New("username does not match the allowed suffixes")
+
+	if !found {
+		return "", errors.New("username does not match the allowed suffixes")
+	}
+
+	u := userInfo{
```
didrocks, line 566: I would factor userInfo creation (so that if one day, we give the option the option to change the shell or expand other options with sane defaults, this is reflected there), wdyt?

didrocks (changes requested):
Excellent work! Just one small connect for future evolution :)

internal/broker/broker.go
denisonbarbosa: Sounds like a good idea, for sure. I'll update it.

didrocks: I’m unsure we understood each other, what I meant here is that the 2 paths we are creating userInfo should be the same function call which is something like newUserInfo(). Does it makes sense?

didrocks (commented):
Commented on the generic userinfo creation :)

The idea is really if the homebasedir change, we only have only one place to update, or if we add a new fields, only one function needs update :)

internal/broker/broker.go
denisonbarbosa: I'm actually ashamed that my brain turned so off for that change :joy:

didrocks (approved):
Indeed, we even had an existing helper for it :p userInfoFromClaims

PR #61, opened 2024-07-16
fix(tests): Add hidden flag to run daemon with custom bus name

denisonbarbosa:
Changing the bus service to use a default bus name broke the parallel tests as we were trying to export the same bus multiple times. Now, each test instance will export its own bus for the daemon tests to avoid conflicts.

UDENG-3553


didrocks (changes requested):
I’m unsure I am a big fan of this. I would really use a similar approach than the one we have with authd on env variable, only exported for testing or integrationtests that we reuse in a single package and so, don’t leak it here.

We could work on that in a pairing session if you want.

denisonbarbosa:
Superseded by #63.

PR #63, opened 2024-07-17
fix(tests): No longer run daemon tests in parallel

denisonbarbosa:
After the changes introduced in #52, the broker bus is always exported with the same name. This means we can't run multiple instances of the daemon as they would fail when competing for the same bus name. Since we can't run multiple daemons in production as well (for the same reason), it's fine to drop the parallel testing and run them sequentially instead.

UDENG-3553


denisonbarbosa:
This supersedes #61. After discussion, we agreed that exporting a flag (even a hidden one) to override the bus name used by the service is not ideal.

didrocks (approved):
Yeah, seems like the best approach

PR #70, opened 2024-07-18
fix: Ignore non-group objects from graphsdk

denisonbarbosa:
The `MemberOf` API call from `msgraph` can also return Directory Roles and Administrative Units for a given user (as can be seen in [issue 431 of authd](https://github.com/ubuntu/authd/issues/431). Since the mentioned types have a different structure and considering we only care about the user groups, we should ignore the objects that do not match the Groups type.


didrocks (approved):
As discussed, happy that conversation could even trigger a potential future path without msgraph in the future!

didrocks:
Merging to unblock our users.

PR #71, opened 2024-07-19
Fix autoupdate brokers specific branch when there is no conflict

didrocks:
We were not pushing the local branch but a local branch with the same name.

UDENG-3664


didrocks:
Ok, let’s try this now with the incoming PR

PR #74, opened 2024-07-19
Really fix automerge

didrocks:
We need to specify the original branch and not only the destination, otherwise the branch is removed.
We already mark the original branch to be autodeleted on the PR, no need to force it too.
Allow manual workflow triggers.


didrocks:
Let’s get that so that we can automerge.

PR #76, opened 2024-07-19
Log group information on parsing error

denisonbarbosa:
We need more information as to why some group objects do not have the expected information attached.


didrocks (approved):
Let’s try to get this looks and see later how to refactor this.

PR #80, opened 2024-07-22
Fix invalid logging for unexpected group

didrocks:
* Fix Print -> Printf
* Ensure we print the value of the pointer and not the pointer itself
* Use pretty (which has less dependencies than pp, due to no color support) to print the actual group content for debugging.


denisonbarbosa (approved):
Yeah, I didn't notice it until it was too late 🥹

Looks good to me, thanks for fixing it!

PR #82, opened 2024-07-22
Only include build-time when you import the package

didrocks:
Let’s trust ourself to import it when needed, that way, we have the linter running when it’s needed.

Now, only withmsentraid.go have the build-tag to import it.


denisonbarbosa (approved):
LGTM!

PR #84, opened 2024-07-22
Dereferencing msGroup when logging

didrocks:
Needs to defer for logging.


denisonbarbosa (approved):
Looks good to me!

Ps.: Don't forget to adjust the PR description as well before merging.

PR #89, opened 2024-07-24
Check MSEntraID group access permissions

didrocks:
Current user group membership always returns, with User.Read.All the list of groups, but without their details.
To access their details, we need to list all groups that the user has potentially access too. Add a check for it to return a better error for those use cases.

### DIFF
--- a/internal/providers/msentraid/msentraid.go
+++ b/internal/providers/msentraid/msentraid.go
@@ -14,6 +14,7 @@ import (
 	"github.com/k0kubun/pp"
 	msgraphsdk "github.com/microsoftgraph/msgraph-sdk-go"
 	msauth "github.com/microsoftgraph/msgraph-sdk-go-core/authentication"
+	msgraphgroups "github.com/microsoftgraph/msgraph-sdk-go/groups"
 	"github.com/microsoftgraph/msgraph-sdk-go/models"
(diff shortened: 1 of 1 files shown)


internal/providers/msentraid/msentraid.go
```diff
@@ -53,6 +54,17 @@ func (p Provider) GetGroups(token *oauth2.Token) ([]group.Info, error) {
...
 
+	// Check GroupMember.Read.All access
+	var topOne int32 = 1
+	requestOptions := &msgraphgroups.GroupsRequestBuilderGetRequestConfiguration{
+		QueryParameters: &msgraphgroups.GroupsRequestBuilderGetQueryParameters{
+			Top: &topOne, // Limit to only one group
+		},
+	}
+	if _, err = client.Groups().Get(context.Background(), requestOptions); err != nil {
+		return nil, fmt.Errorf("could not access to user's groups: %v", err)
```
denisonbarbosa, line 65:
```suggestion
		return nil, fmt.Errorf("could not access user's groups: %v", err)
```

denisonbarbosa (approved):
LGTM, besides a small nitpick on the error message.

internal/providers/msentraid/msentraid.go
didrocks: good one!

PR #92, opened 2024-07-24
feat: Allow broker and sessions to start in offline mode

denisonbarbosa:
Even though we "allowed" offline authentication, the broker would not start without a network connection.

Now, the broker does not rely on a connection to run and now sessions have control over the auth config needed to authenticate the user online. This means we can start sessions offline when we fail to connect to the provider.

UDENG-3591
UDENG-3593
UDENG-3596

### DIFF
--- a/internal/broker/broker.go
+++ b/internal/broker/broker.go
@@ -10,12 +10,12 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"net/http"
 	"os"
 	"path/filepath"
 	"slices"
 	"strings"
 	"sync"
+	"time"
 
 	"github.com/coreos/go-oidc/v3/oidc"
 	"github.com/google/uuid"
@@ -26,7 +26,10 @@ import (
 	"golang.org/x/oauth2"
 )
 
-const maxAuthAttempts = 3
+const (
+	maxAuthAttempts    = 3
+	maxRequestDuration = 5 * time.Second
+)
 
 // Config is the configuration for the broker.
 type Config struct {
@@ -39,8 +42,11 @@ type Config struct {
 
 // Broker is the real implementation of the broker to track sessions and process oidc calls.
 type Broker struct {
-	providerInfo       providers.ProviderInfoer
-	auth               authConfig
+	providerInfo providers.ProviderInfoer
+	issuerURL    string
+	oidcCfg      oidc.Config
+
+	cachePath          string
 	homeDirPath        string
 	allowedSSHSuffixes []string
 
@@ -48,16 +54,6 @@ type Broker struct {
 	currentSessionsMu sync.RWMutex
 
 	privateKey *rsa.PrivateKey
-}
-
-type authConfig struct {
-	cachePath string
-
-	provider    *oidc.Provider
-	providerURL string
-
-	oidcCfg  oidc.Config
-	oauthCfg oauth2.Config
 }
 
 type sessionInfo struct {
@@ -70,12 +66,20 @@ type sessionInfo struct {
 	authModes         []string
 	attemptsPerMode   map[string]int
 
+	authCfg   authConfig
 	authInfo  map[string]any
+	isOffline bool
 	cachePath string
 
 	currentAuthStep int
 
 	isAuthenticating *isAuthenticatedCtx
+}
+
+// authConfig holds the required values for authenticating a user with OIDC.
+type authConfig struct {
+	provider *oidc.Provider
+	oauth    oauth2.Config
 }
 
 type isAuthenticatedCtx struct {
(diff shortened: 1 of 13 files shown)


internal/broker/broker.go
```diff
@@ -191,6 +181,23 @@ func (b *Broker) NewSession(username, lang, mode string) (sessionID, encryptionK
 	return sessionID, base64.StdEncoding.EncodeToString(pubASN1), nil
 }
 
+func (b *Broker) connectToProvider(ctx context.Context) (authCfg authConfig, err error) {
+	ctx, cancel := context.WithTimeout(ctx, maxRequestTime)
+	defer cancel()
+
+	authCfg.provider, err = oidc.NewProvider(ctx, b.issuerURL)
```
didrocks, line 188: I don’t like that we are augmenting the object little by little and it’s not really the pattern that we have.

In general, I think we should assign to new variables, like provider and oauth and `return authConfig{provider: provider, oauth: oauth}, nil` (probably in multiple lines like the other constructors.

internal/broker/broker.go
```diff
@@ -204,24 +211,19 @@ func (b *Broker) GetAuthenticationModes(sessionID string, supportedUILayouts []m
...
 
-	// Checks if the provider is accessible and if device authentication is supported
-	providerReachable := true
-	r, err := http.Get(strings.TrimSuffix(b.auth.providerURL, "/") + "/.well-known/openid-configuration")
-	// This means the provider is not available or something bad happened, so we assume no connection.
-	if err != nil || r.StatusCode != http.StatusOK {
-		providerReachable = false
+	endpoints := make(map[string]struct{})
+	if session.authCfg.provider != nil {
+		if session.authCfg.provider.Endpoint().DeviceAuthURL != "" {
```
didrocks, line 216: this is rather an `&&` in a single if, no?

internal/broker/broker.go
```diff
@@ -204,24 +211,19 @@ func (b *Broker) GetAuthenticationModes(sessionID string, supportedUILayouts []m
...
+			endpoints["device_auth"] = struct{}{}
```
didrocks, line 217: I think in a longer term, we should actually use a struct with common capabilities to declare them.

didrocks (changes requested):
I have some very small syntax change requets, but in general, the code looks good to me! Nice :)

I think you did check that the APIs handles context cancellation, right?

I feel the testing story pretty weak though and I’m sure the coverage reflects this. Can we try to have some endpoints hanging, as they are separated from each other, so that we can trigger the various context cancellation hanging and test those patterns?

internal/broker/broker.go
denisonbarbosa: I was going for a && at first, but then wasn't sure whether Go evaluates the if lazily or not. Do you know?

didrocks: Hem, you could test in a playground :p

But yeah, to answer on that one, it is lazily (like most of the languages)

internal/broker/broker.go
```diff
@@ -433,7 +432,12 @@ func (b *Broker) handleIsAuthenticated(ctx context.Context, session *sessionInfo
 			return AuthDenied, errorMessage{Message: "could not get required response"}
 		}
 
-		t, err := session.authCfg.oauth.DeviceAccessToken(ctx, response, b.providerInfo.AuthOptions()...)
+		if response.Expiry.IsZero() {
+			response.Expiry = time.Now().Add(24 * time.Hour)
```
didrocks, line 436: You are being very generous :) I think less is acceptable to have hanging connection. Probably something like an hour is more than enough and will free up resources if this hangs :p

Why this is not `maxRequestTime` btw?

internal/broker/export_test.go
```diff
@@ -110,5 +110,15 @@ func (b *Broker) FetchUserInfo(sessionID string, cachedInfo *authCachedInfo) (us
 	return uInfo, err
 }
 
+func (b *Broker) IsOffline(sessionID string) (bool, error) {
```
didrocks, line 113: Mind adding a comment here? (even if explicit, but as we can return an error…)

internal/broker/export_test.go
```diff
@@ -110,5 +110,15 @@ func (b *Broker) FetchUserInfo(sessionID string, cachedInfo *authCachedInfo) (us
...
+	if err != nil {
+		return false, err
+	}
+	return session.isOffline, nil
+}
+
 // UserInfo exposes the private userInfo for tests.
 type UserInfo = userInfo
+
+const MaxRequestTime = maxRequestTime
```
didrocks, line 124: ditto

didrocks (changes requested):
Few changes requests, but way better now that we have tests, and I saw you adapated the code, so the tests were necessary ;)

internal/broker/broker.go
denisonbarbosa: This takes a while since we generate the code when the user selects the device authentication mode. Considering the time he would take to open the browser, type everything, and so on... It can't be a short time. Also, when it's zero it technically does not expire, that's why I went for a day (I could replace the context instead, but I think this is simpler).

didrocks: Right, but I’m just thinking about requests piling up. And it won't take a full day to open a browser, let’s make in an hour and not more, if you don’t mind?

denisonbarbosa: Yeah, sure... It doesn't matter, tbh, but I will change it to an hour instead.

didrocks (approved):
Excellent, let’s handle the jwt last request separately, as discussed!

PR #96, opened 2024-07-30
feat: Move fetchUserInfo logic to the providers pkg

denisonbarbosa:
We used to control how the ID token was parsed into claims in the broker. This would become progressively unmaintainable as each provider could understand the claims differently. 

Instead of enforcing the providers to use the email as the username for local authentication, they control what will be parsed from the ID Token and return a UserInfo to the broker, which only fills the empty values with defaults (if any).

ubuntu/authd#439

UDENG-3698

### DIFF
--- a/internal/broker/broker.go
+++ b/internal/broker/broker.go
@@ -20,7 +20,7 @@ import (
 	"github.com/coreos/go-oidc/v3/oidc"
 	"github.com/google/uuid"
 	"github.com/ubuntu/authd-oidc-brokers/internal/providers"
-	"github.com/ubuntu/authd-oidc-brokers/internal/providers/group"
+	"github.com/ubuntu/authd-oidc-brokers/internal/providers/info"
 	"github.com/ubuntu/decorate"
 	"golang.org/x/exp/slog"
 	"golang.org/x/oauth2"
@@ -570,7 +570,8 @@ func (b *Broker) UserPreCheck(username string) (string, error) {
 		return "", errors.New("username does not match the allowed suffixes")
 	}
 
-	encoded, err := json.Marshal(b.userInfoFromClaims(claims{Email: username, Name: username}, nil))
+	u := info.NewUser(username, filepath.Join(b.homeDirPath, username), "", "", "", nil)
+	encoded, err := json.Marshal(u)
 	if err != nil {
 		return "", fmt.Errorf("could not marshal user info: %v", err)
 	}
@@ -604,7 +605,7 @@ func (b *Broker) updateSession(sessionID string, session sessionInfo) error {
 type authCachedInfo struct {
 	Token      *oauth2.Token
 	RawIDToken string
-	UserInfo   userInfo
+	UserInfo   info.User
 }
 
 // cacheAuthInfo serializes the access token and cache it.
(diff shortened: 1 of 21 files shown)


didrocks (changes requested):
This is looking good! I really prefer how the API shapes now around the provider, while still providing some commonalities.

I have few suggestions that I think will articulate the package name a little bit better, but nothing big to change IMHO.

internal/broker/broker.go
```diff
@@ -675,74 +675,42 @@ func (b *Broker) loadAuthInfo(ctx context.Context, session *sessionInfo, passwor
...
 }
 
-func (b *Broker) userClaims(ctx context.Context, session *sessionInfo, rawIDToken, username string) (userClaims claims, err error) {
-	// We need to verify the token to get the user claims.
-	// Due to an internal implementation, the verification process can't be cancelled, so we can't add a timeout for it.
-	idToken, err := session.authCfg.provider.Verifier(&b.oidcCfg).Verify(ctx, rawIDToken)
-	if err != nil {
-		return claims{}, fmt.Errorf("could not verify token: %v", err)
+// fillUserInfoDefaults fills the userinfo struct with default values if they are not set.
+func (b *Broker) fillUserInfoDefaults(u models.UserInfo) models.UserInfo {
```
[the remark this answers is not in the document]
denisonbarbosa: I have an idea, let me know if you like it.

internal/providers/info/info.go
```diff
@@ -0,0 +1,45 @@
...
+
+// NewDefaultUser creates a new user with default values.
+func NewDefaultUser(name string) User {
+	return NewUser(name, "", "", "", "", nil)
+}
+
+// NewUser creates a new user with the specified values.
+//
+// It fills the defaults for Shell and Gecos if they are empty.
+func NewUser(name, home, uuid, shell, gecos string, groups []Group) User {
```
didrocks, line 28: I would still have tests for them. Please ensure that all public API for each package is tested. We have already seen places where you skipped the testNew(), then New() becomes more sofisticated and I need to implement tests for those…

internal/providers/info/info.go
```diff
@@ -0,0 +1,45 @@
...
+	Name   string  `json:"name"`
+	UUID   string  `json:"uuid"`
+	Home   string  `json:"dir"`
+	Shell  string  `json:"shell"`
+	Gecos  string  `json:"gecos"`
+	Groups []Group `json:"groups"`
+}
+
+// NewDefaultUser creates a new user with default values.
+func NewDefaultUser(name string) User {
```
didrocks, line 21: This isn’t called anywhere, is it?

internal/broker/broker.go
```diff
@@ -675,42 +676,33 @@ func (b *Broker) loadAuthInfo(ctx context.Context, session *sessionInfo, passwor
...
-
-// fillUserInfoDefaults fills the userinfo struct with default values if they are not set.
-func (b *Broker) fillUserInfoDefaults(u models.UserInfo) models.UserInfo {
-	if u.Shell == "" {
-		u.Shell = "/usr/bin/bash"
-	}
-	if u.Home == "" {
-		u.Home = filepath.Join(b.homeDirPath, u.Name)
+	// This means that home was not provided by the claims, so we need to set it to the broker default.
+	if userInfo.Home == "" {
```
didrocks, line 703: I would do that slightly differently:
- in the `info.NewUser()` call, if home claims is empty, set it to `Name`.
- then, here, if userInfo.Home isn’t an absolute path: prepend and join `b.homeDirPath`. That way, we allow:
* providers to provide an absolute path and we respect it
* providers to provide a home directory surrname and we respect it
* still provide a default.

Ofc, we need tests for those :)

didrocks (changes requested):
This looks better! Especially now that the defaults (apart from the one that relies on the broker) is where it’s meant to be.

Any package without _test.go and any public API untested should raise a warning now in your mind :)

Also, spotted some unused function + added a suggestion which will be more flexible and fix more use cases in the future I think.

internal/providers/info/info.go
denisonbarbosa: oops, I thought I removed it 😅

internal/broker/broker.go
denisonbarbosa: Well, we do still allow providers to do all of this, no? I'm not sure I like the idea of setting an "invalid" value on the field, as it would basically force us to check the value every time.

didrocks: As said, the 1st one is what will be set by the providers which can override it.

on the second part: well, right now, they set a relative path, we are just taking it and use that "invalid" value :)

This is why at least, here, we would have a contract with the providers:
- send something relative from the broker -> we will prefix it to make it absolute where the config set is and this is a valid use case.
- send something absolute, and we will respect it.
- have a default for complete empty value.

didrocks: answered online.

didrocks (approved):
Great! Looks good now :)

PR #99, opened 2024-08-05
tests: Add tests for concurrent IsAuthenticated calls

denisonbarbosa:
This is to ensure one session does not interfere with another when happening at the same time on the broker. To understand better the tests, run them with the verbosity flag (-v) so that you can see the ordering of the calls.

UDENG-2974


didrocks (approved):
Not the biggest fan of controlling that with a timeout (we could have used another channel to control the order and what starts first), but I think this is good enough to help the race detector here. Thanks!

PR #111, opened 2024-08-19
Re-use the existing local password

adombeck:
When authenticating again via remote auth, do not ask user to set a new password if there already is a local password set. Ask for the existing password instead.

Closes https://github.com/ubuntu/authd/issues/369

UDENG-3116

### DIFF
--- a/internal/providers/msentraid/msentraid.go
+++ b/internal/providers/msentraid/msentraid.go
@@ -198,8 +198,7 @@ func (p Provider) CurrentAuthenticationModesOffered(
 		}
 		if tokenExists {
 			offeredModes = append([]string{"password"}, offeredModes...)
-		}
-		if currentAuthStep > 0 {
+		} else if currentAuthStep > 0 {
 			offeredModes = []string{"newpassword"}
 		}
 	}


internal/providers/msentraid/msentraid.go
```diff
@@ -198,8 +198,7 @@ func (p Provider) CurrentAuthenticationModesOffered(
 		}
 		if tokenExists {
 			offeredModes = append([]string{"password"}, offeredModes...)
-		}
-		if currentAuthStep > 0 {
+		} else if currentAuthStep > 0 {
```
denisonbarbosa, line 201: Ok... So, although this "works", it has some problems. Using the password mode implies also [loading the information that we already have cached for the user](https://github.com/ubuntu/authd-oidc-brokers/blob/1a5fc29619fae63fddcb8ca3ad7dc51d08c52627/internal/broker/broker.go#L460). If the user authenticated online already, we have a new token (with potentially new/changes to user information stored in the ID token) so we can't ignore it and use the cached one instead (which is what will happen if the access token is still valid).

This problem will go away once we implement the new encryption method for the token.

denisonbarbosa (changes requested):

adombeck:
Superseded by https://github.com/ubuntu/authd-oidc-brokers/pull/134

PR #115, opened 2024-08-21
Improve error messages returned by the broker

denisonbarbosa:
Exposing too much on the error messages can allow attackers to grab sensitive information by forcing authentication failures. To prevent leaks, we now print the full error message on the system logs and return a simpler one through dbus.

The difference between what we had and the new ones can be seen in the golden files updated in this PR. The full error message is printed to the logs and can be looked at by running `go test -v`.

UDENG-4076

### DIFF
--- a/internal/broker/broker.go
+++ b/internal/broker/broker.go
@@ -418,10 +418,13 @@ func (b *Broker) IsAuthenticated(sessionID, authenticationData string) (string, 
 }
 
 func (b *Broker) handleIsAuthenticated(ctx context.Context, session *sessionInfo, authData map[string]string) (access string, data isAuthenticatedDataResponse) {
+	defer decorateErrorMessage(&data, "authentication failure")
+
 	// Decrypt challenge if present.
 	challenge, err := decodeRawChallenge(b.privateKey, authData["challenge"])
 	if err != nil {
-		return AuthRetry, errorMessage{Message: fmt.Sprintf("could not decode challenge: %v", err)}
+		slog.Error(err.Error())
+		return AuthRetry, errorMessage{Message: "could not decode challenge"}
 	}
 
 	var authInfo authCachedInfo
@@ -429,6 +432,7 @@ func (b *Broker) handleIsAuthenticated(ctx context.Context, session *sessionInfo
 	case "device_auth":
 		response, ok := session.authInfo["response"].(*oauth2.DeviceAuthResponse)
 		if !ok {
+			slog.Error("could not get device auth response")
 			return AuthDenied, errorMessage{Message: "could not get required response"}
 		}
 
@@ -439,18 +443,21 @@ func (b *Broker) handleIsAuthenticated(ctx context.Context, session *sessionInfo
 		defer cancel()
 		t, err := session.authCfg.oauth.DeviceAccessToken(expiryCtx, response, b.providerInfo.AuthOptions()...)
 		if err != nil {
-			return AuthRetry, errorMessage{Message: fmt.Sprintf("could not authenticate user: %v", err)}
+			slog.Error(err.Error())
+			return AuthRetry, errorMessage{Message: "could not authenticate user remotely"}
 		}
 
 		rawIDToken, ok := t.Extra("id_token").(string)
 		if !ok {
-			return AuthDenied, errorMessage{Message: "could not get id_token"}
+			slog.Error("could not get ID token")
+			return AuthDenied, errorMessage{Message: "could not get ID token"}
 		}
 
 		authInfo = authCachedInfo{Token: t, RawIDToken: rawIDToken}
 		authInfo.UserInfo, err = b.fetchUserInfo(ctx, session, &authInfo)
 		if err != nil {
-			return AuthDenied, errorMessage{Message: fmt.Sprintf("could not get user info: %v", err)}
+			slog.Error(err.Error())
+			return AuthDenied, errorMessage{Message: "could not fetch user info"}
 		}
 
 		session.authInfo["auth_info"] = authInfo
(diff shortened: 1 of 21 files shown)


internal/broker/broker.go
```diff
@@ -450,7 +452,7 @@ func (b *Broker) handleIsAuthenticated(ctx context.Context, session *sessionInfo
 		authInfo = authCachedInfo{Token: t, RawIDToken: rawIDToken}
 		authInfo.UserInfo, err = b.fetchUserInfo(ctx, session, &authInfo)
 		if err != nil {
-			return AuthDenied, errorMessage{Message: fmt.Sprintf("could not get user info: %v", err)}
+			return AuthDenied, errorMessage{Message: err.Error()}
```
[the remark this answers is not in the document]
adombeck: So I would like to change this to:

```golang
if err != nil {
    slog.Error(err)
    return AuthDenied, errorMessage{Message: "could not fetch user info"}
}
```

and not call `logAndRedactError` in `fetchUserInfo`.

internal/broker/broker.go
```diff
@@ -418,6 +418,8 @@ func (b *Broker) IsAuthenticated(sessionID, authenticationData string) (string,
 }
 
 func (b *Broker) handleIsAuthenticated(ctx context.Context, session *sessionInfo, authData map[string]string) (access string, data isAuthenticatedDataResponse) {
+	defer decorateErrorMessage(&data, fmt.Sprintf("failed to authenticate user %q", session.username))
+
```
adombeck, line 422: What's the reasoning behind decorating the error message with this string? Do you expect it to be more helpful to the user?

internal/broker/broker.go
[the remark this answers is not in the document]
denisonbarbosa: We tried this approach already. What ends up happening is that, if we redact the message too high in the chain, the error message isn't clear enough to be helpful. So that's why I decided to redact it only when it matters (which is in the private functions that handle sensitive data directly). Looking at the golden files should give you a good example of how the errors will look in the end.

internal/broker/broker.go
denisonbarbosa: Exactly. I'm still not sure we should print this or not. I'm fine with just appending "failed to authenticate user" to the error also.

internal/broker/broker.go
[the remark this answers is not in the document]
adombeck: But you're only adding "could not fetch user info" to the message in `fetchUserInfo` as well. We know that's what's failing here as well, so I don't see why we should it add it further down the chain.

internal/broker/broker.go
adombeck: No strong opinion, but until we notice in practice that mentioning the username would have helped, I prefer to avoid logging and returning personal identifiers in messages.

internal/broker/broker.go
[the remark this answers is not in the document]
denisonbarbosa: This is mainly to avoid too much redundancy in the code (we would need to add a `slog.Error` call to every error handling branch). It's very similar to the pattern we use when decorating the error messages with `decorate`.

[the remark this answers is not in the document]
adombeck: I see that it removes a small bit of redundancy, but I think that not adding a layer of indirection here actually makes the code more readable and makes it more obvious which messages we're returning to the user.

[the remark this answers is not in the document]
adombeck: I think it's also cleaner if we don't even have to use an `err` which we got from another function for the message which we show the user.

internal/broker/broker.go
adombeck: https://github.com/ubuntu/authd-oidc-brokers/pull/115/commits/c67a78cb0a0c7f1f7879660ceffe141fb7914490 LGTM, feel free to squash that into the other commit

internal/broker/broker.go
```diff
@@ -441,18 +442,20 @@ func (b *Broker) handleIsAuthenticated(ctx context.Context, session *sessionInfo
...
 		if err != nil {
-			return AuthRetry, errorMessage{Message: fmt.Sprintf("could not authenticate user: %v", err)}
+			slog.Error(err.Error())
+			return AuthRetry, errorMessage{Message: "could not authenticate user remotely"}
 		}
 
 		rawIDToken, ok := t.Extra("id_token").(string)
 		if !ok {
-			return AuthDenied, errorMessage{Message: "could not get id_token"}
+			return AuthDenied, errorMessage{Message: "could not get ID token"}
```
adombeck, line 451: The only place I could find where this error message is printed is in the PAM module, whose output doesn't end up in the logs IIUC (but is directly shown to the user). So how about we log any error message we return in this function, to ensure that it's in the logs?

denisonbarbosa: I don't think it's necessary to log everything here. This is a simple error message that does not contain any useful/sensitive info, so I think it's fine to keep it as is.

adombeck: I'm not talking about sensitive or additional info, but don't you think it makes sense to be able to tell from the logs that it's this place in the code where the authentication failed?

If I make the broker return this error message, by applying this patch:

```
index 288eda0..4d46805 100644
@@ -447,7 +447,7 @@ func (b *Broker) handleIsAuthenticated(ctx context.Context, session *sessionInfo
                }
 
                rawIDToken, ok := t.Extra("id_token").(string)
-               if !ok {
+               if true {
                        return AuthDenied, errorMessage{Message: "could not get ID token"}
                }
```

And I do the device authentication via `sudo login`, then "could not get ID token" is printed in the terminal, but there is no message about this authentication failure in the logs (unless debug logging is enabled for the PAM module). When trying to log in via GDM, it doesn't even show the error message in GDM. That's probably a bug, but even if it did, I think it would be helpful for debugging if we could tell from the logs why the authentication failed. What do you think?

adombeck (approved):

PR #116, opened 2024-08-21
Use more descriptive error messages

adombeck:
We used "could not get user info" in too many places, making it hard to tell where this error message originated from.

### DIFF
--- a/internal/providers/msentraid/msentraid.go
+++ b/internal/providers/msentraid/msentraid.go
@@ -73,7 +73,7 @@ type claims struct {
 func (p Provider) userClaims(idToken *oidc.IDToken) (claims, error) {
 	var userClaims claims
 	if err := idToken.Claims(&userClaims); err != nil {
-		return claims{}, fmt.Errorf("could not get user info: %v", err)
+		return claims{}, fmt.Errorf("failed to get ID token claims: %v", err)
 	}
 	return userClaims, nil
 }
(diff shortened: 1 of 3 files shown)


adombeck:
@denisonbarbosa: Unfortunately we were changing the same code yesterday, so this PR and !115 conflict. I'm fine with reviewing and merging !115 first and then I can re-do my changes on top of it.

internal/providers/msentraid/msentraid.go
```diff
@@ -83,12 +83,12 @@ func (p Provider) getGroups(token *oauth2.Token) ([]info.Group, error) {
...
 	auth, err := msgraphauth.NewAzureIdentityAuthenticationProvider(cred)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("failed to create AzureIdentityAuthenticationProvider: %v", err)
 	}
 
 	adapter, err := msgraphsdk.NewGraphRequestAdapter(auth)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("failed to create GraphRequestAdapter: %v", err)
```
denisonbarbosa, line 91: Although I  understand the reasoning behind these changes, this messaging is for engineers, not users. Assuming we merge the other PR for redacting the errors, this error message will be logged instead of returned, and it's ok.

But, currently, it does not make much sense for us to return this kind of message to the end user as it is related to implementation details rather than an error that the user will actually understand.

adombeck: Fully agreed, these error messages should be logged, mainly to help us debug the issue, not shown the user in the GDM or terminal.

adombeck:
Rebased on main and resolved conflicts

denisonbarbosa (approved):
LGTM! Thanks for adjusting these, it's going to help a lot when debugging issues!

jibel (approved):
+1

PR #121, opened 2024-08-27
Improve errors when failing to parse config and create broker

denisonbarbosa:
The details are better explained in the commit messages, but TLDR: we now have more directed errors to specific failures and are now evaluating whether the configuration file was edited or it's the same as the template one.

UDENG-3929

### DIFF
--- a/internal/broker/broker.go
+++ b/internal/broker/broker.go
@@ -98,7 +98,7 @@ type Option func(*option)
 
 // New returns a new oidc Broker with the providers listed in the configuration file.
 func New(cfg Config, args ...Option) (b *Broker, err error) {
-	defer decorate.OnError(&err, "could not create broker with provided issuer and client ID")
+	defer decorate.OnError(&err, "could not create broker")
 
 	opts := option{
 		// This is to avoid too much complexity in the tests.
@@ -110,11 +110,16 @@ func New(cfg Config, args ...Option) (b *Broker, err error) {
 	}
 
 	if cfg.CachePath == "" {
-		return &Broker{}, errors.New("cache path must be provided")
+		err = errors.Join(err, errors.New("cache path is required and was not provided"))
 	}
-
-	if cfg.IssuerURL == "" || cfg.ClientID == "" {
-		return &Broker{}, errors.New("issuer and client ID must be provided")
+	if cfg.IssuerURL == "" {
+		err = errors.Join(err, errors.New("issuer URL is required and was not provided"))
+	}
+	if cfg.ClientID == "" {
+		err = errors.Join(err, errors.New("client ID is required and was not provided"))
+	}
+	if err != nil {
+		return nil, err
 	}
 
 	homeDirPath := "/home"
@@ -125,7 +130,8 @@ func New(cfg Config, args ...Option) (b *Broker, err error) {
 	// Generate a new private key for the broker.
 	privateKey, err := rsa.GenerateKey(rand.Reader, 2048)
 	if err != nil {
-		panic(fmt.Sprintf("could not create an valid rsa key: %v", err))
+		slog.Error(err.Error())
+		return nil, errors.New("failed to generate broker private key")
 	}
 
 	b = &Broker{
--- a/internal/dbusservice/dbusservice.go
+++ b/internal/dbusservice/dbusservice.go
@@ -3,6 +3,7 @@ package dbusservice
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"strings"
 
@@ -128,8 +129,17 @@ func parseConfig(cfgPath string) (map[string]map[string]string, error) {
 	for _, section := range iniCfg.Sections() {
 		cfg[section.Name()] = make(map[string]string)
 		for _, key := range section.Keys() {
+			if strings.Contains(key.String(), "<") && strings.Contains(key.String(), ">") {
+				err = errors.Join(err, fmt.Errorf("found invalid character in section %q, key %q", section.Name(), key.Name()))
+				continue
+			}
 			cfg[section.Name()][key.Name()] = key.String()
 		}
+	}
+
+	// This means we found at least one section that was potentially not edited.
+	if err != nil {
+		return nil, fmt.Errorf("config file has invalid values, did you edit the file %q?\n%w", cfgPath, err)
 	}
 	return cfg, nil
 }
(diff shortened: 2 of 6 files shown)


internal/dbusservice/dbusservice.go
```diff
@@ -119,7 +121,16 @@ func New(_ context.Context, cfgPath, cachePath string) (s *Service, err error) {
 
 // parseConfig parses the config file and returns a map with the configuration keys and values.
 func parseConfig(cfgPath string) (map[string]map[string]string, error) {
-	iniCfg, err := ini.Load(cfgPath)
+	cfgContent, err := os.ReadFile(cfgPath)
+	if err != nil {
+		return nil, err
+	}
+
+	if string(cfgContent) == conf.BrokerConfTemplate {
```
didrocks, line 129: I am unsure about that change. Do you feel this is more interesting to the user than knowing they are missing X or Y fields to be valid?

Let me give you an example when this approache fails:
* you install one broker, don’t configure it
* the broker refreshes and we changed the config (added a comment)
* some days after the user starts the broker: this error won’t be triggered.

I really think a better approach is really to valid the fields we need to be edited (maybe not containing any template markup?)

internal/dbusservice/internal_test.go
```diff
@@ -0,0 +1,99 @@
...
+
+[users]
+home_base_dir = /home
+allowed_ssh_suffixes = @issuer.url.com
+`,
+
+	"template": conf.BrokerConfTemplate,
+}
+
+func TestParseConfig(t *testing.T) {
```
didrocks, line 32: This test is basically testing ini.Load() (apart from the last use case) :) 

My main issue is that everytime we update the config, we will need to update this test, so it could use golden files IMHO(marshalling in yaml for instance the maps to ensure the struct has been changed and loaded as we desired).

didrocks (changes requested):
I really question about the untouched configuration handling as there is no reliable way to ensure this was changed or not.

I would otherwise check for invalid values: values with <> markups that we know are invalid, empty fields…

internal/broker/broker.go
```diff
@@ -110,11 +110,13 @@ func New(cfg Config, args ...Option) (b *Broker, err error) {
...
+		return &Broker{}, errors.New("cache path is required and was not provided")
 	}
-
-	if cfg.IssuerURL == "" || cfg.ClientID == "" {
-		return &Broker{}, errors.New("issuer and client ID must be provided")
+	if cfg.IssuerURL == "" {
+		return &Broker{}, errors.New("issuer URL is required and was not provided")
+	}
+	if cfg.ClientID == "" {
+		return &Broker{}, errors.New("client ID is required and was not provided")
```
didrocks, line 119: maybe we should rather check both and then join the errors? It’s annoying when you get an error, fix it, restart, and get another error :p

internal/dbusservice/dbusservice.go
denisonbarbosa: Agreed, this is a better approach!

internal/dbusservice/dbusservice.go
```diff
@@ -139,9 +129,18 @@ func parseConfig(cfgPath string) (map[string]map[string]string, error) {
...
+				err = errors.Join(err, fmt.Errorf("found invalid character in section %q, key %q", section.Name(), key.Name()))
+				continue
+			}
 			cfg[section.Name()][key.Name()] = key.String()
 		}
 	}
+
+	// This means we found at least one section that was potentially not edited.
+	if err != nil {
+		return nil, fmt.Errorf("config file has invalid values:\n%w", err)
```
didrocks, line 142: Could we add something like, "did you edit your configuration file?" + the path to it to help the user?

internal/broker/broker.go
```diff
@@ -110,11 +110,16 @@ func New(cfg Config, args ...Option) (b *Broker, err error) {
 	}
 
 	if cfg.CachePath == "" {
-		return &Broker{}, errors.New("cache path must be provided")
+		err = errors.Join(err, errors.New("cache path is required and was not provided"))
 	}
-
-	if cfg.IssuerURL == "" || cfg.ClientID == "" {
-		return &Broker{}, errors.New("issuer and client ID must be provided")
+	if cfg.IssuerURL == "" {
```
didrocks, line 115: Maybe we should move the template validation here so that we have the whole validation at the same place? (it’s repeated code though), but at least, we can decide which fields are templated and which are not. I don’t any issue that way for a home dir or suffix containing `<`

didrocks (changes requested):
Thanks for the changes! I have few suggestions to ensure we don’t impact all the keys but only the ones we know that are templatable.

internal/broker/broker.go
denisonbarbosa: IMHO, it wouldn't make sense. The broker does not know the configuration file, so it would be weird to have it check for template values in the `broker.New` function, no?

internal/dbusservice/dbusservice.go
```diff
@@ -128,9 +129,18 @@ func parseConfig(cfgPath string) (map[string]map[string]string, error) {
 	for _, section := range iniCfg.Sections() {
 		cfg[section.Name()] = make(map[string]string)
 		for _, key := range section.Keys() {
+			if strings.Contains(key.String(), "<") || strings.Contains(key.String(), ">") {
```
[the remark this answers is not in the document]
denisonbarbosa: I went with the `||` because, as it is a template, it could happen that the user ended up not replacing all the `<>` and the configuration ended up with something like: `actual_issuer.com>` or `<actual_issuer.com`. Does that make sense?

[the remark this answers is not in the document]
didrocks: Yeah, but it means they edited it, and we are just trying to ensure they at least edited it. If the value is then invalid, so be it and another error will trigger down the line, but they will know where to fix it as they touched it once at least :)

internal/broker/broker.go
didrocks: good point, (but right now, the parsing is done in `dbusservice` which doesn't make more sense TBH :p

I think with the double restriction with the `&&`  we can leave it being applied to all fields then, but I expect we will split that in a separate package at some point and validate required fields there too.

denisonbarbosa: > but right now, the parsing is done in dbusservice

Well, this package used to be called `brokerservice` at the time I added this parsing, so it used to make sense :joy:. But I agree that this should probably be moved to its own package later down the line, especially if the config file ends up becoming bigger and more complex.

didrocks (approved):
Excellent, thanks!

PR #123, opened 2024-08-30
Fix debug logs not printed in some packages

adombeck:
We imported the wrong `slog` package in some places, causing debug logs to not be printed (because the log level is not set for that package, so it defaults to INFO).

UDENG-4337


didrocks (approved):
Well spot!

PR #126, opened 2024-08-30
ci: Block merging PRs with fixup commits

adombeck:
We want to avoid force pushing to PRs which have already been reviewed, so we're doing fixup commits instead. To keep a clean git history, those should be squashed before merging (with `git rebase --autosquash`).

This new GitHub action should avoid that we accidentally merge PRs which have unsquashed fixup commits.

### DIFF
--- a/.github/workflows/git.yml
+++ b/.github/workflows/git.yml
@@ -1,0 +1,12 @@
+name: Git Checks
+
+on: [pull_request]
+
+jobs:
+  block-fixup:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2.0.0
+    - name: Block Fixup Commit Merge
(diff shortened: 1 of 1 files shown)


adombeck:
If you agree that this is useful, I'll create a similar PR for the authd repo

didrocks (commented):
I looked at this action which was not updated for 4 years, so it looks fine. There are some complaints about deprecation notice, but it seems it’s due to the docker container itself.

However, I would have like to see a failing run before (even if it trivially grep on the commit message content, so it looks safe) :) Mind doing this?

adombeck:
> However, I would have like to see a failing run before

Here you go https://github.com/ubuntu/authd-oidc-brokers/actions/runs/10634744090/job/29482774253?pr=126 :)

didrocks (approved):
Perfect! Thanks a lot :)

PR #134, opened 2024-09-05
Store token unencrypted

adombeck:
Currently we encrypt the token with the local password before writing it to the filesystem. We decrypt the token whenever the user logs in and try to refresh it. That requires us to always ask for the local password on login, even if the user authenticated via OIDC.

We discussed various approaches to solve this issue. In the end we decided to go with the simplest approach, i.e. to store the token unencrypted. It is still protected from unprivileged users via filesystem permissions.

To still be able to verify the local password, we now store a hash of the local password in a file.

For a seamless transition, we still support logging in via a local password when there is no password file but an old encrypted token file. This should make the change completely transparent to users.

This PR also uses the term "data directory" for what was previously known as the "cache directory", because we're storing the tokens and password file in there, which should not be considered "cache" (because if the user deletes those, they won't be able to log in with the local password anymore).

UDENG-4372


adombeck:
@didrocks This is not ready to be merged, but I would like to get some feedback from you on the refactorings I did (see the `refactor:` commits). Also, which tests do you want to see here?

didrocks (changes requested):
So, this is not a full review but a very quick initial one (I only spotted some patterns that I think we should fix). In addition to this, my main feedback is that I really think the broker is already doing too much (oidc auth, token storage, password hash storage…), and we are even complexifying it a little bit more.

We are not going to be able to split it right away, but I suggest that at least, we use that opportunity to put the token storage and password hash management each on their own `internal/` package.

That will allow easier testing (smaller scope) and better split of functionalities.

internal/utils/utils.go
```diff
@@ -0,0 +1,34 @@
...
+func FileExists(path string) (bool, error) {
+	_, err := os.Stat(path)
+	if err != nil && !errors.Is(err, os.ErrNotExist) {
+		return false, err
+	}
+	return !errors.Is(err, os.ErrNotExist), nil
+}
+
+// IsDirEmpty checks if the specified directory is empty.
+func IsDirEmpty(path string) (bool, error) {
```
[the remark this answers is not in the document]
didrocks: >> On the file existence, you can
>
> Seems like something is missing here

Yes, scratch that, I removed part of the sentence after seeing the caller side (and why you had to check for the error afterwards)

> That's not a strong argument by itself IMO. I think the Go standard library still lacks a lot of functionality (especially when comparing it to Python's standard library). I like to add simple functions which can be re-used throughout the code into a utils package (once there are many, I split them off into `fileutil`, `stringutil`, `executil`, etc.).

I still think there is a very high value in following the patterns the community in general is following. One article on the util one for instance is https://dave.cheney.net/2019/01/08/avoid-package-names-like-base-util-or-common.

Go is not Python and I don’t write my Go code as my Python code. In Python, I’m following the python pattern, snake case and so on. In Go, I’m following the Go convention and idioms. This allows:
- better onboarding on new Go members who wants to take parts of the projects
- being able to dig into dependencies, in particular, where we are not the author
- have a sense of familiarity and expections when looking at other’s code

Most of the time, without good justifications, diverging from those patterns shows that the person is trying to port other language patterns to it.

I think those as a separate, imported package, would make sense, but clearly, don’t name that `utils` because it will become a mix of unrelated things, which will cross domain boundaries. We could have a package named around `fileutils` to scope them, which would be an improvement.

If we go that route and don’t import anything outside, I suggest though that we have this per domain package with then a complete testsuite on them (this could be easily 100% covered).

> While the functions are only used in one package, we can make them private functions of that package, but I expect that we will want to use these in other packages at some point, and having them as private functions makes them hard to discover.

While I still think we should keep that as private functions at first as we don’t use them anywhere else right now, in large projects in particular, we shouldn’t try to prepare for a future that never comes. This is how we build technical debts, because we will always be wrong. Start small and split up.

This is why I think the argument strongly stand. As I said, you can keep a separate package even if this is not my preferred option, but then, I’m asking to have a better, domain specific, package name and a testsuite for each of them.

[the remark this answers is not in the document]
adombeck: > Go is not Python and I don’t write my Go code as my Python code. 

Neither do I. I didn't say I want to write Go like Python, I said that Go's standard library is lacking and I see the need to extend "standard" functionality in a some way which allows it to be used throughout the project.

> In Python, I’m following the python pattern, snake case and so on. In Go, I’m following the Go convention and idioms. This allows:
> 
>     better onboarding on new Go members who wants to take parts of the projects
>     being able to dig into dependencies, in particular, where we are not the author
>     have a sense of familiarity and expections when looking at other’s code

I agree, but on the one hand, I see a difference between the community's efforts to describe a good and a bad go style (like https://go.dev/wiki/CodeReviewComments) and the opinions of individuals, even if they are respected in the community. On the other hand I see these as guidelines from which you can deviate if you have good reasons to do so.

> we shouldn’t try to prepare for a future that never comes. This is how we build technical debts, because we will always be wrong.

Fully agreed when it comes to added complexity, especially adding abstractions "because it will be useful in the future". I don't see the separate package adding complexity, but I'm fine with moving the functions to the broker package for now and split them off once we want to use them somewhere else.

internal/utils/utils.go
```diff
@@ -0,0 +1,34 @@
...
+	"os"
+)
+
+// FileExists checks if a file exists at the given path.
+func FileExists(path string) (bool, error) {
+	_, err := os.Stat(path)
+	if err != nil && !errors.Is(err, os.ErrNotExist) {
+		return false, err
+	}
+	return !errors.Is(err, os.ErrNotExist), nil
```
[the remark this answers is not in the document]
adombeck: I don't think that's correct. In the case that `err` is `os.ErrNotExist`, we didn't return the error before, and we want to return `false` here.

internal/utils/utils.go
```diff
@@ -0,0 +1,34 @@
...
+
+import (
+	"errors"
+	"os"
+)
+
+// FileExists checks if a file exists at the given path.
+func FileExists(path string) (bool, error) {
+	_, err := os.Stat(path)
+	if err != nil && !errors.Is(err, os.ErrNotExist) {
```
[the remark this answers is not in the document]
adombeck: I don't think we can inline this, because we need `err` outside the if block (see comment below).

internal/utils/utils.go
```diff
@@ -0,0 +1,34 @@
...
+// IsDirEmpty checks if the specified directory is empty.
+func IsDirEmpty(path string) (bool, error) {
+	f, err := os.Open(path)
+	if err != nil {
+		return false, err
+	}
+	defer f.Close()
+
+	_, err = f.Readdirnames(1)
+	if errors.Is(err, os.ErrNotExist) {
```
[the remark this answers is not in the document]
adombeck: `if err != nil && errors.Is(err, os.ErrNotExist)` is redundant, `errors.Is` only returns `true` when `err` is not nil.

internal/dbusservice/dbusservice.go
```diff
@@ -61,35 +58,13 @@ type Service struct {
 }
 
 // New returns a new dbus service after exporting to the system bus our name.
-func New(_ context.Context, cfgPath, cachePath string) (s *Service, err error) {
-	cfg, err := parseConfig(cfgPath)
```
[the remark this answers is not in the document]
adombeck: I'm glad we're on the page regarding that :)

internal/utils/utils.go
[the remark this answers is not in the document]
didrocks: ack on the below comment.

cmd/authd-oidc/daemon/daemon.go
```diff
@@ -101,12 +111,37 @@ func New(name string) *App {
...
+	if err != nil {
+		return err
+	}
+
+	var allowedSSHSuffixes []string
+	if cfg[usersSection][sshSuffixesKey] != "" {
+		allowedSSHSuffixes = strings.Split(cfg[usersSection][sshSuffixesKey], ",")
+	}
+
+	b, err := broker.New(broker.Config{
```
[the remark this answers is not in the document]
adombeck: > I think we shuld have an internal/config package that takes a config file and returns its own Config struct (which will probably allow passing the config to just more than internal/broker) 

I also prefer a struct over a map, so I'll implement that. Currently, the broker config file is only used by the broker, so I don't think we should create a separate package to make it reusable until we actually need that. I moved it to the broker package now.

internal/broker/broker.go
```diff
@@ -475,12 +486,51 @@ func (b *Broker) handleIsAuthenticated(ctx context.Context, session *sessionInfo
...
 			slog.Error(err.Error())
 			return AuthRetry, errorMessage{Message: "could not load cached info"}
 		}
 
+		if useOldEncryptedToken {
+			// We were able to decrypt the old token with the password, so we can now hash and store the password in the
+			// new format.
+			if err = b.hashAndStorePassword(challenge, session.passwordPath); err != nil {
+				slog.Error(err.Error())
+				return AuthDenied, errorMessage{Message: "could not store password"}
```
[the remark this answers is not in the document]
adombeck: > As a rule of thumb, we don’t log the error and return it at each point.

And we also don't do that here. We log the original error and return a generic error message.

> I suggest that we use a similar mechanism here to the redactError in authd: We wrap the initial error with the error we want to return outside to not leak information.
> 
> Then, at the edge of the return caller (in a defer function), we check for the error type and then log the complete error + return the redacted one.
> 
> That way, you ensure that you don’t forget one spot you didn’t log the error before redacting (I’m seeing a lot of slog.Error() + "returning an error" in that function for instance.

I see some problems with that approach, but I'm not sure this PR is the best place to discuss that, because it doesn't actually do anything different regarding the error handling in the broker, it simply follows what we're already doing. Should we move that discussion somewhere else?

internal/broker/broker.go
```diff
@@ -735,3 +840,40 @@ func decorateErrorMessage(data *isAuthenticatedDataResponse, msg string) {
...
 }
+
+func cleanupOldEncryptedToken(path string) {
+	if err := os.Remove(path); err != nil {
+		slog.Warn(fmt.Sprintf("Failed to remove old encrypted token %s: %v", path, err))
+		return
+	}
+
+	// Also remove the parent directory and the parent's parent directory if they are empty. The directory structure was:
+	//   $SNAP_DATA/cache/$ISSUER/$USERNAME.cache
```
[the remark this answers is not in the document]
adombeck: Where do you think such a comment would make sense? I'm hesitant, because I expect that such a comment is either in a place where it doesn't benefit us much, because it's easy to figure out the path by reading the surrounding code (like https://github.com/ubuntu/authd-oidc-brokers/blob/369-store-token-unencrypted/internal/broker/broker.go#L184-L184) or in a place which we're likely to forget to update if we ever change the path (because it's not close to where the path is defined).

internal/broker/broker.go
```diff
@@ -506,17 +556,27 @@ func (b *Broker) handleIsAuthenticated(ctx context.Context, session *sessionInfo
...
 
-	if err := b.cacheAuthInfo(session, authInfo, challenge); err != nil {
+	if err := b.cacheAuthInfo(session, authInfo); err != nil {
 		slog.Error(err.Error())
 		return AuthDenied, errorMessage{Message: "could not cache user info"}
 	}
 
+	// At this point we successfully stored the hashed password and a new token, so we can now safely remove any old
+	// encrypted token.
+	cleanupOldEncryptedToken(session.oldEncryptedTokenPath)
```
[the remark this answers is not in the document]
adombeck: Fixed by returning early in `cleanupOldEncryptedToken` if the old token doesn't exist

internal/broker/broker.go
```diff
@@ -695,6 +757,49 @@ func (b *Broker) loadAuthInfo(ctx context.Context, session *sessionInfo, passwor
...
+
+	hash := hashPassword(password, salt)
+	s := base64.StdEncoding.EncodeToString(append(salt, hash...))
+	if err := os.WriteFile(path, []byte(s), 0600); err != nil {
+		return fmt.Errorf("could not store password: %v", err)
+	}
+	return nil
+}
+
+func (b *Broker) checkPassword(password, path string) (bool, error) {
```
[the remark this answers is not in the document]
adombeck: done

internal/broker/broker_test.go
```diff
@@ -64,7 +64,7 @@ func TestNew(t *testing.T) {
 			bCfg := broker.Config{
 				IssuerURL: tc.issuer,
 				ClientID:  tc.clientID,
-				CachePath: tc.cachePath,
+				DataDir:   tc.cachePath,
```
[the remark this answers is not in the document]
adombeck: Right, looks like I overlooked some occurrences in the this file. Fixed that now.

internal/broker/broker.go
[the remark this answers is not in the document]
didrocks: > And we also don't do that here. We log the original error and return a generic error message.

I agree, but we still handle one error which mirrored to a redacted error.

> I see some problems with that approach, but I'm not sure this PR is the best place to discuss that, because it doesn't actually do anything different regarding the error handling in the broker, it simply follows what we're already doing. Should we move that discussion somewhere else?

Definitively, I’m happy with that part as is, and then, I’ll do another PR with what I had in mind (and change what was merged too during my holidays about the error logging).

internal/broker/broker.go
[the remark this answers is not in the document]
didrocks: The part you hilighted only define the base path. What I thought in particular was on the latter part (the `$ISSUER/$USERNAQME.cache`) or rather its replacement, to where the code lives. Does it make sense?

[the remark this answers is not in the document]
adombeck: > What I thought in particular was on the latter part (the $ISSUER/$USERNAQME.cache) or rather its replacement

That's defined just above, in https://github.com/ubuntu/authd-oidc-brokers/blob/0309d926a9304614b1ab91dec513f811306fc40b/internal/broker/broker.go#L179-L182

I added comments above the assignments to `session.tokenPath` and `session.passwordPath`.

internal/utils/utils.go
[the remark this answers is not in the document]
adombeck: >  I'm fine with moving the functions to the broker package for now and split them off once we want to use them somewhere else.

I'm using it in the `password_test` package as well, so I moved it to it's own package again, named it `fileutils` and added tests.

adombeck:
>  I suggest that at least, we use that opportunity to put the token storage and password hash management each on their own internal/ package.

Done

adombeck:
Strange, the tests passed locally but in the CI the ``TestIsAuthenticated/Authenticating_with_password_refreshes_expired_token` pass` fails. Marking as draft again while I look into that.

internal/broker/export_test.go
```diff
@@ -2,13 +2,39 @@ package broker
...
+}
+
+func (cfg *Config) ClientID() string {
+	return cfg.clientID
+}
+
+func (cfg *Config) IssuerURL() string {
+	return cfg.issuerURL
+}
+
```
denisonbarbosa, line 37: In Go, we tend to drop this kind of builder pattern for the variadic arguments approach. You can see an example here of how we do it: https://github.com/ubuntu/authd/blob/223a072ec87fbd432c3f31861e684042dfe1a941/internal/testutils/golden.go#L32.

internal/broker/export_test.go
```diff
@@ -19,12 +45,36 @@ func (b *Broker) TokenPathForSession(sessionID string) string {
...
 	}
 
-	return session.cachePath
+	return session.tokenPath
 }
 
-// CachePath returns the path to the cache directory for tests.
-func (b *Broker) CachePath() string {
-	return b.cachePath
+func (b *Broker) PasswordFilepathForSession(sessionID string) string {
```
denisonbarbosa, line 51: Unsure why the linter did not complain about this not having a comment, but it should (even though it's on the tests)

internal/broker/testdata/TestConcurrentIsAuthenticated/golden/first_auth_starts_and_finishes_before_second/first_auth
```diff
@@ -1,3 +1,3 @@
-access: granted
-data: '{"userinfo":{"name":"user-timeout-0","uuid":"saved-user-id","dir":"/home/user-timeout-0","shell":"/usr/bin/bash","gecos":"user-timeout-0","groups":[{"name":"remote-group","ugid":"12345"},{"name":"linux-local-group","ugid":""}]}}'
+access: denied
+data: '{"message":"authentication failure: password file does not exist"}'
```
denisonbarbosa, line 2: Be more careful when updating golden files. This is not the expected behavior of the test...

internal/broker/testdata/TestIsAuthenticated/golden/authenticating_still_allowed_if_token_is_missing_scopes/data/provider_url/test-user@email.com/password
```diff
@@ -0,0 +1 @@
+Definitely a hashed password
```
denisonbarbosa, line 1: Thanks for keeping the little joke :stuck_out_tongue_closed_eyes:!

internal/fileutils/fileutils_test.go
```diff
@@ -0,0 +1,145 @@
...
+
+		expectExists bool
+		expectError  bool
+	}{
+		{name: "Returns true when file exists", fileExists: true, expectExists: true},
+		{name: "Returns false when file does not exist", fileExists: false, expectExists: false},
+		{name: "Returns false when parent directory does not exist", fileExists: false, expectExists: false},
+
+		{name: "Error when parent directory is a file", parentDirIsFile: true, expectError: true},
+	}
```
denisonbarbosa, line 26: I'm a bit confused about this... Why did you use a `[]struct{}` instead of our usual `map[string]struct{}`?

internal/providers/msentraid/msentraid.go
```diff
@@ -215,6 +215,7 @@ func (p Provider) CurrentAuthenticationModesOffered(
 	endpoints map[string]struct{},
 	currentAuthStep int,
 ) ([]string, error) {
+	slog.Debug(fmt.Sprintf("In CurrentAuthenticationModesOffered: sessionMode=%q, supportedAuthModes=%q, tokenExists=%t, providerReachable=%t, endpoints=%q, currentAuthStep=%d\n", sessionMode, supportedAuthModes, tokenExists, providerReachable, endpoints, currentAuthStep))
```
denisonbarbosa, line 218: Is this a leftover from a debug session? Feels like too much for an actual log message that we would want to show.

internal/providers/msentraid/msentraid.go
```diff
@@ -237,6 +238,7 @@ func (p Provider) CurrentAuthenticationModesOffered(
 			offeredModes = []string{"newpassword"}
 		}
 	}
+	slog.Debug(fmt.Sprintf("Offered modes: %q", offeredModes))
```
denisonbarbosa, line 241: Does `%q` work with slices? I had no idea about it

internal/token/migration.go
denisonbarbosa, line 1: We need tests for the migration.

internal/token/token.go
```diff
@@ -0,0 +1,53 @@
...
+
+// AuthCachedInfo represents the token that will be saved on disk for offline authentication.
+type AuthCachedInfo struct {
+	Token      *oauth2.Token
+	RawIDToken string
+	UserInfo   info.User
+}
+
+// CacheAuthInfo saves the token to the given path.
+func CacheAuthInfo(path string, token AuthCachedInfo) (err error) {
```
denisonbarbosa, line 22: Maybe a better name for the `AuthCachedInfo` argument? `token` could be misleading since we store more things than just the token itself.

denisonbarbosa (changes requested):
Oof, this is a big one... Overall, it's looking good, but I do have some observations and requests.

internal/broker/broker.go
```diff
@@ -502,22 +586,32 @@ func (b *Broker) handleIsAuthenticated(ctx context.Context, session *sessionInfo
...
 		if !ok {
 			slog.Error("could not get required information")
 			return AuthDenied, errorMessage{Message: "could not get required information"}
 		}
+
+		err = password.HashAndStorePassword(challenge, session.passwordPath)
+		if err != nil {
+			slog.Error(err.Error())
+			return AuthDenied, errorMessage{Message: "could not store password"}
+		}
```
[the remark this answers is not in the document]
adombeck: Yeah I'll probably keep forgetting that for some time, because I'm used to the opposite (try to avoid doing value assignment and comparison on the same line, because it makes the code a bit less readable, especially when the assignment part is quite long, pushing the comparison further to the right). I tried to get my IDE to warn me when I'm not doing it, and also tried enabling all the linters in golangci-lint, but none supported warnings about this, which to me is an indication that it's not widespread to strictly use that format. I looked if https://go.dev/doc/effective_go has recommendations on it and it does not, but it does put the `if err != nil` on a separate line in the [web server example code](https://go.dev/doc/effective_go#web_server). So maybe we could be less strict about it?

internal/broker/broker.go
```diff
@@ -476,10 +498,72 @@ func (b *Broker) handleIsAuthenticated(ctx context.Context, session *sessionInfo
...
+				return AuthRetry, errorMessage{Message: "invalid password"}
+			}
+
+			authInfo, err = tokenPkg.LoadAuthInfo(session.tokenPath)
+			if err != nil {
+				slog.Error(err.Error())
+				return AuthDenied, errorMessage{Message: "could not load cached info"}
+			}
+		}
+
```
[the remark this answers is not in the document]
adombeck: > The amount of nesting and conditionals here indicates that this should have its own function. handleIsAuthenticated does not care how the authInfo is loaded, it only cares about the object itself, so this logic could be moved into a loadAuthInfo function (like we had previously) and it seems like it doesn't even need to be attached to broker.

I'll try to extract something.

> Also, we are exposing way too much information on the error messages here. If you take a look at other modules (e.g. pam_unix) and SSH, we only get an "authentication failure" message to prevent giving too much information to a possible attacker (that's why we had only the "could not load auth info" message before).

I don't think the reason for a generic error message is to avoid giving too much information to an attacker. An attacker cannot do anything with the information that the password file does not exist or similar. The reason for a generic error message here is likely UX. The user doesn't know what "could not load auth info" means and what they can do about it. A good error message should explain what went wrong to the user in terms they can understand, and what they can do about it. A generic error message like "An unexpected error occurred. Please report this issue on <URL>" would be more useful to the user. 

On the other hand, a user facing such an error can't log in (at least not with the current user and authentication method), it might be hard for them to get the system logs and attach them when they report the problem. They might just report the error message they see instead. That's a reason to provide the information in the error message that *we* need to debug the issue.

So I think we should either go with a generic error message which the user can understand, or one which tells us exactly what went wrong, but "could not load auth info" is neither.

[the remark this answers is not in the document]
adombeck: > I'll try to extract something.

I extracted a `UseOldEncryptedToken` function to the `migration.go` file of the `token` package. I tried to also extract the whole `if useOldEncryptedToken { ... } else { ... }` block into a `CheckPasswordAndLoadAuthInfo` function, but it took too many arguments, and if we let `handleIsAuthenticated` return the same error message for any error returned by it, we would hide too much information.

cmd/authd-oidc/daemon/daemon.go
```diff
@@ -52,16 +54,23 @@ func New(name string) *App {
 			a.rootCmd.SilenceUsage = true
 
 			// Set config defaults
```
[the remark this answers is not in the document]
adombeck: done

internal/broker/broker.go
```diff
@@ -35,23 +37,23 @@ const (
...
-	HomeBaseDir        string
-	AllowedSSHSuffixes []string
+	ConfigPath            string
+	DataDir               string
+	OldEncryptedTokensDir string
+
+	clientID           string
+	issuerURL          string
+	homeBaseDir        string
+	allowedSSHSuffixes []string
```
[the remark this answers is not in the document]
adombeck: > I think that the Config struct knowing about the path we use to parse it is an antipattern.

Yeah I'm also not sure how I feel about that. We could move the config path out of the struct and pass it as a separate argument to `broker.New`.

internal/broker/broker.go
```diff
@@ -102,6 +107,13 @@ type Option func(*option)
 func New(cfg Config, args ...Option) (b *Broker, err error) {
 	defer decorate.OnError(&err, "could not create broker")
 
+	if cfg.ConfigPath != "" {
+		err = parseConfig(cfg.ConfigPath, &cfg)
+		if err != nil {
+			return nil, fmt.Errorf("could not parse config: %v", err)
+		}
+	}
```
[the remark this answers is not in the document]
adombeck: I don't think having two structs `broker.Config` and a `broker.brokerConfig` is a good idea. I prefer using a single struct if they don't serve a different purpose (which the difficulty to find different names shows). I don't see a problem with encapsulation, `parseConfig` is an unexported function which sets unexported fields of the struct. There is no object here which sees something it shouldn't see.

internal/broker/broker.go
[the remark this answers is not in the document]
denisonbarbosa: > I don't think the reason for a generic error message is to avoid giving too much information to an attacker

This is not a question of thinking. I told you what other modules do and they do not give you, the user trying to authenticate, any information as to why the authentication failed, only that it failed. We need to always keep in mind that we support SSH and so on, which means "anyone" could try and authenticate against authd on an exposed server (even worse, since we do go around one of the many layers regarding SSH safety measures [the user existence in the system]).

> On the other hand, a user facing such an error can't log in (at least not with the current user and authentication method), it might be hard for them to get the system logs and attach them when they report the problem. They might just report the error message they see instead. 

This is expected. We won't be able to get the logs unless a user, capable of authenticating, logs into the system and grabs them for us, same thing happens in other authenticators also (and we are already being quite gentle in what we print in the error messages).

> That's a reason to provide the information in the error message that we need to debug the issue.

We should not use this criteria to decide what we print to the _user_. That's why we log the original errors and print something else. If the user has a problem with authentication, they will likely talk to their IT admin before coming to us (or we will ask they to do so, as we'll need logs for it regardless of what error message we decide to print).

> So I think we should either go with a generic error message which the user can understand, or one which tells us exactly what went wrong, but "could not load auth info" is neither

That was not my point. I'm 100% down to use a better message if you have one.

internal/broker/broker.go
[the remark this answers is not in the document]
denisonbarbosa: Could be an approach, for sure. Not sure how it would like. Doesn't it feel wrong to pass a file path as an argument to the broker creation, though? I think that splitting into two configuration structs (one for what we need to create the broker and one for the broker itself) might be better. But passing `ConfigPath`, `DataDir` and `OldEncryptedTokensDir` as arguments to broker.New also could be good.

Honestly, I have no strong opinion regarding which of the two is the best, so it's your call.

internal/broker/broker.go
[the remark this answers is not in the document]
denisonbarbosa: Sure, but it should take a path as an argument and return a parsed configuration rather than filling a pointer.

internal/broker/broker.go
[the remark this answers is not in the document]
adombeck: > This is not a question of thinking.

Of course it is. Other modules use a generic error message. Your interpretation is that they don't want to give too much information to an attacker. I gave you my interpretation. I try to avoid presenting my views and thoughts as facts, as is appropriate in a respectful conversation in my opinion.

> We need to always keep in mind that we support SSH and so on

Yes, we need to make sure that we don't leak any sensitive information or anything an attacker could use. I don't see anything that could be useful for an attacker in those error messages though. 

> That was not my point.

You refer to other modules which use a very generic error message like "authentication failure". My point is that we are not using such a generic error message, we already have "could not decode challenge", "could not get required response",  "could not authenticate user remotely",  "could not get ID token",  "could not fetch user info",  "could not load cached info",  "could not fetch user info",  "empty challenge",  "could not get required information", and "could not cache user info". I'm just adding to the already long list. I'm keeping the pattern. I'm not saying that we should keep doing that, and I explain why I think we should not. But if we decide to do it differently, let's please do so in a separate PR.

> I'm 100% down to use a better message if you have one.

 "An unexpected error occurred. Please report this issue." If the user is using a text interface (i.e. SSH or `sudo login` on the command-line, but not gdm), we can also include the URL which they can report issues at:  "An unexpected error occurred. Please report this issue on https://github.com/ubuntu/authd/issues".

internal/broker/broker.go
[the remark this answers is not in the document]
adombeck: > Doesn't it feel wrong to pass a file path as an argument to the broker creation, though?

Doesn't feel wrong to me.

> I think that splitting into two configuration structs (one for what we need to create the broker and one for the broker itself) might be better.

Can you propose sensible names for those two structs? I can't think of any, and I think they serve the same purpose, so I would like to keep it as one struct.

internal/broker/broker.go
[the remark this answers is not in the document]
adombeck: >  but it should take a path as an argument 

It already does.

> and return a parsed configuration rather than filling a pointer

Not if it requires making the code more complex, which it does AFAICT. I don't understand the strong opposition to passing a pointer and letting the called function modify the fields when the function signature makes it clear that that's what happens.

internal/broker/broker.go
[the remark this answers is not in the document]
denisonbarbosa: The only thing that comes to mind now is `broker.Paths`, which would be the required thing for creating the broker and `broker.Config` which would be the values that we parse from the configuration file. How does it sound? Honest opinion.

internal/broker/export_test.go
adombeck: done

internal/fileutils/fileutils_test.go
adombeck: fixed

internal/providers/msentraid/msentraid.go
adombeck: I did add it during a debug session but I left it on purpose in a [separate commit](b4b505cc049422d03d8a0ecccf1a9eaa3e09ba93) because I think it will be useful in the future:

```
Add debug output
    
These messages should help debug cases where no authentication mode is
offered.
```

internal/token/migration.go
adombeck: added

internal/token/token.go
adombeck: I actually wanted to propose to rename `CacheAuthInfo` / `LoadAuthInfo` to `StoreToken` / `LoadToken`, and to refer to the content of the `token.json` file as "token" instead of "AuthInfo" or "CachedInfo" (as we already do in many comments and log messages). Yes, that content is more than just the *OAuth2* token, but I think we can still refer to it as our (authd broker) token. What do you think? Would also be interested in @didrocks opinion.

internal/broker/export_test.go
adombeck: If I understand correctly, that requires adding a function `NewConfig(opts ...func(*Config)) *Config` and functions like:

```
func WithClientID(clientID string) func(*Config) {
	return func(cfg *Config) {
		cfg.clientID = clientID
	}
}
```

which can then be passed to `NewConfig`, right? That doesn't allow setting fields after initialization of the struct, as we do [here]( https://github.com/ubuntu/authd-oidc-brokers/blob/dd8e0c086ea07c470dd6d0f04499db0369eebdfa/internal/broker/helper_test.go#L35-L37) though, right?

adombeck: Also, I have the impression that the setters I added follow the pattern of the other functions defined in the same file more closely, for example [`SetAuthInfo`](https://github.com/ubuntu/authd-oidc-brokers/blob/dd8e0c086ea07c470dd6d0f04499db0369eebdfa/internal/broker/export_test.go#L97-L97) and [`SetAvailableMode`](https://github.com/ubuntu/authd-oidc-brokers/blob/dd8e0c086ea07c470dd6d0f04499db0369eebdfa/internal/broker/export_test.go#L111-L111).

internal/broker/broker.go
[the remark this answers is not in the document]
adombeck: I could live with it, but it feels wrong that `homeBaseDir` is part of the `broker.Config` struct and not of the `broker.Paths` struct (and the same applies for any other paths which we might want to make configurable in the config file at some point).

adombeck:
> Strange, the tests passed locally but in the CI the TestIsAuthenticated/Authenticating_with_password_refreshes_expired_token test fails. Marking as draft again while I look into that.

I finally figured it out. The groups listed in the golden files of this test were different in the CI and locally:

```
             	            	expected: testutils.treeAttrs{content:"access: granted\ndata: '{\"userinfo\":{\"name\":\"test-user@email.com\",\"uuid\":\"test-user-id\",\"dir\":\"/home/test-user@email.com\",\"shell\":\"/usr/bin/bash\",\"gecos\":\"test-user@email.com\",\"groups\":[{\"name\":\"remote-group\",\"ugid\":\"12345\"},{\"name\":\"linux-local-group\",\"ugid\":\"\"}]}}'\nerr: <nil>\n", path:"/first_call", executable:false}
            	            	actual  : testutils.treeAttrs{content:"access: granted\ndata: '{\"userinfo\":{\"name\":\"test-user@email.com\",\"uuid\":\"saved-user-id\",\"dir\":\"/home/test-user@email.com\",\"shell\":\"/usr/bin/bash\",\"gecos\":\"test-user@email.com\",\"groups\":[{\"name\":\"saved-remote-group\",\"ugid\":\"12345\"},{\"name\":\"saved-local-group\",\"ugid\":\"\"}]}}'\nerr: <nil>\n", path:"/first_call", executable:false}
```

The reason was 30d23aca2baedb29b3c0c60e87d8a300c0ff50b4, which made `IsAuthenticated` update the groups of the token when the session is not offline, regardless of whether the token is expired or not. Seems like `IsOffline` is true in the CI for some reason, so there the token was not updated with the groups from the mock provider. When running the test locally, `IsOffline` is false and the token is updated with the groups from the mock provider.

I fixed that by merging main into this branch, which includes the revert of 30d23aca2baedb29b3c0c60e87d8a300c0ff50b4.

internal/broker/broker.go
```diff
@@ -21,9 +20,12 @@ import (
 	"github.com/coreos/go-oidc/v3/oidc"
 	"github.com/google/uuid"
 	"github.com/ubuntu/authd-oidc-brokers/internal/consts"
+	"github.com/ubuntu/authd-oidc-brokers/internal/fileutils"
+	"github.com/ubuntu/authd-oidc-brokers/internal/password"
 	"github.com/ubuntu/authd-oidc-brokers/internal/providers"
 	providerErrors "github.com/ubuntu/authd-oidc-brokers/internal/providers/errors"
 	"github.com/ubuntu/authd-oidc-brokers/internal/providers/info"
+	tokenPkg "github.com/ubuntu/authd-oidc-brokers/internal/token"
```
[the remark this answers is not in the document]
adombeck: I pushed a commit which imports the package without an alias

internal/broker/broker.go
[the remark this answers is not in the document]
denisonbarbosa: I can see that. However, the `homeBaseDir` is related to the generated user info rather than the paths used by the broker itself. Maybe we can reflect this in the naming? e.g. `usersHomeBaseDir`

internal/broker/broker.go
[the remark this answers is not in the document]
denisonbarbosa: My issue with this is mainly that this is more of a C-like approach to a function than a Go one. It also depends on the decision we take on the previous comment regarding using different structs. But, like I've said, it's more straightforward and Go-like to have a function that takes the path as an argument and returns a concrete type with the expected content. You can see this type of difference when looking at `os.ReadFile`, for example. In C, it is common to allocate the buffer and then read something into it, but in Go we tend to return what was read instead.

I'm not sure if I was able to explain it as clear as possible. Did it make sense? We can talk on MM also if you're not satisfied.

internal/broker/broker.go
```diff
@@ -435,7 +457,7 @@ func (b *Broker) handleIsAuthenticated(ctx context.Context, session *sessionInfo
 		return AuthRetry, errorMessage{Message: "could not decode challenge"}
 	}
 
-	var authInfo authCachedInfo
+	var authInfo token.AuthCachedInfo
```
denisonbarbosa, line 460: Also, how do you feel about splitting this? It used to be the same type to make things easier+slightly faster when decrypting all stuff. Now that we don't have encryption, maybe we can split the token from the cached user information? i.e. having a user`.token` and a `user.info` file, for example... It would require a bit more refactoring in the code, so I'm down to just open a card for it and then we can tackle it in another PR.

internal/broker/export_test.go
denisonbarbosa: > That doesn't allow setting fields after initialization of the struct, as we do here though, right? 

It definitely doesn't. However, you could create a new object based on the current one (to set the field that you need) Ps.: This is also subject to what we decide regarding the config structs.

> Also, I have the impression that the setters I added follow the pattern of the other functions defined in the same file more closely, for example SetAuthInfo and SetAvailableMode.

In this case, the sets were due to having to alter an existing (complex) object (which, yeah, I know we do that currently for the config also [when looking at the example you've mentioned] but it could be avoided, no?).

internal/broker/helper_test.go
```diff
@@ -15,31 +17,32 @@ import (
...
 
 	b, err := broker.New(
 		cfg,
 		broker.WithSkipSignatureCheck(),
 		broker.WithCustomProviderInfo(&testutils.MockProviderInfoer{
 			Groups: []info.Group{
-				{Name: "remote-group", UGID: "12345"},
-				{Name: "linux-local-group", UGID: ""},
+				{Name: "new-broker-for-tests-remote-group", UGID: "12345"},
+				{Name: "linux-new-broker-for-tests-local-group", UGID: ""},
```
denisonbarbosa, line 45: Why do we need this?

internal/broker/testdata/TestIsAuthenticated/golden/authenticating_with_password_refreshes_expired_token/first_call
```diff
@@ -1,3 +1,3 @@
 access: granted
-data: '{"userinfo":{"name":"test-user@email.com","uuid":"test-user-id","dir":"/home/test-user@email.com","shell":"/usr/bin/bash","gecos":"test-user@email.com","groups":[{"name":"remote-group","ugid":"12345"},{"name":"linux-local-group","ugid":""}]}}'
+data: '{"userinfo":{"name":"test-user@email.com","uuid":"saved-user-id","dir":"/home/test-user@email.com","shell":"/usr/bin/bash","gecos":"test-user@email.com","groups":[{"name":"saved-remote-group","ugid":"12345"},{"name":"saved-local-group","ugid":""}]}}'
```
denisonbarbosa, line 2: Again a diff that shouldn't exist, no?

internal/token/migration_test.go
```diff
@@ -0,0 +1,170 @@
...
+		expectedRet bool
+		expectError bool
+	}{
+		"Password file exists": {
+			setup: func(t *testing.T, passwordPath, oldEncryptedTokenPath string) {
+				err := os.WriteFile(passwordPath, []byte("password"), 0600)
+				require.NoError(t, err, "WriteFile should not return an error")
+			},
+			expectedRet: false,
+		},
```
denisonbarbosa, line 30: It's better to make the test function more complex in order to keep the test case simple. So, here, it's better to use something like `wantOldEncryptedToken` and `wantPasswordFile` as controllers for the test case and then, inside the test body, properly handle the creation of these assets for the test itself. Does it make sense?

internal/token/token.go
denisonbarbosa: Sounds good. Also relates to my other comment about splitting the struct into two different ones.

internal/token/token_test.go
```diff
@@ -0,0 +1,135 @@
...
+		"Successfully load token from existing file": {
+			setup: func(t *testing.T, path string) {
+				err := os.MkdirAll(filepath.Dir(path), 0700)
+				require.NoError(t, err, "MkdirAll should not return an error")
+				err = token.CacheAuthInfo(path, testToken)
+				require.NoError(t, err, "CacheAuthInfo should not return an error")
+			},
+			expectedRet: testToken,
+			expectError: false,
+		},
```
denisonbarbosa, line 103: Same comment regarding test case complexity...

internal/fileutils/fileutils_test.go
adombeck: I fixed it one of the tests in this file but forgot the other ones. Fixed it in all of them now.

internal/broker/broker.go
adombeck: I also considered that but didn't do it because I think it might still simplify things to have the secrets in a single file (well, it's already two now with the password in a separate file) when we use use systemd credentials to protect them (see https://github.com/ubuntu/authd/issues/472).

I would prefer to go with what I proposed in https://github.com/ubuntu/authd-oidc-brokers/pull/134#discussion_r1757031321, i.e. use the term "token" for the whole JSON document we store persistently, which includes the OAuth2 token and the user info.

internal/broker/helper_test.go
adombeck: From the commit message:

```
Use different group names in different tests
    
To make it easier to figure out where in the code a group name comes
from.
```

I found that useful when I [debugged the failing test](https://github.com/ubuntu/authd-oidc-brokers/pull/134#issuecomment-2348571619).

denisonbarbosa: Ack. I thought the `saved-` prefix would be enough, but I see the reason of using a name that's more different to help with "mental overhead".

internal/broker/broker.go
denisonbarbosa: Sounds good to me!

internal/token/migration_test.go
adombeck: > It's better to make the test function more complex in order to keep the test case simple.

Why?

internal/broker/testdata/TestIsAuthenticated/golden/authenticating_with_password_refreshes_expired_token/first_call
adombeck: The change comes from the merge commit where I merged main to get the revert of 30d23aca2baedb29b3c0c60e87d8a300c0ff50b4. The tests fail if I undo the change. I want to do a git bisect to find out which change caused the tests to produce a different output. In order to do that, I'm rebasing on main to have a linear git history which I can bisect.

internal/token/migration_test.go
denisonbarbosa: Usually, the test cases are mainly to provide control values for the test function itself. In rare cases, they involve some programming logic, but most of the time we have just function arguments and some other control values that we need to set up the environment.

You could argue that adding the setup function to the test case makes the test function smaller and "easier" to debug (although that's a question of preference), but, in general, it's best to look at the test cases and see something small, concise and easy to get through whilst understanding what is the contract of that specific function (the one that's being tested). The setup to get the environment in a way that's ready for the test is a kind of implementation detail that it's better to keep inside the actual test itself.

Did it make sense? We can talk more during a HO if you're not convinced.

internal/broker/testdata/TestIsAuthenticated/golden/authenticating_with_password_refreshes_expired_token/first_call
adombeck: So the reason for the diff in the golden file is that `broker.loadAuthInfo` used to not load the `UserInfo` field from the cache, only the , and the `Token` and `RawIDToken` fields. The new `token.LoadAuthInfo` implementation also loads the `UserInfo` field. Do you know why the old implementation didn't set that field?

internal/broker/broker.go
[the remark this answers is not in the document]
adombeck: I extracted a `userConfig` struct which I embed in the `Config` struct in the latest commit I pushed (295c8713bbcdb0cb6bcb3ebee50841e4235c23ab). What do you think?

internal/broker/broker.go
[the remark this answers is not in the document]
adombeck: `parseConfigFile` now returns a `userConfig` struct.

internal/broker/export_test.go
adombeck: Since my change follows the pattern in that file, can we keep it as is in this PR and if we want to use a different pattern in that file, do that in a separate PR?

internal/token/token.go
adombeck: I would like to hear @didrocks opinion before I implement this, because it will change a lot of lines.  We might also want to do that in a separate PR, which we can review and merge quickly to avoid merge conflicts.

internal/token/migration_test.go
adombeck: I find that this pattern makes it very difficult to understand what exactly happens in a particular test case. A good example is the [`TestIsAuthenticated`](https://github.com/ubuntu/authd-oidc-brokers/blob/369-store-token-unencrypted/internal/broker/broker_test.go#L346-L346) test which has many test cases which set some fields that alter the behavior of the test. I frequently spend a lot of time debugging test cases of this test, which this pattern makes very hard in my experience. I think it's a downside of table testing in general, and this particular pattern to move all the test logic into a single test implementation in particular.

internal/token/token.go
didrocks: I think indeed, now that token grew into its own separate entity, this is really what we got from msentraid, and we a more utility base approach like Store/Load sounds good to me.

internal/broker/testdata/TestIsAuthenticated/golden/authenticating_with_password_refreshes_expired_token/first_call
denisonbarbosa: If we refresh the token, it means that user information could also have changed so we need to fetch it again. The only cases where we rely on the user information in the cache are:
1) Access token is still valid, so we don't refresh anything (it will change once we go back to handling this part)
2) We are in an offline session, so we can't fetch the user information from the provider.

internal/token/migration_test.go
denisonbarbosa: I understand your reasoning and agree that some tests do get quite complicated. The issue with the `TestIsAuthenticated` is that it is a unit test but also kind of an integration one (since it tests the notion of MFA we have). And we do have other complicated tests in ADSys also.

Still, I think we should keep the pattern that is used in the rest of the code for now, open up this for discussion, and then maybe reevaluate some of those tests. We should avoid having multiple ways of writing the tests and having them diverging between packages and/or projects.

internal/broker/broker.go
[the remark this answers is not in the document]
denisonbarbosa: Seems good to me. Do we gain anything from having it embedded into the Config struct?

internal/token/migration_test.go
adombeck: If we agree that the current pattern is making the tests complicated, then why don't we try out something else? I don't see why we have to use the same pattern in each test. Some patterns make sense in some case but not in another.

I changed the test cases of the `TestUseOldEncryptedToken` to set properties instead of defining a `setup` function now, because I think there it doesn't make the test more complicated. I kept the `setup` function in `TestLoadOldEncryptedAuthInfo`, `TestCacheAuthInfo`, and `TestLoadAuthInfo` because I think there it would make the tests more complicated.

internal/broker/testdata/TestIsAuthenticated/golden/authenticating_with_password_refreshes_expired_token/first_call
adombeck: I got rid of the diff by always refreshing the user info exactly then when we also refresh the token (by moving the `fetchUserInfo` call inside the `if !authInfo.Token.Valid() && !session.isOffline` block in https://github.com/ubuntu/authd-oidc-brokers/blob/aec153c9576b5f0367a883d8232b813d52d39368/internal/broker/broker.go#L557-L562). 

That's not the behavior we want, but we already track that in https://github.com/ubuntu/authd/issues/520, which we shouldn't fix in this PR.

adombeck:
I pushed some new commits which refactor the tests in preparation for the "refactor: Extract token package" commit, which now doesn't check `if authInfo.UserInfo.Name == ""` anymore to fetch the user info (see https://github.com/ubuntu/authd-oidc-brokers/pull/134#discussion_r1763554955). Please review these new commits:

* e0669adcb3ddf72936731f3f3034b505bf3d4541
* 8c07531ced37fbabe159d3b91ec35f56a7837d82

And commit 8dda69b55abf58405aa4e754456157f9bd24b1fa has a small diff compared to what you reviewed before:

```diff
diff --git a/internal/broker/broker.go b/internal/broker/broker.go
index b8ea8ae..3522e72 100644
--- a/internal/broker/broker.go
+++ b/internal/broker/broker.go
@@ -553,9 +553,8 @@ func (b *Broker) handleIsAuthenticated(ctx context.Context, session *sessionInfo
                        if ok {
                                authInfo.RawIDToken = refreshedIDToken
                        }
-               }
 
-               if authInfo.UserInfo.Name == "" {
+                       // Refresh the user info
                        authInfo.UserInfo, err = b.fetchUserInfo(ctx, session, &authInfo)
                        if err != nil {
                                slog.Error(err.Error())
```

internal/broker/broker.go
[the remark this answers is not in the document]
adombeck: > Do we gain anything from having it embedded into the Config struct?

The `broker` can access all of its config via `b.cfg`.

internal/broker/export_test.go
denisonbarbosa: Even though the names are similar, it is not the same pattern. The ones you've mentioned only alter the object instead of returning it (like you did here). That's what I meant by "builder pattern". If you need something like this, it's better to follow the convention and use the optional arguments instead (the whole `WithXX` syntax I pointed to).

I believe we have three ways of handling this in a Go-like way:

1) Create a new Config struct with the exposed fields in `export_test.go` and remap it to the new config file (the worse option);
2) Use the `SetXX`s, but as they are meant to be used: altering the config object itself without returning anything (the functions you've mentioned are named SetXX but are not Setters like the ones you did here) (good option and require less refactoring);
3) Use optional arguments like I've mentioned (IMO, best option although it will require more work to implement);

internal/fileutils/fileutils_test.go
```diff
@@ -0,0 +1,144 @@
...
+		expectError  bool
+	}{
+		"Returns true when file exists":                      {fileExists: true, expectExists: true},
+		"Returns false when file does not exist":             {fileExists: false, expectExists: false},
+		"Returns false when parent directory does not exist": {fileExists: false, expectExists: false},
+
+		"Error when parent directory is a file": {parentDirIsFile: true, expectError: true},
+	}
+
+	for _, tc := range tests {
```
denisonbarbosa, line 28: Nitpick: Remove space between the map and the loop (It's meant to represent a "single structure" of defining the tests and running them)

internal/fileutils/fileutils_test.go
```diff
@@ -0,0 +1,144 @@
...
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	"github.com/ubuntu/authd-oidc-brokers/internal/fileutils"
+)
+
+func TestFileExists(t *testing.T) {
+	tests := map[string]struct {
```
denisonbarbosa, line 13: We can make this parallel, no?

I won't comment on every single instance, so please take a look at the other tests also :wink:

internal/fileutils/fileutils_test.go
```diff
@@ -0,0 +1,144 @@
...
+				err := fileutils.Touch(filepath.Join(tempDir, "file"))
+				require.NoError(t, err, "Touch should not return an error")
+			}
+
+			exists, err := fileutils.FileExists(path)
+			if tc.expectError {
+				require.Error(t, err, "FileExists should return an error")
+			} else {
+				require.NoError(t, err, "FileExists should not return an error")
+			}
```
denisonbarbosa, line 47: As we talked about in the HO: return early if expecting the error instead of the else.

```suggestion
			if tc.expectError {
				require.Error(t, err, "FileExists should return an error")
				return
			} 
			require.NoError(t, err, "FileExists should not return an error")
```

internal/token/migration_test.go
```diff
@@ -0,0 +1,162 @@
...
+			}
+		})
+	}
+}
+
+func TestLoadOldEncryptedAuthInfo(t *testing.T) {
+	tests := map[string]struct {
+		setup       func(t *testing.T, path, password string)
+		password    string
+		expectedRet token.AuthCachedInfo
```
denisonbarbosa, line 61: Missing the whitespace between the input and the expected returns

internal/token/migration_test.go
```diff
@@ -0,0 +1,162 @@
...
+		"Successfully load old encrypted token": {
+			setup: func(t *testing.T, path, password string) {
+				// Encrypt the test token and write it to the file
+				jsonData, err := json.Marshal(testToken)
+				require.NoError(t, err, "Marshal should not return an error")
+				encryptedData, err := encrypt(jsonData, []byte(password))
+				require.NoError(t, err, "Encrypt should not return an error")
+				err = os.WriteFile(path, encryptedData, 0600)
+				require.NoError(t, err, "WriteFile should not return an error")
+			},
```
denisonbarbosa, line 73: I'll still be adamant about the setup being part of the test case. I'd rather add more complexity to the test than add it to the test case. As we talked about in the HO, the test cases should be easy and descriptive with as little logic attached as possible. 

You can see by looking at the three setups you have here that there are some repeated function calls, so we could write a separate helper function that does this instead of adding this logic to the test body (if you think it will reduce complexity).

internal/token/token.go
```diff
@@ -0,0 +1,53 @@
...
+	"fmt"
+	"os"
+	"path/filepath"
+
+	"github.com/ubuntu/authd-oidc-brokers/internal/providers/info"
+	"golang.org/x/oauth2"
+)
+
+// AuthCachedInfo represents the token that will be saved on disk for offline authentication.
+type AuthCachedInfo struct {
```
denisonbarbosa, line 15: The more I look at it, the more I hate myself for using this name. Now that we have it in its package, can you think of a better name? I was thinking about using `CacheEntry` or just `CachedInfo`.

denisonbarbosa (changes requested):
It's definitely look way better! Thanks for addressing all the comments and for the discussions, they were very productive. I do have a couple of other requests, but after those I think we'll be good for merging!

internal/broker/broker.go
```diff
@@ -102,6 +111,13 @@ type Option func(*option)
 func New(cfg Config, args ...Option) (b *Broker, err error) {
 	defer decorate.OnError(&err, "could not create broker")
 
+	if cfg.ConfigFile != "" {
+		cfg.userConfig, err = parseConfigFile(cfg.ConfigFile)
+		if err != nil {
+			return nil, fmt.Errorf("could not parse config: %v", err)
+		}
+	}
+
```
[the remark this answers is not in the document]
adombeck: From its name, I would expect the `parseConfigFile` function to return an error if there is nothing to parse. I think that we want to continue in that case though, so that we don't have to create and parse a config file in the tests. We could achieve that by returning a custom error in `parseConfigFile` and handling that by ignoring it, but the current approach is simpler and achieves the same.

internal/fileutils/fileutils_test.go
adombeck: done

internal/fileutils/fileutils_test.go
adombeck: good catch, fixed

internal/fileutils/fileutils_test.go
adombeck: done

internal/token/migration_test.go
adombeck: added

internal/token/migration_test.go
adombeck: done

internal/token/token.go
adombeck: My plan is to rename it to `Token` in a follow-up PR, see https://github.com/ubuntu/authd-oidc-brokers/pull/134#discussion_r1757031321

internal/broker/broker_test.go
```diff
@@ -615,39 +622,38 @@ func TestIsAuthenticated(t *testing.T) {
...
+		secondCallDelay int
 
 		timeBetween time.Duration
 	}{
-		"First auth starts and finishes before second":                  {timeBetween: 2 * time.Second},
-		"First auth starts first but second finishes first":             {firstUser: "user-timeout-3", timeBetween: time.Second},
-		"First auth starts first then second starts and first finishes": {firstUser: "user-timeout-2", secondUser: "user-timeout-3", timeBetween: time.Second},
+		"First auth starts and finishes before second":                  {secondCallDelay: 1, timeBetween: 2 * time.Second},
+		"First auth starts first but second finishes first":             {firstCallDelay: 3, timeBetween: time.Second},
+		"First auth starts first then second starts and first finishes": {firstCallDelay: 2, secondCallDelay: 3, timeBetween: time.Second},
```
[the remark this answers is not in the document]
adombeck: That's now finally done. I rebased the PR on https://github.com/ubuntu/authd-oidc-brokers/pull/171 and added another commit "refactor: Use fewer magic string constants in tests" to achieve that without making the tests more complicated.

Because i had to rebase anyway, I also squashed the fixup commits and resolved a lot of conflicts, both with main and https://github.com/ubuntu/authd-oidc-brokers/pull/171. I hope I didn't mess up any merge conflict resolution, but I gave my best to be thorough (e.g. `golangci-lint` and `go test ./internal/broker/...` pass on all commits).

adombeck:
I rebased on main and resolved the merge conflicts.

internal/token/migration.go
```diff
@@ -0,0 +1,140 @@
...
+// UseOldEncryptedToken checks if the password file or the old encrypted token file exists. It returns false if the
+// password file exists, true if only the old encrypted token file exists, and an error if neither exists.
+func UseOldEncryptedToken(passwordPath, oldEncryptedTokenPath string) (bool, error) {
+	exists, err := fileutils.FileExists(passwordPath)
+	if err != nil {
+		return false, fmt.Errorf("could not check password file: %w", err)
+	}
+	if exists {
+		return false, nil
+	}
```
adombeck, line 27: Note to self: This causes things to break if the password file exists but the token does not.

internal/broker/testdata/TestConcurrentIsAuthenticated/golden/first_auth_starts_and_finishes_before_second/second_auth
```diff
@@ -1,3 +1,3 @@
 access: granted
-data: '{"userinfo":{"name":"user-timeout-1","uuid":"saved-user-id","dir":"/home/user-timeout-1","shell":"/usr/bin/bash","gecos":"user-timeout-1","groups":[{"name":"remote-group","ugid":"12345"},{"name":"linux-local-group","ugid":""}]}}'
+data: '{"userinfo":{"name":"test-user@email.com","uuid":"saved-user-id","dir":"/home/test-user@email.com","shell":"/usr/bin/bash","gecos":"test-user@email.com","groups":[{"name":"saved-remote-group","ugid":"12345"},{"name":"saved-local-group","ugid":""}]}}'
```
denisonbarbosa: Although it might be following the right order in the code, the golden files do not mirror the expected behavior (i.e. we are getting the same user info for both calls, even though we used different usernames), so we need to try and make it resemble more the real scenario.

internal/broker/testdata/TestIsAuthenticated/golden/authenticating_with_password_refreshes_expired_token/first_call
```diff
@@ -1,3 +1,3 @@
 access: granted
-data: '{"userinfo":{"name":"test-user@email.com","uuid":"test-user-id","dir":"/home/test-user@email.com","shell":"/usr/bin/bash","gecos":"test-user@email.com","groups":[{"name":"remote-group","ugid":"12345"},{"name":"linux-local-group","ugid":""}]}}'
+data: '{"userinfo":{"name":"test-user@email.com","uuid":"saved-user-id","dir":"/home/test-user@email.com","shell":"/usr/bin/bash","gecos":"test-user@email.com","groups":[{"name":"remote-group","ugid":"12345"},{"name":"linux-local-group","ugid":""}]}}'
```
denisonbarbosa: Is this diff expected?

internal/token/token_test.go
```diff
@@ -0,0 +1,143 @@
...
+				err := os.MkdirAll(filepath.Dir(path), 0700)
+				require.NoError(t, err, "MkdirAll should not return an error")
+			},
+		},
+		"Successfully store token with existing file": {
+			setup: func(t *testing.T, path string) {
+				err := os.MkdirAll(filepath.Dir(path), 0700)
+				require.NoError(t, err, "MkdirAll should not return an error")
+				err = token.CacheAuthInfo(path, testToken)
+				require.NoError(t, err, "CacheAuthInfo should not return an error")
```
denisonbarbosa, line 56: We still have the setup functions in the test cases...

internal/token/token_test.go
```diff
@@ -0,0 +1,143 @@
...
+		},
+	},
+}
+
+func TestCacheAuthInfo(t *testing.T) {
+	t.Parallel()
+
+	tests := map[string]struct {
+		setup       func(t *testing.T, path string)
+		expectError bool
```
denisonbarbosa, line 38: s/expect/want

Let's keep the wording the same as we have for all of our projects.

denisonbarbosa (changes requested):
Some new comments, some old ones... Still, the changes are looking great. Congratulations on the nice work!

internal/broker/testdata/TestConcurrentIsAuthenticated/golden/first_auth_starts_and_finishes_before_second/second_auth
adombeck: I'm not sure how this happened, but it seems like your comment is made on an outdated diff. On commit 7cfda81 (pushed two weeks ago), I see the following diff (which contains different usernames in the response of the two calls):

```diff
diff --git a/internal/broker/testdata/TestConcurrentIsAuthenticated/golden/first_auth_starts_and_finishes_before_second/first_auth b/internal/broker/testdata/TestConcurrentIsAuthenticated/golden/first_auth_starts_and_finishes_before_second/first_auth
index bd6241b..13dbd87 100644
--- a/internal/broker/testdata/TestConcurrentIsAuthenticated/golden/first_auth_starts_and_finishes_before_second/first_auth
+++ b/internal/broker/testdata/TestConcurrentIsAuthenticated/golden/first_auth_starts_and_finishes_before_second/first_auth
@@ -1,3 +1,3 @@
 access: granted
-data: '{"userinfo":{"name":"user-timeout-0","uuid":"saved-user-id","dir":"/home/user-timeout-0","shell":"/usr/bin/bash","gecos":"user-timeout-0","groups":[{"name":"remote-group","ugid":"12345"},{"name":"linux-local-group","ugid":""}]}}'
+data: '{"userinfo":{"name":"user1","uuid":"saved-user-id","dir":"/home/user1","shell":"/usr/bin/bash","gecos":"user1","groups":[{"name":"new-broker-for-tests-remote-group","ugid":"12345"},{"name":"linux-new-broker-for-tests-local-group","ugid":""}]}}'
 err: <nil>
diff --git a/internal/broker/testdata/TestConcurrentIsAuthenticated/golden/first_auth_starts_and_finishes_before_second/second_auth b/internal/broker/testdata/TestConcurrentIsAuthenticated/golden/first_auth_starts_and_finishes_before_second/second_auth
index 1fee6db..c0b252f 100644
--- a/internal/broker/testdata/TestConcurrentIsAuthenticated/golden/first_auth_starts_and_finishes_before_second/second_auth
+++ b/internal/broker/testdata/TestConcurrentIsAuthenticated/golden/first_auth_starts_and_finishes_before_second/second_auth
@@ -1,3 +1,3 @@
 access: granted
-data: '{"userinfo":{"name":"user-timeout-1","uuid":"saved-user-id","dir":"/home/user-timeout-1","shell":"/usr/bin/bash","gecos":"user-timeout-1","groups":[{"name":"remote-group","ugid":"12345"},{"name":"linux-local-group","ugid":""}]}}'
+data: '{"userinfo":{"name":"user2","uuid":"saved-user-id","dir":"/home/user2","shell":"/usr/bin/bash","gecos":"user2","groups":[{"name":"new-broker-for-tests-remote-group","ugid":"12345"},{"name":"linux-new-broker-for-tests-local-group","ugid":""}]}}'
 err: <nil>
```

internal/providers/msentraid/msentraid.go
```diff
@@ -209,7 +209,12 @@ func getAllUserGroups(client *msgraphsdk.GraphServiceClient) ([]msgraphmodels.Gr
 		groups = append(groups, result.GetValue()...)
 	}
 
-	slog.Debug(fmt.Sprintf("Got groups: %v", groups))
+	var groupNames []string
+	for _, group := range groups {
+		groupNames = append(groupNames, *group.GetDisplayName())
```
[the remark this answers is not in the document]
adombeck: right, fixed

internal/token/token_test.go
adombeck: I changed that now

internal/token/token_test.go
adombeck: done

internal/broker/testdata/TestConcurrentIsAuthenticated/golden/first_auth_starts_and_finishes_before_second/second_auth
denisonbarbosa: Yeah, it was outdated indeed. Not sure why I was seeing the outdated version, but now I can see that it's correct.

internal/token/token_test.go
denisonbarbosa: Well, you fixed that on one of the tests... But the other still have them

internal/token/token_test.go
denisonbarbosa: Same thing for "fixed for one test, but others still have the same behavior"

internal/broker/testdata/TestIsAuthenticated/golden/authenticating_with_password_refreshes_expired_token/first_call
denisonbarbosa: I was expecting the UUID to be updated alongside the token, as it was previously, or am I missing some change regarding the way we create the cached token/the token we return from the mock?

internal/token/token_test.go
adombeck: Oops. I think I updated them all now.

internal/token/token_test.go
adombeck: Should now all be updated

internal/broker/testdata/TestIsAuthenticated/golden/authenticating_with_password_refreshes_expired_token/first_call
adombeck: That's outdated now, one of the changes I did yesterday changed the diff to:

```diff
diff --git a/internal/broker/testdata/TestIsAuthenticated/golden/authenticating_with_password_refreshes_expired_token/first_call b/internal/broker/testdata/TestIsAuthenticated/golden/authenticating_with_password_refreshes_expired_token/first_call
index 3c4c96b..12ca514 100644
--- a/internal/broker/testdata/TestIsAuthenticated/golden/authenticating_with_password_refreshes_expired_token/first_call
+++ b/internal/broker/testdata/TestIsAuthenticated/golden/authenticating_with_password_refreshes_expired_token/first_call
@@ -1,3 +1,3 @@
 access: granted
-data: '{"userinfo":{"name":"test-user@email.com","uuid":"test-user-id","dir":"/home/test-user@email.com","shell":"/usr/bin/bash","gecos":"test-user@email.com","groups":[{"name":"remote-group","ugid":"12345"},{"name":"linux-local-group","ugid":""}]}}'
+data: '{"userinfo":{"name":"test-user@email.com","uuid":"test-user-id","dir":"/home/test-user@email.com","shell":"/usr/bin/bash","gecos":"test-user@email.com","groups":[{"name":"new-broker-for-tests-remote-group","ugid":"12345"},{"name":"linux-new-broker-for-tests-local-group","ugid":""}]}}'
 err: <nil>
```

internal/token/migration.go
adombeck: I fixed that yesterday

internal/token/migration_test.go
```diff
@@ -0,0 +1,161 @@
...
+				require.NoError(t, err, "WriteFile should not return an error")
+			}
+
+			got, err := token.UseOldEncryptedToken(passwordPath, tokenPath, oldEncryptedTokenPath)
+			if tc.wantError {
+				require.Error(t, err, "UseOldEncryptedToken should return an error")
+			} else {
+				require.NoError(t, err, "UseOldEncryptedToken should not return an error")
+				require.Equal(t, tc.wantRet, got, "UseOldEncryptedToken should return the expected value")
+			}
```
denisonbarbosa, line 64: I'm pretty sure we have addressed this before, so I'm not sure if got caught on a rebase or something, but we prefer returning early instead of an else here in these cases (it happens on other tests as well, so take a closer look at the other cases).

internal/fileutils/fileutils_test.go
```diff
@@ -0,0 +1,156 @@
...
+				err := fileutils.Touch(filepath.Join(tempDir, "dir", "file"))
+				require.NoError(t, err, "Touch should not return an error")
+			}
+			if tc.isFile {
+				path = filepath.Join(tempDir, "file")
+				err := fileutils.Touch(path)
+				require.NoError(t, err, "Touch should not return an error")
+			}
+
+			empty, err := fileutils.IsDirEmpty(path)
```
denisonbarbosa, line 96: I still think we are missing an opportunity here. Not sure if you had the time to take a look at the "Go Proverbs" video that I sent you, but one of the things he talks about is that "errors are values". It's very unusual to have to check multiple values to fully understand the output of the function.

In this case, it's quite easy to update this since the functions are part of a `fileutils` package. We should define some errors there, e.g. `fileutils.EmptyDirError` or `fileutils.NotEmptyDirError` (depends on what you want the "happy" case to be), and return those as the `false` value for the `IsDirEmpty` function (which should probably be renamed if we remove the bool part of its return).

denisonbarbosa (changes requested):
Almost there... There's still one thing regarding the tests and a design improvement suggestion.

internal/fileutils/fileutils_test.go
denisonbarbosa: Since this PR is quite long already, let's think about this later and then we do a refactoring of it.

adombeck: I'm not sure I understand your request. Do you want me to change the function from:

```go
func IsDirEmpty(path string) (bool, error)
```

to

```go
func CheckDirEmpty(path string) error
```
?

and then instead of calling it like this:

```go
exists, err := fileutils.IsDirEmpty(path)
if err != nil {
    return err
}
// Do something with `empty`
```

to call it like this

```go
err := CheckDirEmpty(path)
exists := true
if errors.Is(err, fileutils.EmptyDirError) {
    exists = false
} else if err != nil {
   return err
}
```
?

If so, I think that would make things more complicated for the caller (having to check for different error types). I think here it's better to only return an error if something unexpected happened and use a second, non-error return value which the caller can work with in the non-error case.

internal/token/migration_test.go
adombeck: done

internal/fileutils/fileutils_test.go
denisonbarbosa: That's not complicated for the caller, tbh. In some types of function, it's kind of expected (look at the `os` and `fs` packages and also at out cache package [in authd]). Errors in Go not necessarily mean that something "broke" or "went wrong", it's more in the sense of "the return is not what you expected".

denisonbarbosa (approved):
Looks good to me now! Thanks for addressing the comments and for the amazing work.

internal/fileutils/fileutils_test.go
adombeck: I disagree but I don't see the need to continue this discussion now.

PR #135, opened 2024-09-05
Fix permission error when listing groups of a guest account

adombeck:
When the user account is a guest account on the tenant, the existing permissions are not enough to list all the memberships, so the `TransitiveMemberOf().Get()` call results in a permission denied error.

The existing permissions are enough to list only the groups though, and that's the only thing we're interested in.


internal/providers/msentraid/msentraid.go
```diff
@@ -111,6 +117,15 @@ func (p Provider) userClaims(idToken *oidc.IDToken) (claims, error) {
 
 // getGroups access the Microsoft Graph API to get the groups the user is a member of.
 func (p Provider) getGroups(token *oauth2.Token) ([]info.Group, error) {
+	// Check if the token has the GroupMember.Read.All scope
+	scopes, err := p.getTokenScopes(token)
+	if err != nil {
+		return nil, err
+	}
+	if !slices.Contains(scopes, "GroupMember.Read.All") {
+		return nil, errors.New("app is missing the GroupMember.Read.All permission")
```
denisonbarbosa, line 126: IMO, `app` does not say much here. Can we be a bit clearer regarding what's missing the permission?

denisonbarbosa (changes requested):
Only a small comment. However, I see that the test went from a success case to an error one and we have no changes in the golden file, so there's something wrong with that specific test. Would you mind checking it?

adombeck:
> I see that the test went from a success case to an error one and we have no changes in the golden file, so there's something wrong with that specific test. Would you mind checking it?

Right, good catch! So the part of the code where we now return an error if the `GroupMember.Read.All` permission is missing is actually not tested by the broker tests, because it's part of the msentraid provider which we mock in that test. Now I'm not sure what to do with the `Authenticating still allowed if token is missing scopes` broker test. The broker code itself doesn't deny access when scopes are missing, but the only provider implementation we use in production does, so the "story" which the test tells is misleading.

The `getGroups` method which now returns an error if the scope is missing is called by `GetUserInfo` method of the `msentraid` package. We don't have tests for that yet, and I'm not sure how practical it is to add one, because we would need an `oauth2.Token` for it.

adombeck:
@didrocks any thoughts on https://github.com/ubuntu/authd-oidc-brokers/pull/135#issuecomment-2333853239?

internal/providers/msentraid/msentraid.go
adombeck: How about `The Microsoft Entra ID app is missing the GroupMember.Read.All permission`?

adombeck: Also, if we want to display this error to the user instead of just logging it, we need to implement special case handling for it.

adombeck: > Also, if we want to display this error to the user instead of just logging it, we need to implement special case handling for it.

Done in https://github.com/ubuntu/authd-oidc-brokers/pull/135/commits/0be47303180b16b239c805cdd0376179bd00d0ad

denisonbarbosa:
> > I see that the test went from a success case to an error one and we have no changes in the golden file, so there's something wrong with that specific test. Would you mind checking it?
> 
> Right, good catch! So the part of the code where we now return an error if the `GroupMember.Read.All` permission is missing is actually not tested by the broker tests, because it's part of the msentraid provider which we mock in that test. Now I'm not sure what to do with the `Authenticating still allowed if token is missing scopes` broker test. The broker code itself doesn't deny access when scopes are missing, but the only provider implementation we use in production does, so the "story" which the test tells is misleading.

That's not exactly true. The test is currently named "Error when token is missing GroupMember.Read.All scope" but this can easily be renamed to "Error when token is missing required scope" and be implemented generically (we have a mock for the providers, after all). I know it might feel like you'd be "testing the mock" but that's not true. We want to have this behavior registered in a test case and that will be useful when implementing the e2e tests regarding MSEntraID and so on (to understand our baseline of tests and build upon it). 

It looks like we have required scopes (like the one mentioned) and others that are not that important. So, IMO, the tests should indicate that and the mock should follow this behavior also (i.e. have a test that fails when the token is missing a required scope and one that passes when the token is missing an optional one). 

> The `getGroups` method which now returns an error if the scope is missing is called by `GetUserInfo` method of the `msentraid` package. We don't have tests for that yet, and I'm not sure how practical it is to add one, because we would need an `oauth2.Token` for it.

That's exactly why we need to keep the mock as close as possible to the actual thing, especially due to the lack of e2e testing (which will be fixed soon, hopefully). And that implies adding the mentioned behavior regarding required and optional scopes.

adombeck:
> That's not exactly true. The test is currently named "Error when token is missing GroupMember.Read.All scope" but this can easily be renamed to "Error when token is missing required scope" and be implemented generically (we have a mock for the providers, after all). I know it might feel like you'd be "testing the mock" but that's not true. We want to have this behavior registered in a test case and that will be useful when implementing the e2e tests regarding MSEntraID and so on (to understand our baseline of tests and build upon it).
> 
> It looks like we have required scopes (like the one mentioned) and others that are not that important. So, IMO, the tests should indicate that and the mock should follow this behavior also (i.e. have a test that fails when the token is missing a required scope and one that passes when the token is missing an optional one).

So even though the broker handles errors returned by `CheckTokenScopes` by printing a warning and proceeding, you want to assume that the implementation of `GetUserInfo` returns an error if any scopes are missing, and change the mock provider to also do that? I'm not sure that we can make that assumption, and include that in the (implicit) API contract by adding a test for it.

adombeck:
> you want to assume that the implementation of GetUserInfo returns an error if any scopes are missing

That's not even the case for the msentraid broker, there we only return if the `GroupMember.Read.All` scope is missing, but not if any of the other scopes are missing.

didrocks:
> @didrocks I would like to also hear your thoughts on https://github.com/ubuntu/authd-oidc-brokers/pull/135#issuecomment-2333853239

Just a note: apart from my couple of comments on improving the error I didn’t review this, just focused on the part you mentioned.

I’m unsure I’m understanding this looking at the global diff: we have somes golden files for a test that doesn’t exist. Is that intended?
On what I’m grasping from the discussion here, we have a test that relies on MSEntraID specific implementation and we shouldn’t, right? I agree with that idea. From the "generic broker" perspective, we are failing because `GetUserInfo()` is failing, and this is what matters.

We should have tests for MSEntraID talking about the scopes, but those are MSEntraID specific, and it’s something we should work on now that we have base tests on it. Like, giving it a token (first valid, then with missing scopes…) and having tests asserting on behaviour regarding that token.

The complex part here will be ofc to mock the msgraph calls. I guess this is why this wasn’t done in the first implementation, but I feel that once we make a public announcement and fix the existing bugs, that will be a good opportunity (alongside end to end tests) to tackle those. For reference, I did mock some msgraph part on aad-auth.

I hope that I understood the request and context correctly :)

internal/broker/broker.go
```diff
@@ -735,3 +736,13 @@ func decorateErrorMessage(data *isAuthenticatedDataResponse, msg string) {
 	errMsg.Message = fmt.Sprintf("%s: %s", msg, errMsg.Message)
 	*data = errMsg
 }
+
+// Checks if the provided error is of type ForDisplayError. If it is, it returns the error message. Else, it returns
+// the provided fallback message.
+func errorMessageForDisplay(err error, fallback string) errorMessage {
```
didrocks, line 742: Just a note while I am on this: I think we should take this (as explained on the uncrypted token PR) as a way to logging in the initial error and returning the error we want to display to the user (probably on the defer of each dbus calls). I wonder if all those redacted messages are always coming from the provider? If so, it makes sense to have them as a subpackage of providers), otherwise, we can move it directly under `internal/`. To not conflict with the "errors" I would just name it to something similar you imported it with: "providerserrors" or even directly "errorfordisplay" as this is exactly what's its scope is.

We can go on this intermediate step and fix this in the token PR too if you prefer.

internal/providers/errors/errors.go
```diff
@@ -0,0 +1,17 @@
...
+// ForDisplayError is an error type for errors that are meant to be displayed to the user.
+type ForDisplayError struct {
+	message string
+}
+
+func (e *ForDisplayError) Error() string {
+	return e.message
+}
+
+func NewForDisplayError(format string, v ...interface{}) *ForDisplayError {
```
didrocks, line 15: Also, do not return a pointer unless we really need it.

internal/broker/testdata/TestIsAuthenticated/golden/error_when_token_is_missing_the_groupmember.read.all_scope/cache/provider_url/test-user@email.com.cache
denisonbarbosa, line 1: Ps.: Don't forget to remove this leftover golden file

internal/broker/testdata/TestIsAuthenticated/golden/error_when_token_is_missing_the_groupmember.read.all_scope/first_call
denisonbarbosa, line 1: Ps.: Don't forget to remove this leftover golden file

internal/broker/testdata/TestIsAuthenticated/golden/error_when_token_is_missing_the_groupmember.read.all_scope/second_call
denisonbarbosa, line 1: Ps.: Don't forget to remove this leftover golden file

denisonbarbosa (approved):
Regarding our previous discussion: You are right regarding the test, let's drop it and wait until we have e2e tests to check for that.

On the error redacting, I agree with Didier regarding the package itself. Maybe we could move it to its own `internal/` package (also, we should have another name to avoid conflicts with the actual `errors` standard package) and reuse that throughout the code? That would imply another refactoring regarding the way we handle what error is logged and what is printed, but maybe it's for the best (?)

I'll approve the PR on my side and let you follow with the discussion with @didrocks regarding the error handling. I just have a couple of reminders for you to remove some leftover golden files from the previous test that was removed.

Nice work!

internal/providers/errors/errors.go
adombeck: done

internal/broker/testdata/TestIsAuthenticated/golden/error_when_token_is_missing_the_groupmember.read.all_scope/cache/provider_url/test-user@email.com.cache
adombeck: done

internal/broker/testdata/TestIsAuthenticated/golden/error_when_token_is_missing_the_groupmember.read.all_scope/first_call
adombeck: done

internal/broker/broker.go
didrocks: As told on the error one, I’m happy to give a pass on that one, and I’ll refactor it later this pulse, hoping I’m getting time for it.

internal/providers/msentraid/msentraid.go
```diff
@@ -111,6 +118,15 @@ func (p Provider) userClaims(idToken *oidc.IDToken) (claims, error) {
 
 // getGroups access the Microsoft Graph API to get the groups the user is a member of.
 func (p Provider) getGroups(token *oauth2.Token) ([]info.Group, error) {
+	// Check if the token has the GroupMember.Read.All scope
+	scopes, err := p.getTokenScopes(token)
+	if err != nil {
+		return nil, err
+	}
+	if !slices.Contains(scopes, "GroupMember.Read.All") {
+		return nil, providerErrors.NewForDisplayError(("the Microsoft Entra ID app is missing the GroupMember.Read.All permission"))
```
didrocks, line 127: nitpick: double parentheses

didrocks (approved):
Given that the "done" items are pushed and fixed, I only have one nitpick and you can merge afterwards.

PR #142, opened 2024-09-09
Avoid false positive nolintlint error

adombeck:
When running golangci-lint locally, the "nolintlint" linter complained
that the nolint:revive directive is unused:

	directive `//nolint:revive // DaemonConfig is a type alias for tests` is unused for linter "revive" (nolintlint)

When removing that directive, the "revive" linter complains in the
GitHub Action:

	unexported-return: exported method Config returns unexported type daemon.daemonConfig, which can be annoying to use (revive)

This commit disables both "revive" and "nolintlint" to make
golangci-lint pass both in the CI and locally.

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -8,7 +8,7 @@ linters:
     - durationcheck
     - errname
     - errorlint
-    - exportloopref
+    - copyloopvar
     - forbidigo
     - forcetypeassert
     - gci
--- a/cmd/authd-oidc/daemon/export_test.go
+++ b/cmd/authd-oidc/daemon/export_test.go
@@ -76,7 +76,7 @@ client_id = client_id
 
 // Config returns a DaemonConfig for tests.
 //
-//nolint:revive // DaemonConfig is a type alias for tests
+//nolint:revive,nolintlint // DaemonConfig is a type alias for tests
 func (a App) Config() DaemonConfig {
 	return a.config
 }


denisonbarbosa (changes requested):
I'm not a big fan of this kind of change due to differences between local and GitHub. Do you know why this is happening? I don't have this behavior on my side.

adombeck:
Can we please merge this? It's a constant source of frustration in my workflow.

denisonbarbosa:
> Can we please merge this? It's a constant source of frustration in my workflow.

AFAIK, this is only failing on your side. So it's more a question of trying to understand why the behavior is different than merging this. Otherwise, it opens up precedence for more of this to happen again and it's a behavior that should be avoided.

adombeck:
> So it's more a question of trying to understand why the behavior is different than merging this.

I already spent more time on trying to understand it than I think is reasonable. The fix is small and we applied it in other places as well (as mentioned in https://github.com/ubuntu/authd-oidc-brokers/pull/142#issuecomment-2340574888). Lets use our time efficiently please.

jibel:
my 2c

I can reproduce the problem locally golangci-lint 1.60.3

```
$ golangci-lint --version
golangci-lint has version v1.60.3 built with go1.23.0 from (unknown, modified: ?, mod sum: "h1:l38A5de24ZeDlcFF+EB7m3W5joPD99/hS5SIHJPyZa0=") on (unknown)
$ golangci-lint run --enable nolintlint --tests
WARN The linter 'exportloopref' is deprecated (since v1.60.2) due to: Since Go1.22 (loopvar) this linter is no longer relevant. Replaced by copyloopvar.
cmd/authd-oidc/daemon/export_test.go:79:1: directive `//nolint:revive // DaemonConfig is a type alias for tests` is unused for linter "revive" (nolintlint)
//nolint:revive // DaemonConfig is a type alias for tests
^
```

adombeck:
Updating `golangci-lint` to 1.61.0 fixed the issue

PR #144, opened 2024-09-10
Update user info even on local authentication

denisonbarbosa:
We used to rely on the cached information if the access token was still valid. Now, we always try to fetch the user information (username, groups, etc) if the session is online, even if we did not refresh the token.

UDENG-4436

### DIFF
--- a/internal/broker/broker.go
+++ b/internal/broker/broker.go
@@ -482,7 +482,7 @@ func (b *Broker) handleIsAuthenticated(ctx context.Context, session *sessionInfo
 			return AuthRetry, errorMessage{Message: "could not load cached info"}
 		}
 
-		if authInfo.UserInfo.Name == "" {
+		if !session.isOffline {
 			authInfo.UserInfo, err = b.fetchUserInfo(ctx, session, &authInfo)
 			if err != nil {
 				slog.Error(err.Error())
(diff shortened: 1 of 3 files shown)


didrocks (approved):
Ack on the change and golden file change. From what I understand, we are going to fail if we were online (so got a token, but didn’t refresh it), and groups retrieval fails, is this correct? That behaviour makes sense to me.

internal/broker/broker.go
```diff
@@ -482,7 +482,7 @@ func (b *Broker) handleIsAuthenticated(ctx context.Context, session *sessionInfo
 			return AuthRetry, errorMessage{Message: "could not load cached info"}
 		}
 
-		if authInfo.UserInfo.Name == "" {
+		if !session.isOffline {
 			authInfo.UserInfo, err = b.fetchUserInfo(ctx, session, &authInfo)
 			if err != nil {
 				slog.Error(err.Error())
```
adombeck, line 488: This has the effect that the we deny authentication when something goes wrong during refresh, even if the local token is still valid. I think it would be better to log the error and proceed with the login in that case.

denisonbarbosa: That's intended. If we want to refresh the user information every authentication, we need to fail if something goes wrong. IMO, we should not have this kind of nondeterministic behavior, especially when considering PAM and the user information on the system.

PR #149, opened 2024-09-13
Revert "Check that usernames comply with the Microsoft Entra username policy"

adombeck:
Commit 4b7f8f35d89ba2a9720a691f133a323015c9c827 broke device authentication with Microsoft Entra because it doesn't allow "@" in the username, but all Microsoft Entra user principal names (and emails) include an "@".

This reverts commit 4b7f8f35d89ba2a9720a691f133a323015c9c827.

UDENG-4563


didrocks:
Ok, let’s go with it.

didrocks (approved):

PR #151, opened 2024-09-13
Reapply and fix "Check that usernames comply with the Microsoft Entra username policy"

adombeck:
UDENG-4563

### DIFF
--- a/internal/providers/msentraid/msentraid_test.go
+++ b/internal/providers/msentraid/msentraid_test.go
@@ -61,15 +61,15 @@ func TestVerifyUsername(t *testing.T) {
 
 		wantErr bool
 	}{
-		"Success when usernames are the same":   {requestedUsername: "foo@bar", authenticatedUser: "foo@bar"},
-		"Success when usernames differ in case": {requestedUsername: "foo@bar", authenticatedUser: "Foo@bar"},
+		"Success when usernames are the same":   {requestedUsername: "foo-bar@example", authenticatedUser: "foo-bar@example"},
+		"Success when usernames differ in case": {requestedUsername: "foo-bar@example", authenticatedUser: "Foo-Bar@example"},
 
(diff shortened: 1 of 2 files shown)


internal/providers/msentraid/msentraid_test.go
```diff
@@ -61,10 +61,16 @@ func TestVerifyUsername(t *testing.T) {
 
 		wantErr bool
 	}{
-		"Success when usernames are the same":   {requestedUsername: "foo@bar", authenticatedUser: "foo@bar"},
-		"Success when usernames differ in case": {requestedUsername: "foo@bar", authenticatedUser: "Foo@bar"},
-
-		"Error when usernames differ": {requestedUsername: "foo@bar", authenticatedUser: "bar@foo", wantErr: true},
+		"Success when usernames are the same":   {requestedUsername: "foo-bar@example", authenticatedUser: "foo-bar@example"},
```
denisonbarbosa, line 64: I'm a bit surprised that we didn't get the bug with `@` in the username before. Do you know what happened?

denisonbarbosa (commented):
Looks good to me. But I have a question regarding the tests...

internal/providers/msentraid/msentraid_test.go
adombeck: The `.-_` in the regex was interpreted as "characters ranging from `.` to `_` which I assume must include `@` but not `-` (I didn't not check but that's the only explanation for the behavior we saw), so usernames which included a `@` were actually allowed but usernames which included a `-` were not.

denisonbarbosa: Oh, I see. So the issue was the `-` not the `@` as it was described in the revert PR?

adombeck: Well, both. The `@` was missing in the allowed characters. It was still allowed by accident because the `-` was not escaped so it was interpreted as a character range.

denisonbarbosa (approved):
Thanks for fixing it and answering the comments. Nice work!

PR #161, opened 2024-09-18
generate snap version from current git tree state

didrocks:
We are starting to generate the snap version using the following rules: When considering a tag, if starting with "<current branch name>-" it will have its prefix removed.
For instance:
* 0.1 -> 0.1
* msentraid-0.1 -> 0.1 on msentraid branch.

1. If current commit is tagged, the version is directly the tag name.
2. If current commit is not tagged, the version is:
    a. <last tag name on current branch>+<commit_sha> for main branch.
    b. <last tag name on current branch (not in main)>+<commit_sha>.<last_commit_merged_from_main> for other branches.

Any of those version will be annoted with +dirty if there are local changes.

UDENG-4431

### DIFF
--- a/snap/get_version
+++ b/snap/get_version
@@ -1,0 +1,91 @@
+#!/bin/sh
+set -eu
+
+# Script to build the version to set to the snap.
+
+# When considering a tag, if starting with "<current branch name>-" it will have its prefix removed.
+# For instance:
+# * 0.1 -> 0.1
+# * msentraid-0.1 -> 0.1 on msentraid branch.
+
+# 1. If current commit is tagged, the version is directly the tag name.
+# 2. If current commit is not tagged, the version is:
+#    a. <last tag name on current branch>+<commit_sha> for main branch.
+#    b. <last tag name on current branch (not in main)>+<commit_sha>.<last_commit_merged_from_main> for other branches.
+#
+# Any of those version will be annoted with +dirty if there are local changes.
+
+# set_version will markup the version in the snapcraft.yaml file after amending it with a dirty markup if necessary.
+# $1: version: the version to set.
+set_version() {
+    version="${1}"
+
+    version=$(annotate_with_dirty "${version}")
+    craftctl set version="${version}"
+}
+
+# annotate_with_dirty may amend the version with a dirty markup if there are local changes.
+# $1: version: the version to annotate.
+annotate_with_dirty() {
+    version="${1}"
+
+    # check if current tree content is dirty.
+    is_dirty=$(git -C "${SNAPCRAFT_PART_SRC}" status --porcelain)
+    if [ -n "${is_dirty}" ]; then
+        version="${version}+dirty"
+    fi
+
+    echo "${version}"
+}
+
+# strip_branch_tag_prefix will remove the branch name prefix from the tag name.
+# $1: tag: the tag name to strip the prefix from.
+# $2: current_branch: the branch name to strip from the tag.
+strip_branch_tag_prefix() {
+    tag="${1}"
+    current_branch="${2}"
+
+    echo "${tag#"${current_branch}-"}"
+}
+
+
+current_branch=$(git -C "${SNAPCRAFT_PART_SRC}" branch --show-current)
+
+# Try to get most recent tag on that branch not coming from main.
+# Main will just get the most recent tag merged into it.
+tag_cmd_suffix=""
+if [ "${current_branch}" != "main" ]; then
+    tag_cmd_suffix="--no-merged=main"
+fi
+
+# Get most recent tag on that branch not coming from the other branch.
+tag=$(git tag --sort=-v:refname --merged="${current_branch}" ${tag_cmd_suffix} | head -1)
+
+version="${tag}"
+if [ -z "${version}" ]; then
+    # No tag found, use "notag" as version.
+    version="notag"
+fi
+version=$(strip_branch_tag_prefix "${version}" "${current_branch}")
+
+# If the most recent tag is on the current commit, taking it as is once transformed as a version.
+if [ -n "${tag}" ] && [ "$(git describe --tags --exact-match 2>/dev/null)" = "${tag}" ]; then
+    set_version "${version}"
+    exit 0
+fi
+
+# Current commit is not tagged, append commit(s) sha.
+version="${version}+$(git -C ${SNAPCRAFT_PART_SRC} rev-parse --short=7 HEAD)"
+
+# Main branch will be set as is.
+if [ "${current_branch}" = "main" ]; then
+    set_version "${version}"
+    exit 0
+fi
+
+# Get the short version of last commit merged from the main branch.
+last_commit_on_main=$(git -C "${SNAPCRAFT_PART_SRC}" merge-base main HEAD)
+last_commit_on_main=$(git -C "${SNAPCRAFT_PART_SRC}" rev-parse --short=7 "${last_commit_on_main}")
+version="${version}.${last_commit_on_main}"
+
+set_version "${version}"
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -2,7 +2,7 @@ name: authd-oidc
 summary: OIDC Broker for authd
 description: |
   Broker that enables OIDC authentication for authd.
-version: git
+adopt-info: version
 grade: stable
 base: core24
 confinement: strict
@@ -38,3 +38,10 @@ parts:
     organize:
       "authd.conf": "conf/authd/oidc.conf"
       "broker.conf": "conf/broker.conf.orig"
+  # Build the snap version from the git repository and current tree state.
+  version:
+    source: .
+    plugin: nil
+    build-packages:
+      - git # The script needs Git.
+    override-build: ./snap/get_version


snap/get_version
```diff
@@ -0,0 +1,114 @@
+#!/bin/dash
```
jibel, line 1: OOC why /bin/dash and not /bin/sh

snap/get_version
```diff
@@ -0,0 +1,114 @@
...
+    craftctl set version="${version}"
+}
+
+# annotate_with_dirty may amend the version with a dirty markup if there are local changes.
+# $1: version: the version to annotate.
+annotate_with_dirty() {
+    version="${1}"
+
+    # check if current tree content is dirty.
+    if [ -n "$(git -C ${SNAPCRAFT_PART_SRC} status --porcelain)" ]; then
```
jibel, line 33: This should be done in separate calls. If git fails for some reason, it'll display the error to stderr but the script will happily continue and exit 0

snap/get_version
```diff
@@ -0,0 +1,114 @@
...
+
+# last_tag_on_branch returns the last tag on that branch that is not in another provided branch.
+# $1: origin_branch: branch we start listing from.
+# $2: other_branch: the branch where we don’t want to tags to be reachable from.
+# If we don’t have another branch to compare with, it will just return the most recent tag.
+last_tag_on_branch() {
+    origin_branch="${1}"
+    other_branch="${2}"
+
+    all_tags_on_origin_branch=$(git -C ${SNAPCRAFT_PART_SRC} tag --sort=-committerdate --merged "${origin_branch}")
```
jibel, line 48: should we rely on -e to exit on error or explicit error handling?

snap/get_version
```diff
@@ -0,0 +1,114 @@
...
+    for tag in ${all_tags_on_origin_branch}; do
+        # Just return first commit if there is no branch to compare against the main branch.
+        if [ -z "${other_branch}" ]; then
+            tag=$(strip_branch_tag_prefix "${tag}" "${origin_branch}")
+            echo ${tag}
+            return
+        fi
+
+        # Skip the tags that are reachable from the other branch.
+        if git -C ${SNAPCRAFT_PART_SRC} merge-base --is-ancestor $(git -C ${SNAPCRAFT_PART_SRC} rev-list -n 1 "${tag}") "${other_branch}"; then
```
jibel, line 58: better to split for error handling.

snap/get_version
jibel, line 1: As a general comment, since this script can only run in a snapcraft build environment, it should make sure it is actually running in a snapcraft environment.

jibel (changes requested):

snap/get_version
didrocks: Local testing enforcement, forgot to remove it, nice spot!

didrocks: For debugging or adjusting, I still like to be able to run it. This is under `snap/` directory, and it’s using `SNAPCRAFT_PART_SRC` which needs to be exported with `set -u`. Do you think this is not enough?

snap/get_version
didrocks: agreed, for debugging, especially if we add -x, it will be easier handled in 2 calls.

snap/get_version
didrocks: Good catch!

snap/get_version
didrocks: here, I thought relying on -e for most of git commands and assignements.

If there are bugs we can’t reproduce and this is only CI, `-x` will probably be needed anyway.

didrocks:
Comments should have been addressed in the fixup commit.

snap/get_version
adombeck: This `if` expression doesn't parse

```suggestion
        if git -C ${SNAPCRAFT_PART_SRC} merge-base --is-ancestor $(git -C ${SNAPCRAFT_PART_SRC} rev-list -n 1 "${tag}" "${other_branch}"; then
```

snap/get_version
adombeck: All occurrences of `${SNAPCRAFT_PART_SRC} ` in this file need to be double-quoted to prevent globbing and word splitting

```suggestion
    if [ -n "$(git -C "${SNAPCRAFT_PART_SRC}" status --porcelain)" ]; then
```

snap/get_version
```diff
@@ -0,0 +1,114 @@
...
+last_tag_on_branch() {
+    origin_branch="${1}"
+    other_branch="${2}"
+
+    all_tags_on_origin_branch=$(git -C ${SNAPCRAFT_PART_SRC} tag --sort=-committerdate --merged "${origin_branch}")
+    for tag in ${all_tags_on_origin_branch}; do
+        # Just return first commit if there is no branch to compare against the main branch.
+        if [ -z "${other_branch}" ]; then
+            tag=$(strip_branch_tag_prefix "${tag}" "${origin_branch}")
+            echo ${tag}
```
adombeck, line 54: Double quote to prevent globbing and word splitting
```suggestion
            echo "${tag}"
```

snap/get_version
didrocks: Are you sure about your change? (btw, you are commenting on the older version of the code). However, your suggested change remove the closing parentheses and there is no match for the internal git code, isn’t it?

snap/get_version
```diff
@@ -0,0 +1,114 @@
...
+}
+
+# strip_branch_tag_prefix will remove the branch name prefix from the tag name.
+# $1: tag: the tag name to strip the prefix from.
+# $2: current_branch: the branch name to strip from the tag.
+strip_branch_tag_prefix() {
+    tag="${1}"
+    current_branch="${2}"
+
+    echo ${tag#${current_branch}-}
```
adombeck, line 79: Double quote to prevent globbing and word splitting

```suggestion
    echo "${tag#${current_branch}-}"
```

snap/get_version
didrocks: This closes the older double quote if I am correct and reopen it after ${SNAPCRAFT_PART_SRC}?

snap/get_version
```diff
@@ -0,0 +1,114 @@
...
+
+# Main branch will be set as is.
+if [ "${current_branch}" = "main" ]; then
+    set_version "${version}"
+    exit 0
+fi
+
+# Get the short version of last commit merged from the main branch.
+last_commit_on_main=$(git -C ${SNAPCRAFT_PART_SRC} merge-base main HEAD)
+last_commit_on_main=$(git -C ${SNAPCRAFT_PART_SRC} rev-parse --short ${last_commit_on_main})
```
adombeck, line 113: Double quote to prevent globbing and word splitting

```suggestion
last_commit_on_main=$(git -C ${SNAPCRAFT_PART_SRC} rev-parse --short "${last_commit_on_main}")
```

snap/get_version
didrocks: That was on the outdated code. The new one had the flaw too, I fixed it. Thanks.

snap/get_version
```diff
@@ -0,0 +1,114 @@
...
+# $2: current_branch: the branch name to strip from the tag.
+strip_branch_tag_prefix() {
+    tag="${1}"
+    current_branch="${2}"
+
+    echo ${tag#${current_branch}-}
+}
+
+
+current_branch=$(git -C ${SNAPCRAFT_PART_SRC} branch --show-current)
```
adombeck: We could `cd "${SNAPCRAFT_PART_SRC}"` at the beginning of the script to avoid having to use `-C ${SNAPCRAFT_PART_SRC}` in all git commands.

didrocks:
Thanks for the reviews. Fixup commits pushed. Note that I only didn’t take the suggestion on the invalid parsing which I think was correct at first (and anyway, this was part of the previous revision as based on jibel’s comment I split it in 2)

snap/get_version
didrocks: You shouldn’t do that in general with snapcraft parts, because you are supposed to be in another directory where you can put build artefacts for the finale results. The `"${SNAPCRAFT_PART_SRC}"` is a complete different tree directory that is mapped (in lxd) to the source directory.

My view is that we will produce later some version files to be consumed and compiled inside the finale broker binaries, hence keeping it cleaned and separated into its own build directory for now.

snap/get_version
```diff
@@ -0,0 +1,114 @@
...
+
+# If current commit on that branch is tagged, this is the version we are using.
+current_commit_tag=$(git -C ${SNAPCRAFT_PART_SRC} describe --exact-match --abbrev=0 --tags 2>/dev/null) || true
+if [ -n "${current_commit_tag}" ]; then
```
adombeck: Could be simplified to 
```suggestion
if ! git -C "${SNAPCRAFT_PART_SRC}" describe --exact-match --abbrev=0 --tags 2>/dev/null; then
```

adombeck: Ah nevermind, we're using the `$current_commit_tag` value below

didrocks: Exactly!

snap/get_version
jibel: what about craftctl? Does this command exists outside of snapcraft?

snap/get_version
```diff
@@ -0,0 +1,116 @@
...
+        # Just return first commit if there is no branch to compare against the main branch.
+        if [ -z "${other_branch}" ]; then
+            tag=$(strip_branch_tag_prefix "${tag}" "${origin_branch}")
+            echo "${tag}"
+            return
+        fi
+
+        # Skip the tags that are reachable from the other branch.
+        commit_from_tag=$(git -C "${SNAPCRAFT_PART_SRC}" rev-list -n 1 "${tag}")
+        if git -C "${SNAPCRAFT_PART_SRC}" merge-base --is-ancestor "${commit_from_tag}") "${other_branch}"; then
```
adombeck, line 60: This if expression doesn't parse

```suggestion
        if git -C "${SNAPCRAFT_PART_SRC}" merge-base --is-ancestor "${commit_from_tag}" "${other_branch}"; then
```

snap/get_version
didrocks: As answered on MM, the command will fail (which is generally a no-op). What‘s the issue with this. But again, the script will first fail if you don’t explicit set the SNAPCRAFT env variable. This is not that different from the install hooks/ in `snap/hooks` or in the debian context you have debian/rules that you shouldn’t run by hand. If the script was outside of the `snap/` directory, I would udnderstand, but here, I’m not understanding what‘s the actual concern?

snap/get_version
adombeck: Wrong version of the code indeed, not sure how that happened. This is the line which doesn't parse: https://github.com/ubuntu/authd-oidc-brokers/pull/161#discussion_r1766369012

snap/get_version
didrocks: you are right, this was from the first fixup commit this morning when asked to split up the nested call. Fixed.

snap/get_version
didrocks: makes sense now!

snap/get_version
```diff
@@ -0,0 +1,116 @@
...
+other_branch="main"
+if [ "${current_branch}" = "main" ]; then
+    other_branch=""
+fi
+
+# Get most recent tag on that branch not coming from the other branch.
+version=$(last_tag_on_branch "${current_branch}" "${other_branch}")
+
+# Append current commit sha.
+version="${version}+$(git -C ${SNAPCRAFT_PART_SRC} describe --always)"
```
adombeck, line 103: This is still missing double quotes

```suggestion
version="${version}+$(git -C "${SNAPCRAFT_PART_SRC}" describe --always)"
```

snap/get_version
```diff
@@ -0,0 +1,116 @@
...
+}
+
+# strip_branch_tag_prefix will remove the branch name prefix from the tag name.
+# $1: tag: the tag name to strip the prefix from.
+# $2: current_branch: the branch name to strip from the tag.
+strip_branch_tag_prefix() {
+    tag="${1}"
+    current_branch="${2}"
+
+    echo "${tag#${current_branch}-}"
```
adombeck, line 79: The nested expansion needs to be double quoted as well

```suggestion
    echo "${tag#"${current_branch}-"}"
```

snap/get_version
adombeck: I'm fine with that

snap/get_version
jibel: okay then

snap/get_version
```diff
@@ -0,0 +1,116 @@
...
+
+# last_tag_on_branch returns the last tag on that branch that is not in another provided branch.
+# $1: origin_branch: branch we start listing from.
+# $2: other_branch: the branch where we don’t want to tags to be reachable from.
+# If we don’t have another branch to compare with, it will just return the most recent tag.
+last_tag_on_branch() {
+    origin_branch="${1}"
+    other_branch="${2}"
+
+    all_tags_on_origin_branch=$(git -C "${SNAPCRAFT_PART_SRC}" tag --sort=-committerdate --merged "${origin_branch}")
```
adombeck, line 49: This lists the oldest tag first, so this function returns the oldest commit, not the latest

adombeck, line 49: Could we use the `--no-merged` option of `git tag` to avoid listing tags from $other_branch? Then the `last_tag_on_branch` function would just be

```
git -C "${SNAPCRAFT_PART_SRC}" tag --sort "${SORT}" --merged "${origin_branch}" --no-merged "${other_branch}" | head -n1
```

snap/get_version
```diff
@@ -0,0 +1,116 @@
...
+
+current_branch=$(git -C "${SNAPCRAFT_PART_SRC}" branch --show-current)
+
+# If current commit on that branch is tagged, this is the version we are using.
+current_commit_tag=$(git -C "${SNAPCRAFT_PART_SRC}" describe --exact-match --abbrev=0 --tags 2>/dev/null) || true
+if [ -n "${current_commit_tag}" ]; then
+    current_commit_tag=$(strip_branch_tag_prefix "${current_commit_tag}" "${current_branch}")
+    set_version "${current_commit_tag}"
+    exit 0
+fi
```
adombeck, line 91: Why do we not skip the current commit if it comes from main?

snap/get_version
```diff
@@ -0,0 +1,116 @@
...
+other_branch="main"
+if [ "${current_branch}" = "main" ]; then
+    other_branch=""
+fi
+
+# Get most recent tag on that branch not coming from the other branch.
+version=$(last_tag_on_branch "${current_branch}" "${other_branch}")
+
+# Append current commit sha.
+version="${version}+$(git -C "${SNAPCRAFT_PART_SRC}" describe --always)"
```
adombeck, line 103: This appends the tag if the current commit is tagged. I think we want to append `git rev-parse HEAD` instead.

snap/get_version
```diff
@@ -0,0 +1,116 @@
...
+# Main branch will be set as is.
+if [ "${current_branch}" = "main" ]; then
+    set_version "${version}"
+    exit 0
+fi
+
+# Get the short version of last commit merged from the main branch.
+last_commit_on_main=$(git -C "${SNAPCRAFT_PART_SRC}" merge-base main HEAD)
+last_commit_on_main=$(git -C "${SNAPCRAFT_PART_SRC}" rev-parse --short "${last_commit_on_main}")
+version="${version}.${last_commit_on_main}"
```
adombeck, line 114: The `.` doesn't make it clear which commit is from which branch. How about this instead:

```suggestion
version="${version}+main@${last_commit_on_main}"
```

snap/get_version
didrocks: Are you sure about this? `--sort=committerdate` would sort with the older commits first, but `--sort=-committerdate` defintively lists the more recent ones first. This is also what I can verify with my manual tests.

didrocks: This doesn’t give any outputs here, do you have other results in your end? I replace "${SORT}" with `-committerdate`.

I’m unsure this could actually work as the main tag is technically merged into the other_branch.

adombeck: I also tested it and in my tests it definitely doesn't list the most recent commit first. Sorting by version works though:

```
git tag --sort=-v:refname
```

snap/get_version
didrocks: This is not what the code is doing here. Here, if the current commit is tagged, then, we are using directly that tag name version. See the comment too.

snap/get_version
didrocks: We don’t want to use `git rev-parse` which is giving a long sha and not a short one.

See my previous comment, if the current commit is tagged, we stopped. So the current commit is not tag and will thus append the short sha.

snap/get_version
adombeck: Yes, I understand what the code is doing. If the current commit is tagged, we use that tag. My question is: Why do we not do the same check as below, i.e. that the commit is from main.

snap/get_version
adombeck: Yes, it does work in my tests

adombeck: > I’m unsure this could actually work as the main tag is technically merged into the other_branch.

What do you mean? We don't merge `msentraid-0.1` into the main branch, do we?

snap/get_version
didrocks: So, as told yesterday on MM, `@` is not accepted by snapcraft as a revision. Also I replaced `+` by `-` as a separator first due to the separate between `VERSION+<SHA(s)>+dirtry`.

Finally, as told yesterday too, the limit on version 32 characters. So. if you take into account `xx.yy.zz+<shortsha1>.<shortsha2>+dirty`. this is 361 characters, we can’t add `main` with another character.

I think the version with the 2 shas are for us mainly, and seeing a version like:
`24.09.3+d441319.e6aac67`, we can certainly makes some sense to it with our habit.

snap/get_version
didrocks: I’m still unsure to understand you, this is independent of branch? Meaning that if this is the main branch or any other and the current commit is tagged, we take this. Why would we skip the current commit if it comes from main?

snap/get_version
adombeck: >  if the current commit is tagged, we stopped. So the current commit is not tag

True.

We still don't want to append the output of `git describe` though, because even if the current commit is not a tag, it's not guaranteed to be the short commit SHA. If there is any tag on the branch (in a previous commit), the output is `${TAG}-${SHORT-COMMIT}`.

If we want the short commit SHA, we should use `git rev-parse --short HEAD` instead. I'm not sure we want that though, because the short SHA might not uniquely identify the commit.

snap/get_version
didrocks: The issue with refname is that we shouldn’t really on the version ascii reference.

But I just saw that you resolve it, so this is working now for you as intended?

snap/get_version
didrocks: Unfortuntely, we have to use the short version. I’m happy to switch to this, even if I think this `${TAG}-${SHORT-COMMIT}` is only available if we use the `--all` or `--tags` options, and we don’t use them here.

snap/get_version
didrocks: > What do you mean? We don't merge msentraid-0.1 into the main branch, do we?

Yes, we don’t

> Yes, it does work in my tests

I did retry with more tagging in a forked repo. Indeed, this works. I need then to change the strategy for the main branch, but that’s easy and we can simplify the function, thanks!

snap/get_version
adombeck: I can reproduce getting the `${TAG}-${SHORT-COMMIT}` output from `git describe --always`. So yeah, let's switch to `git rev-parse`. 

The length limit is unfortunate. I think we should specify the length of the SHA we want to include though, because else the length depends on how many commits the git repo has. I think we want the SHA to be 8 characters long, so that the whole version string doesn't exceed the 32 character limit. So we should use `git rev-parse --short=8 HEAD`.

didrocks: > I can reproduce getting the `${TAG}-${SHORT-COMMIT}` output from git describe --always. So yeah, let's switch to git rev-parse.

Waow, out of interest, can you give me reproducible steps for getting it? Learning everyday and we can learn something everyday :)

> The length limit is unfortunate. I think we should specify the length of the SHA we want to include though, because else the length depends on how many commits the git repo has. I think we want the SHA to be 8 characters long, so that the whole version string doesn't exceed the 32 character limit. So we should use git rev-parse --short=8 HEAD.

Right, but 8 characters still don’t cut it (see my last comment, with a "normal versionning" we would be at 31 characters due to dirty.

Or we can maybe change the "dirty" marker is it’s only for us and local builds anyway. Like just appending the `+` in the end (meaning it’s "this plus something"). What do you think?

Even if there is a clash, it means that then we can use our judgement as we have the tag we should looking from and even if possible, the chance of a clash for some informational content in such a short number of elements is negligeable and we could live with it IMHO. (But if we can go for 8 chars, let’s)

snap/get_version
adombeck: No, it doesn't work for me. Not sure why the conversation was resolved, must have been by accident.

adombeck: > The issue with refname is that we shouldn’t really on the version ascii reference.

Why not? We don't plan to ever go down with the version number, right? And `v:refname` sorts versions correctly, it doesn't just do compare the ASCII characters. So for example it correctly sorts `0.10.0` as a higher version than `0.9.0`.

didrocks: Can you share the output of your sort command on the current repo, which has 0.1, msentraid-0.1 (and you can add a newer commit on main and msentraid branch to test).

To me, this is giving:
```
$ git -C . tag --sort=-committerdate --merged msentraid
msentraid-0.2test
v0.2test
msentraid-0.1
0.1
```
Which is the order I expected by recency.

I’m fine going with the refname if you think they will always make logical sense. However, I want to ensure we are not missing anything as you clearly have different results on the sort parameter already.

snap/get_version
adombeck: > Waow, out of interest, can you give me reproducible steps for getting it?

Sure, for example on the main branch of my checkout of the adsys repo I get the following output:

```
$ git describe --always
0.8.4-1825-gda480813
```

> Right, but 8 characters still don’t cut it (see my last comment, with a "normal versionning" we would be at 31 characters due to dirty.

Why can't we go with 31 characters? The character limit for the version is 32, right?

If we need to use even fewer characters, we might want to consider dropping the commit from main. We can get it via `git merge-base main $COMMIT` where `$COMMIT` is the commit from the current branch. So it doesn't really give us any additional information but is "just" for developer convenience (which is a valid reason, I'm all for developer convenience :)).

didrocks: > Sure, for example on the main branch of my checkout of the adsys repo […]

Waow, indeed, there are definitively some black magic (well, it’s just code that we would need to have a look at :p) in term of I add the tag or not as it’s not in authd-oidc-brokers. Anyway, let’s use `rev-parse`.


> Why can't we go with 31 characters? The character limit for the version is 32, right?

You didn’t get me right, I meant, with the current situation in the example I gave with  7 characters in the [corresponding comment](https://github.com/ubuntu/authd-oidc-brokers/pull/161#discussion_r1766540899), this is 31 characters, so we can’t go with 8 characters for both current branch sha and main sha as this would be 33). Hence my proposal to simplify the `+dirty` suffix to just `+`. That way, we can even go up to 10 characters for each SHA.
We would then have: `24.09.3+d44131927.e6aac67a72+` which is 30 characters, marked as dirty and letting us having 2 digits for the patch version if the needs comes (which I hope will never happen). Does that sounds reasonable to you?

adombeck: > You didn’t get me right, I meant, with the current situation in the example I gave with 7 characters in the https://github.com/ubuntu/authd-oidc-brokers/pull/161#discussion_r1766540899, this is 31 characters

Ah ok, got it now.

> Hence my proposal to simplify the +dirty suffix to just +

Just appending `+` feels a bit too subtle to me. I don't want it to go unnoticed when there are changes in a snap version which are untracked. I would prefer going with 7 characters or dropping the commit from main.

didrocks: >> Hence my proposal to simplify the +dirty suffix to just +
> Just appending + feels a bit too subtle to me. I don't want it to go unnoticed when there are changes in a snap version which are untracked. I would prefer going with 7 characters or dropping the commit from main.

Ack, let’s forget about dropping dirty. I don’t really want to drop the commit from main either, if we have a general issue on brokers, this is a good way to ensure we know what level of code we have without digging through git history as we are mostly working from main.

So, if I get you right, you are ok to keep the current implementation with 7 characters for sha? (again, if we have collision, as we are between 2 tags, the chance of it should be really minor and it’s not a crucial info)

snap/get_version
adombeck: We do. As documented in the comment above it, the `last_tag_on_branch` function ignores tags which are reachable from the other branch (main).

If I add a new tag `0.2` on main and merge main into the `msentraid` branch and run the script, I get `0.1+c13d899.e6aac67`. So we do skip the `0.2` tag which comes from main.

snap/get_version
adombeck: > Can you share the output of your sort command on the current repo, which has 0.1, msentraid-0.1 (and you can add a newer commit on main and msentraid branch to test).

Sure, lets look at a snippet of the output in the `git.launchpad.net/ubuntu/+source/apport` repo (on branch `ubuntu/noble`):

```
$ git tag --sort=-committerdate
applied/0.1
applied/0.10
applied/0.100
applied/0.101
<snip>
applied/0.18
applied/0.19
applied/0.2
applied/0.20
applied/0.21
<snip>
```

The `applied/0.1` is not the latest tag, and it was not created after the `applied/0.10` tag.

didrocks: Indeed, I don’t have tangible explanation looking at committer and author dates on those. I don’t know how you ended up trying this on apport git-ubuntu history to find such an example, but that’s ok.

snap/get_version
adombeck: > This is expected, we are considering only the releasing on the current branch not coming from other.
> 
> You rejected the whole idea of "only tagging main" I had at first and not tagging any other branches. This is why then, when you are on msentraid, the tag on main is ignored.

I understand why you did that.

My question remains, why do we don't do the same for the current commit? We could do it like this instead:

```
# Get most recent tag on that branch not coming from the other branch.
tag=$(last_tag_on_branch "${current_branch}" "${other_branch}")

if [ -n "$tag" ] && [ "$tag" = "$(git describe --tags --exact-match)" ]; then
    # The tag is on the current commit, so we don't need to append the commit SHA.
    set_version "${tag}"
    exit 0
fi
```

didrocks: I don’t see the advantage of that approach? If there is a tag, we don’t append the commit SHA anyway and do an early return, as you did here?

I think we really don’t understand each other and I think we should HO before the meeting, now?

snap/get_version
didrocks: This was what I meant. We are in agreement then!

snap/get_version
didrocks: I had to adapt this a little bit as `git describe --tags --exact-match` is failing if there is no tag on it. I think it’s fine just sending stderr to /dev/null and then, we have thus an empty tag which is an acceptable output in that situation.

didrocks:
Ok, the last commit should resolve all opened comments. I did retry it on the current repo, with the various branches and layout and it seems to give the expected outputs.

snap/get_version
adombeck: FTR, I tried it on various git repos, apport was just an example, and it lists the tags in an unexpected order in almost all of them.

snap/get_version
```diff
@@ -0,0 +1,91 @@
...
+version=$(strip_branch_tag_prefix "${version}" "${current_branch}")
+
+# If the most recent tag is on the current commit, taking it as is once transformed as a version.
+if [ -n "${tag}" ] && [ "$(git describe --tags --exact-match 2>/dev/null)" = "${tag}" ]; then
+    set_version "${version}"
+    exit 0
+fi
+
+# Current commit is not tagged, append commit(s) sha.
+version="${version}+$(git -C ${SNAPCRAFT_PART_SRC} rev-parse --short=7 HEAD)"
```
adombeck, line 78:
```suggestion
version="${version}+$(git -C "${SNAPCRAFT_PART_SRC}" rev-parse --short=7 HEAD)"
```

adombeck (approved):

snap/get_version
didrocks: > Yes, that's why I added the [ -n "$tag" ] check before, so that we don't end up setting an empty version if the output of git describe --tags --exact-match is empty.

Right, I was rather explaining on adding `2>/dev/null` to avoid stderr output in the "expected last tag is not versionned" flow.

jibel (approved):
Thanks for doing this! I really like this auto versioning of the snaps.

adombeck:
> Thanks for doing this! I really like this auto versioning of the snaps.

+1

PR #162, opened 2024-09-19
Implement JWT structure and logic for ID token in the provider mock

denisonbarbosa:
This implements the `/key` endpoint for the provider and also reworks the ID token generation in the tests to resemble the "real" ones. By doing this, we are better covered in our tests. More detailed explanations can be seen in the commit messages.

UDENG-3724


internal/testutils/provider.go
```diff
@@ -14,11 +15,23 @@ import (
...
 	"golang.org/x/oauth2"
 )
 
+// MockKey is the RSA key used to sign the JWTs for the mock provider.
+var MockKey *rsa.PrivateKey
+
+func init() {
+	key, err := rsa.GenerateKey(rand.Reader, 2048)
+	if err != nil {
+		panic("Setup: Could not generate RSA key for the Mock")
```
3v1n0, line 36:
```go
panic(fmt.Sprintf("Setup: Could not generate RSA key for the Mock: %v"))
```

So in the unlikely case of failure (I see it happen more in some not-fully configured CI than in real world) we can get it better?

internal/testutils/provider.go
```diff
@@ -24,12 +28,39 @@ import (
...
 	}
 	MockKey = key
+
+	certTemplate := &x509.Certificate{
+		SerialNumber: big.NewInt(2024),
+		Subject: pkix.Name{
+			Organization: []string{"Mocks ltd."},
+		},
+		NotBefore:             time.Now(),
+		NotAfter:              time.Now().AddDate(100, 0, 0),
```
3v1n0, line 46: Hehe, it's gonna last long! Can also just AddDate(0,0,1) no?

3v1n0 (approved):
Nice one, I like when tests are closer to reality!

Just few nits, but all good on my side

adombeck:
> We used to ignore the JWT format in the tests, so we could manually
> write the ID Token. Now that we decided to implement the JWT logic
> also

I'm missing the context here. Why do we implement JWT logic now?

internal/testutils/provider.go
adombeck: +1 for

```suggestion
		panic(fmt.Sprintf("Setup: Could not generate RSA key for the Mock: %v", err))
```

adombeck:
> This is one of the key (pun intended) endpoints of an OIDC provider. It
returns the required keys for decripting some of the payloads. We only
care about the JWK needed to decrypt the tokens, so it only returns that
one.

s/decripting/decrypting

denisonbarbosa:
> > We used to ignore the JWT format in the tests, so we could manually
> > write the ID Token. Now that we decided to implement the JWT logic
> > also
> 
> I'm missing the context here. Why do we implement JWT logic now?

Because this is the format used by OIDC to validate and safeguard the ID tokens. We used to ignore it to make our life easier when creating the mock tokens, but it's better to have our code closer to reality.

PR #163, opened 2024-09-20
Fix snap build by adding git as a build dependency

theovaka:


didrocks (commented):
Hey! Thanks for contributing to authd and its brokers and helping to make it better.

I am unsure about the need for git as build-packages in snapcraft:
- the source is `.`, so snapcraft just copy the current tree (including .git)
- go packages don’t need git by default, dependencies are downloaded through the go proxy itself. The go plugin is using it.

I did a manual build again locally, ensuring that the build environment is cleaned up:
`snapcraft clean`.

And then, just running `snapcraft`, which builds it in a LXD container, I sshed into it, and end up with a `.snap` file and I can confirm that git is not installed and the build is successful.

Is there anything that I am missing?

theovaka:
Hey! I was asked to troubleshoot this by Peter De Sousa because the snap build process was failing. When I tested it, it failed and I saw the following error in the build logs: FileNotFoundError: [Errno 2] No such file or directory: 'git'. Then, I thought that a dependency was missing and included the 'git' declaration in the snapcraft.yaml file. After building again the process finished successfully and the snap was created. I am not sure why git is required but since it fixed the snap build process I proposed this fix.

didrocks:
Thanks!

Even before the recent change to fetch version which depends on git, I tried to rebuild main, just running `snapcraft` and this works here. Theorically, we don’t need git for the go part alone, this is why I think if this is still an issue for you, we should dig deeper.

Can you retry latest main branch? If you still get some issues, it’s worth debugging this together if you have time.

theovaka:
Hey! I have tried latest branch and I can confirm that the build was successful.

didrocks:
+1! Let me close this PR then. Thanks for confirming!

PR #170, opened 2024-10-09
Use paging to get all groups from graph API

adombeck:
By default, Microsoft Graph API only returns 100 objects per call. If
there are more objects, it returns a "@odata.nextLink" property with a
link to the next page of results.

Closes #549

UDENG-4908

### DIFF
--- a/internal/providers/msentraid/msentraid.go
+++ b/internal/providers/msentraid/msentraid.go
@@ -150,40 +150,29 @@ func (p Provider) getGroups(token *oauth2.Token) ([]info.Group, error) {
 
 	var groups []info.Group
 	for _, msGroup := range graphGroups {
-		v, err := msGroup.GetBackingStore().Get("displayName")
-		if err != nil {
-			return nil, fmt.Errorf("failed to get displayName from group object: %v", err)
+		idPtr := msGroup.GetId()
+		if idPtr == nil {
+			slog.Warn(pp.Sprintf("Could not get ID for group: %v", msGroup))
+			return nil, errors.New("could not get group id")
 		}
-		name, ok := v.(*string)
-		if !ok || name == nil {
-			id := msGroup.GetId()
-			if id == nil {
-				unknown := "Unknown"
-				id = &unknown
-			}
-			slog.Warn(pp.Sprintf("Could not get displayName from group object (ID: %s) found: %v", *id, msGroup))
-			return nil, errors.New("could not parse group name")
-		}
-		groupName := strings.ToLower(*name)
+		id := *idPtr
 
-		// Local group
+		groupNamePtr := msGroup.GetDisplayName()
+		if groupNamePtr == nil {
+			slog.Warn(pp.Sprintf("Could not get display name for group object (ID: %s): %v", id, msGroup))
+			return nil, errors.New("could not get group name")
+		}
+		groupName := strings.ToLower(*groupNamePtr)
+
+		// Check if the group is a local group, in which case we don't set the UGID (because that's how the user manager
+		// differentiates between local and remote groups).
 		if strings.HasPrefix(groupName, localGroupPrefix) {
 			groupName = strings.TrimPrefix(groupName, localGroupPrefix)
 			groups = append(groups, info.Group{Name: groupName})
 			continue
 		}
 
-		v, err = msGroup.GetBackingStore().Get("id")
-		if err != nil {
-			return nil, fmt.Errorf("failed to get id from group object: %v", err)
-		}
-		id, ok := v.(*string)
-		if !ok || id == nil {
-			slog.Warn(pp.Sprintf("Could not get ID for group %q: %v", groupName, msGroup))
-			return nil, errors.New("could not parse group id")
-		}
-
-		groups = append(groups, info.Group{Name: groupName, UGID: *id})
+		groups = append(groups, info.Group{Name: groupName, UGID: id})
 	}
 
 	return groups, nil
--- a/internal/providers/msentraid/msentraid.go
+++ b/internal/providers/msentraid/msentraid.go
@@ -143,30 +143,13 @@ func (p Provider) getGroups(token *oauth2.Token) ([]info.Group, error) {
 
 	// Get the groups (only the groups, not directory roles or administrative units, because that would require
 	// additional permissions) which the user is a member of.
-	graphGroups, err := client.Me().TransitiveMemberOf().GraphGroup().Get(context.Background(), nil)
+	graphGroups, err := getAllUserGroups(client)
 	if err != nil {
 		return nil, fmt.Errorf("failed to get user groups: %v", err)
 	}
 
 	var groups []info.Group
-	for _, obj := range graphGroups.GetValue() {
-		unknown := "Unknown"
-		msGroup, ok := obj.(*msgraphmodels.Group)
-		if !ok {
-			id, oType := obj.GetId(), obj.GetOdataType()
-			if id == nil {
-				id = &unknown
-			}
-			if oType == nil {
-				oType = &unknown
-			}
-			slog.Debug(fmt.Sprintf(
-				"Found non-group object with ID: %q of type: %q in graphsdk response. Ignoring it",
-				*id, *oType,
-			))
-			continue
-		}
-
+	for _, msGroup := range graphGroups {
 		v, err := msGroup.GetBackingStore().Get("displayName")
 		if err != nil {
 			return nil, fmt.Errorf("failed to get displayName from group object: %v", err)
@@ -175,9 +158,10 @@ func (p Provider) getGroups(token *oauth2.Token) ([]info.Group, error) {
 		if !ok || name == nil {
 			id := msGroup.GetId()
 			if id == nil {
+				unknown := "Unknown"
 				id = &unknown
 			}
-			slog.Warn(pp.Sprintf("Could not get displayName from group object (ID: %s) found: %v", *id, *msGroup))
+			slog.Warn(pp.Sprintf("Could not get displayName from group object (ID: %s) found: %v", *id, msGroup))
 			return nil, errors.New("could not parse group name")
 		}
 		groupName := strings.ToLower(*name)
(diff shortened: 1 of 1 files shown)


internal/providers/msentraid/msentraid.go
```diff
@@ -143,30 +143,13 @@ func (p Provider) getGroups(token *oauth2.Token) ([]info.Group, error) {
...
-			}
-			if oType == nil {
-				oType = &unknown
-			}
-			slog.Debug(fmt.Sprintf(
-				"Found non-group object with ID: %q of type: %q in graphsdk response. Ignoring it",
-				*id, *oType,
-			))
-			continue
-		}
```
adombeck, line 168: Note that we got rid of this check by using the `msgraphmodels.Groupable` interface (which is the type of the `ItemTransitiveMemberOfGraphGroupRequestBuilder.Get` return value) instead of `*msgraphmodels.Group`.

internal/providers/msentraid/msentraid.go
```diff
@@ -175,9 +158,10 @@ func (p Provider) getGroups(token *oauth2.Token) ([]info.Group, error) {
 		if !ok || name == nil {
 			id := msGroup.GetId()
 			if id == nil {
+				unknown := "Unknown"
 				id = &unknown
 			}
-			slog.Warn(pp.Sprintf("Could not get displayName from group object (ID: %s) found: %v", *id, *msGroup))
+			slog.Warn(pp.Sprintf("Could not get displayName from group object (ID: %s) found: %v", *id, msGroup))
 			return nil, errors.New("could not parse group name")
 		}
```
adombeck, line 166: Could we use `msGroup.GetDisplayName()` here instead of `msGroup.GetBackingStore().Get("displayName")` to get rid of this check as well? Do we not trust the graph API to return the types it's supposed to?

internal/providers/msentraid/msentraid.go
```diff
@@ -175,9 +158,10 @@ func (p Provider) getGroups(token *oauth2.Token) ([]info.Group, error) {
 		if !ok || name == nil {
```
adombeck, line 158: Could we use `msGroup.GetDisplayName()` here instead of `msGroup.GetBackingStore().Get("displayName")` to get rid of this check as well? Do we not trust the graph API to return the types it's supposed to?

didrocks: The issue with that call is that `GetDisplayName()` can panic in two different ways:

First place:
```
    val, err := m.GetBackingStore().Get("displayName")
    if err != nil {
        panic(err)
    }
```

Second place:
```
return val.(*string)
```

Not letting us an idiomatic way to get back the value (or we would wrap the call in its own function to `recover` and turn that into an error), which is kind of equivalent of doing the check here.
If those can `GetBackingStore().Get()` can return an error, it means there are probabilities of it returning this. IIRC, we got some use cases in the past from user report who didn’t grant some permissions/hard some access policy filtering and we received an error here. If we didn’t handle it ourself, the whole broker would have panicked.

didrocks: > To me, the fact that GetDisplayName() panics in error cases indicates that the msgraph-sdk guarantees that it can return the display name and considers it a programming error on their side if one of those two cases you mentioned occurs. I don't think we should check for programming errors in other packages, but trust that package's API (except when we know that the API doesn't do what it's expected to do, in which case we should report the bug and can implement a workaround until it's fixed).

Maybe I was not clear above, but this was not a programming error but a configuration error on the Azure side. We should not panic under those circumstances and in general, for non progamming error, we should always handle the error.

I agree that now, we have a first call protecting against it, but if we are removing it again for reason to pivot to something else, or we have some changes to not filter the group types or ending up with groups we can’t read the display name, this would panic again.

Even on the go stdlib could use panic internally, but it unwinds it and transform it to an error to not crash the application.

The MSGraph Go API is transpiled automatically from their API, my feeling is that they map the exception base system fom C#, java and other to panic for Golang. It’s clear that the library itself is automatically generated and not idiomatic at all. I wouldn’t take the fact that they panics as something that is carefully thought on their side.

> That would be a good reason, and we document it in a comment if that happened (and report it). I wonder how that could have happened though, because AFAICT, we have always used GetBackingStore().Get() since GetGroups was first implemented in https://github.com/ubuntu/authd-oidc-brokers/commit/cad35ac813857f8723a2414f8c2100384b3d92ed.

It clearly did happen. I might be wrong but we discovered it when we adding more logs to it in July: https://github.com/ubuntu/authd-oidc-brokers/pull/67/files for instance. That was when we didn’t filter from the parameters the directory objects and when the user restricted some access on some groups we couldn’t read through the MS Entra ID portal.

adombeck: > this was not a programming error but a configuration error on the Azure side.

If the API server actually sends incorrect data types, that's still a bug on Microsoft's side and IMO we should handle it the same way (don't assume that it can happen until it does, if so work around it and report it).

> I agree that now, we have a first call protecting against it

Where do we have a first call protecting against it?

> It’s clear that the library itself is automatically generated and not idiomatic at all.

Agreed

> I wouldn’t take the fact that they panics as something that is carefully thought on their side.

IMO it's acceptable to use panic if your code guarantees that the panic is never actually called, i.e. it's a programming error when the panic is called.

If we still decide that we don't trust the msgraph SDK, I think we should make that clear in a comment and ensure that we do the same checks whenever we call a similar function of the SDK.

> It clearly did happen. I might be wrong but we discovered it when we adding more logs to it in July: https://github.com/ubuntu/authd-oidc-brokers/pull/67/files for instance. That was when we didn’t filter from the parameters the directory objects and when the user restricted some access on some groups we couldn’t read through the MS Entra ID portal.

I expect that ` msGroup.GetBackingStore().Get("displayName")` returned `nil` in that case (and not a value of an incorrect type), which `msGroup.GetDisplayName()` would also do.

didrocks: > Where do we have a first call protecting against it?

The fact that we filter the group not being of the right time with the filtering we discussed the other day.

> IMO it's acceptable to use panic if your code guarantees that the panic is never actually called, i.e. it's a programming error when the panic is called.

I’m unsure if I’m not clear at all on that but I will let @denisonbarbosa expanding, but as said, we had this particular use case with an incorrect configuration on MS Entra ID and not a programming error itself. So, we are in agreement that we should not panic in that situation. From what I know, every situation were we know we can get a panic for an invalid MS Entra ID configuration has been handled similarly. If you want to add a comment explaning why this check is done, please go for it.

denisonbarbosa: > IMO it's acceptable to use panic if your code guarantees that the panic is never actually called, i.e. it's a programming error when the panic is called.

We can not assert that a third-party dependency will never panic. Especially, as Didier mentioned, since we are dealing with generated bindings rather than a Go implementation.

Also, we are talking about libraries and services here. If the PAM module panics, you have a big problem, and the user might be locked out completely. If the broker daemon panics, the user will be locked out until the machine is restarted or a local admin restarts the broker service. So no, I will never trust panics from dependencies (and we did have a lot of issues with Microsoft dependencies bumps in the past during the development of  `aad-auth`).

internal/providers/msentraid/msentraid.go
```diff
@@ -175,9 +158,10 @@ func (p Provider) getGroups(token *oauth2.Token) ([]info.Group, error) {
...
 			id := msGroup.GetId()
 			if id == nil {
```
denisonbarbosa, line 160: Btw, this isolated call for `GetId` is something that we did in a hurry to help a user debug some issues and we forgot to change it to the same format that we use in lines 176-180.

denisonbarbosa (changes requested):
Changes look fine to me, nice work on that! However, we need better descriptions for the PR description and the commit message here... "Closes #{issue}" does not say much.

internal/providers/msentraid/msentraid.go
didrocks: To summarize our discussion online:

The `Get<Foo>` calls can only panic in 2 cases:
* if the backend memory returns an error. We only found the `InMemory` backend. It returns an error if the key is empty (which is a programming error)
* the other case is due to the returned pointer to string being nil, probably not on the casting itself, so we can drop the `!ok` but we keep the rest of the check (as `Get<Foo>` will return nil for nil object)

The initial perception was that the InMemory backend store was only used for testing, but it seems it’s not. Consequently, the agreed path is to test that we can assert we are using the InMemory backend (just a one time manual test), and if so, we can use the `Get<Foo>` calls, which will allow us to remove the first `if err != nil` check and `!ok` in the `||` line.

adombeck:
> Changes look fine to me, nice work on that! 

Thanks!

> However, we need better descriptions for the PR description and the commit message here... "Closes #{issue}" does not say much.

I thought the first line of the commit message ("Use paging to get all groups from graph API") was explanation enough, but I now added some more explanation in another paragraph.

denisonbarbosa:
> I thought the first line of the commit message ("Use paging to get all groups from graph API") was explanation enough, but I now added some more explanation in another paragraph.

It was explanation enough for the "what/how" questions but was missing the answer to the "why" question (it's fine to answer it only in the PR description). We should always try to answer that in the pull requests so it takes less time for our future selves to remember why this was done in the first place.

adombeck:
> It was explanation enough for the "what/how" questions but was missing the answer to the "why" question (it's fine to answer it only in the PR description). We should always try to answer that in the pull requests so it takes less time for our future selves to remember why this was done in the first place.

Fully agreed, IMO the *why* part is even more important than the *what* part, because the latter can always be understood from the changes themself, but the former cannot.

internal/providers/msentraid/msentraid.go
```diff
@@ -143,63 +143,64 @@ func (p Provider) getGroups(token *oauth2.Token) ([]info.Group, error) {
...
 
-		v, err = msGroup.GetBackingStore().Get("id")
+		groups = append(groups, info.Group{Name: groupName, UGID: id})
+	}
+
+	return groups, nil
+}
+
+func getAllUserGroups(client *msgraphsdk.GraphServiceClient) ([]msgraphmodels.Groupable, error) {
+	var groups []msgraphmodels.Groupable
```
denisonbarbosa, line 182: We should move this declaration closer to its use (i.e. line 192). Can we also avoid the `var` directive and go for `:=` instead?

denisonbarbosa (changes requested):
One last nitpick and then we're good :upside_down_face:

denisonbarbosa (approved):
Thanks for the changes. Nice work!

PS.: Don't forget to update the PR description

PR #178, opened 2024-10-14
Optional client secret for some providers

didrocks:
Some providers, like google, request a client secret for the device authentication. Make it optional, so that the admin can set it if required in the generic broker.

There is no integration tests for now that could make use of it, but all existing tests are passing. I have also tested manually with google and msentraid.

UDENG-4954

### DIFF
--- a/conf/broker.conf
+++ b/conf/broker.conf
@@ -1,6 +1,9 @@
 [oidc]
 issuer = https://<ISSUER_URL>
 client_id = <CLIENT_ID>
+# Client secret is needed for some specific auth flows depending on the provider.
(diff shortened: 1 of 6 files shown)


didrocks:
Note: I’m keeping this one as draft until PR #134 is merged as it’s changing some configuration handling.

didrocks:
> Should https://github.com/ubuntu/authd-oidc-brokers/pull/204 be merged into optional_client_secret before we merge this?

No, this PR works by itself for Google personal versions for instance, so I prefer smaller PRs and than getting #204 separated and rebased on main.

adombeck (approved):

PR #179, opened 2024-10-15
Use email for generic provider as the identifier

didrocks:
Most web login platform are using the user email address as an identifier. Let’s align the generic "no provider" on this, which can then be overriden by specific provider code.
Align tests on using emails as identifier for generic provider.

UDENG-4950


3v1n0:
Nothing to say, all good on my side!

3v1n0 (approved):

PR #180, opened 2024-10-15
Google provider

didrocks:
We need to override the generic provider to remove the offline_access scope. For TV and limited input devices apps, Google APIs will reject any request with this scope, while still allowing non interactive user access token refresh.
All the rest of the functionalities is aligned with the NoProvider one.
Added tests and "withgoogle" build tag for it.

UDENG-4953


3v1n0 (approved):
Nice one, cool that can start having this with minor effort.

Looking forward for the base renaming/factorizing branch!

PR #183, opened 2024-10-16
Add instructions on how to report security issues

seb128:
Quoting SEC0026 - SSDLC

> SECURITY.md (V1.1)
> Any public repository must also include a SECURITY.md file in the root directory, which educates users/contributors on
> how to report a security concern.

The content of the file is based on the template referenced in the spec and what other Canonical repository are already using

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,8 @@
+# How to report a security issue with authd oidc-based brokers
+
+The easiest way to report a security issue is through [GitHub](https://github.com/ubuntu/authd-oidc-brokers/security/advisories/new).
+See [Privately reporting a security vulnerability](https://docs.github.com/en/code-security/security-advisories/guidance-on-reporting-and-writing/privately-reporting-a-security-vulnerability)
+for instructions.
+
+The [Ubuntu Security disclosure and embargo policy](https://ubuntu.com/security/disclosure-policy)
+contains more information about what you can expect when you contact us and what we expect from you.


SECURITY.md
```diff
@@ -0,0 +1,8 @@
+# How to report a security issue with authd-msentraid
```
didrocks, line 1: This is generic and not only the msentra id brokers. Do you mind rephrasing that this is for any brokers here?

Or this line could be with "authd oidc-based broker" and then, we can have one line listing them (authd-msentraid for now).

didrocks (changes requested):
Once small comments/adjustement that I think is necessary for this monorepo with different project targets before we merge it. Tell me what you think?

SECURITY.md
seb128: It is targetting the [msentraid](https://github.com/ubuntu/authd-oidc-brokers/tree/msentraid) branch and matching the name set in the snapcraft.yaml but I'm fine changing it to be generic if you prefer

didrocks: Right, I don’t think people will really go to each branch to read about security.md for that specific branch, also GitHub integrates when you open an issue a read of `security.md` which seems to be following the content in HEAD (like it does for the bug template). So I would:
1. Target only the main branch (automerge will copy them in dedicated branches anyway)
2. Have generic content there, that will cover all brokers

didrocks (approved):
Looks good now! Thanks!

PR #186, opened 2024-10-24
broker: Support qrcode rendering field and adapt device authentication label when not supported

3v1n0:
This is the broker side of https://github.com/ubuntu/authd/pull/586

However, I've used a conservative approach here too, so that version differences won't create issues, by using the old behavior in case authd is not new enough to expose the broker whether it supports rendering.

Otherwise, if the rendering is explicitly not supported, we change the layout text.

Comes with some tests and some cleanups, see commits for details.

UDENG-4814


internal/broker/authmodes/consts.go
```diff
@@ -0,0 +1,16 @@
+// Package authmodes lists the authentication modes that providers can support.
+package authmodes
+
+const (
+	// PasswordID is the ID of the password authentication method.
+	PasswordID = "password"
```
denisonbarbosa, line 6: I think we can drop the ID suffix altogether, wdyt? It would become `authmodes.Password`, which feels quite good IMO.

internal/broker/broker.go
```diff
@@ -311,35 +326,45 @@ func (b *Broker) generateUILayout(session *sessionInfo, authModeID string) (map[
...
 				"Scan the QR code or access %q and use the provided login code",
 				response.VerificationURI,
-			),
+			)
+		} else {
+			label = fmt.Sprintf(
+				"Access %q and use the provided login code",
+				response.VerificationURI,
+			)
+		}
```
denisonbarbosa, line 349: I think it's best to assume the "default" label as being without the code here and then assign the other one in case QRCode is supported. This way we can drop the else 🙃

internal/broker/broker_test.go
```diff
@@ -139,6 +140,15 @@ var supportedUILayouts = map[string]map[string]string{
 	"qrcode-without-wait": {
 		"type": "qrcode",
 	},
+	"qrcode-without-qrcode": {
```
denisonbarbosa, line 143: 🤔

denisonbarbosa (changes requested):
I have some naming suggestions, but everything else looks great!

internal/broker/authmodes/consts.go
3v1n0: Oh, sure... I kept `ID` since initially I was putting them as part of the `brokers` namespace, but since that didn't work (as per circular references), I've created a new namespace... We can even call it `authmodesids` but what we have should be good enough.

internal/broker/broker.go
3v1n0: Hehe, yeah, I was initially doing this, but then I decided not to follow the happy-path as there was really no point to allocate things if not needed...

I can move this to a function though :)

internal/broker/broker_test.go
3v1n0: Yep, that was on purpose :joy:

3v1n0:
@denisonbarbosa, @adombeck I've no permissions to merge here, so maybe can you do it for me?

It's rebased already...

adombeck:
>  broker: Expose supported authentication modes in const variables 

Thank you for this! The use of literal strings instead of string constants makes the code hard to navigate and maintain. I planned to extract some string constants, and this change was definitely one of those.

3v1n0:
> >  broker: Expose supported authentication modes in const variables 
> 
> Thank you for this! The use of literal strings instead of string constants makes the code hard to navigate and maintain. I planned to extract some string constants, and this change was definitely one of those.

Yeah, I agree... I've some similar changes for authd in my stash too. Will reiterate.

adombeck:
There are a lot of merge conflicts with https://github.com/ubuntu/authd-oidc-brokers/pull/134. Could we maybe merge this after that and resolve the conflicts here?

3v1n0:
Fair enough...

adombeck:
So since @denisonbarbosa already merged this, are you fine with me reverting the merge?

3v1n0:
> So since @denisonbarbosa already merged this, are you fine with me reverting the merge?

Oh... If you want I can help rebasing at this point...

adombeck:
> Oh... If you want I can help rebasing at this point...

I really think rebasing https://github.com/ubuntu/authd-oidc-brokers/pull/134 will be a lot more effort than the other way around, because it has a lot more commits with a lot more changes.

PR #187, opened 2024-11-05
config option to refer to the oidc configuration

didrocks:
We unfortunately have 2 configurations as I just noticed:
- one for the system paths + verbosity (the viper/daemon configuration)
- one for the oidc part.

--config was referring to the first one. This PR introduce --paths-config for the first one and --config is now targetting the second one, overriding this specific system path configuration for broker config.

Those 2 configurations should really be merged in a single file and let viper handling this. This would have the net benefit of supporting multiple configuration format too and remove a bunch of manual parsing. However, PR #134 has quite some configuration related changes already and that would conflict with it.

So, here, this change is to minimize the conflict with it, while still allowing being merged beforehand to unblock other teams.

A followup action (once the config change PR will be merged) will be on merging both and handling the parsing by viper, which should greatly simplify the overall structure.

UDENG-5083

### DIFF
--- a/cmd/authd-oidc/daemon/config.go
+++ b/cmd/authd-oidc/daemon/config.go
@@ -30,7 +30,7 @@ func initViperConfig(name string, cmd *cobra.Command, vip *viper.Viper) (err err
 	setVerboseMode(v)
 
 	// Handle configuration.
-	if v, err := cmd.Flags().GetString("config"); err == nil && v != "" {
+	if v, err := cmd.Flags().GetString("paths-config"); err == nil && v != "" {
 		vip.SetConfigFile(v)
 	} else {
 		vip.SetConfigName(name)
--- a/cmd/authd-oidc/daemon/daemon.go
+++ b/cmd/authd-oidc/daemon/daemon.go
@@ -90,6 +90,8 @@ func New(name string) *App {
 
 	installVerbosityFlag(&a.rootCmd, a.viper)
 	installConfigFlag(&a.rootCmd)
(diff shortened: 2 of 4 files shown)


denisonbarbosa (approved):
Looks great and needed. Well done!

adombeck:
> So, here, this change is to minimize the conflict with it, while still allowing being merged beforehand to unblock other teams.

Which teams are blocked by this?

adombeck:
> --config was referring to the first one. This PR introduce --paths-config for the first one and --config is now targetting the second one, overriding this specific system path configuration for broker config.

Changing the behavior of command-line options is technically a breaking change, so we should consider the impact of that. Since the broker runs in a snap, I wonder how the user can even use those command-line options.

adombeck:
> Which teams are blocked by this?

FTR, this was explained out-of-band

didrocks:
> Changing the behavior of command-line options is technically a breaking change, so we should consider the impact of that. Since the broker runs in a snap, I wonder how the user can even use those command-line options.

It indeed is, but I want to point 2 things:
-  we are pre 1.0, so I don’t think we commit to anything thus far
- as you told, right now, in the snap, apart if you do some systemd unit override on the autogenerated one, this is not accessible.

Given the current state of the projects, number of users and usability/reach  of this command line option, I think this is acceptable.

adombeck:
> Given the current state of the projects, number of users and usability/reach of this command line option, I think this is acceptable.

I agree. Just wanted to make sure we're making a conscious decision, and are mindful of introducing breaking changes :)

adombeck (approved):

PR #190, opened 2024-11-05
Handle file changes done by `snapcraft --destructive-mode`

adombeck:
* Ignore directories created by `snapcraft --destructive-mode`
* Make `snap/hooks/install` executable (that's done by `snapcraft --destructive-mode`)
    
Rationale:  `snapcraft --use-lxd` currently fails on my system with some hard-to-debug error, so I'm using `snapcraft --destructive-mode` for development builds of the authd-msentraid snap for now.


didrocks (approved):
Sounds good, I don’t see any rationale to not ignore them or our willingness to one day depends on them, so let’s go with it :)

PR #193, opened 2024-11-06
Fix D-Bus XML

adombeck:
"a{s}s" is not a valid GVariant type, it should be "a{ss}" instead.


adombeck:
Thanks to @michaelwildvarian for spotting this!

3v1n0:
+1.

From my tests for some reason, go.dbus doesn't consider this an issue sadly... So it would be nice to try to understand the root issue and maybe try to fix it upstream. But meanwhile this is the right change for us.

denisonbarbosa (approved):

PR #197, opened 2024-11-07
Fix broker tests

adombeck:
* Remove the key "token" in `firstAuthInfo` map, because it's not used anywhere and as far as I can tell it never was.
* Since the "token" key is not used, there are three test cases with exactly the same preconditions.  Change the preconditions so that the test cases now actually test what they are supposed to test.


adombeck:
Another thing I find odd:

These tests: 

https://github.com/ubuntu/authd-oidc-brokers/blob/41661e6bbf9cc6af5fa5e55dd3d38aa8ff0af927/internal/broker/broker_test.go#L468-L483

Seem to be exactly the same as these tests:

https://github.com/ubuntu/authd-oidc-brokers/blob/41661e6bbf9cc6af5fa5e55dd3d38aa8ff0af927/internal/broker/broker_test.go#L484-L499

@3v1n0 it seems like you introduced those recently in 5a59439f6a95a3c1601a052474c85da51c08b70d.

3v1n0:
Ouch, yes I think I forgot to set `firstMode: authmodes.Device` (instead of `authmodes.DeviceQr`)? :thinking:

internal/broker/testdata/TestIsAuthenticated/golden/error_when_mode_is_newpassword_and_id_token_is_not_set/data/provider_url/test-user@email.com/token.json
```diff
@@ -0,0 +1 @@
+Definitely an encrypted token
```
denisonbarbosa, line 1: Is this highlighted red only on my screen?

internal/broker/testdata/TestIsAuthenticated/golden/error_when_mode_is_newpassword_and_fetching_user_info_fails/first_call
```diff
@@ -0,0 +1,3 @@
+access: denied
+data: '{"message":"authentication failure: could not get required information"}'
+err: <nil>
```
denisonbarbosa, line 3: As far as I'm aware, `newpassword` does not care about the userInfo, so this test is a bit confusing for me. Maybe I understood it wrong?

internal/broker/broker_test.go
```diff
@@ -500,18 +499,15 @@ func TestIsAuthenticated(t *testing.T) {
 		},
 		"Error when empty challenge is provided for local password": {firstChallenge: "-", wantSecondCall: true, secondChallenge: "-"},
 		"Error when mode is newpassword and token is not set": {
-			firstMode:     authmodes.NewPassword,
-			firstAuthInfo: map[string]any{"token": nil},
+			firstMode: authmodes.NewPassword,
```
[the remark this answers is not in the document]
adombeck: I still don't understand. Where are we asserting on the presence of `session.authInfo`?

internal/broker/testdata/TestIsAuthenticated/golden/error_when_mode_is_newpassword_and_id_token_is_not_set/data/provider_url/test-user@email.com/token.json
adombeck: In GitHub I also see it highlighted in red. Maybe because it's not valid JSON?

denisonbarbosa: Could be... But it should be happening to other files as well then, no?

adombeck: Looking at the git history, those files were created without the `.json` suffix in a69d5884391b867edcfdb6e2760adfe5371808ef. They were renamed in 524fb8216f6aef0781d3b0efea60f70cafa98d7c, but because they were only renamed, GitHub didn't show the content in the diff. So this might actually be the first PR which creates a new `token.json` file with non-JSON content.

denisonbarbosa (approved):
Looks good to me. Thanks for the fine-tuning!

adombeck:
> Ouch, yes I think I forgot to set firstMode: authmodes.Device (instead of authmodes.DeviceQr)? 🤔

Tracked in https://github.com/ubuntu/authd/issues/629

PR #198, opened 2024-11-07
Rename tests

adombeck:
~~**This is based on https://github.com/ubuntu/authd-oidc-brokers/pull/197, please review and merge that first.**~~

FTR, this is the regex which I used: `s/("[A-Z][^"]*[^"\s])\s+([^"]*[^"\s]":\s+\{)/$1_$2/g`

UDENG-5165


3v1n0:
I see the rationale here and I agree that it's annoying to search for a test result, my problem here is how we preserve this policy without having a linter preventing adding tests that don't follow this pattern.

denisonbarbosa (changes requested):
Ok... I can understand the appeal of the change. However, this brings up an interesting discussion that we should have with the whole squad to get to a common ground. This makes debugging tests easier, for sure, but it affects the readability of the test cases and it's something that we focused "a lot of energy" to maintain until this point (that's the reason I pushed back on the whole "setup functions inside test cases" discussion that we had). I think this is a topic to raise in one of our meetings, with the whole team together, and then, if we decide to go for this, we adjust the naming on our projects.

adombeck:
> I see the rationale here and I agree that it's annoying to search for a test result

It's not even searching for the test result which annoys me and motivated me to make this change, but that when I want to run a specific test case with `go test -run $TESTNAME`, I can't copy-paste the test name but need to manually type it (or manually replace the spaces with underscores).

> This makes debugging tests easier, for sure, but it affects the readability of the test cases

To me, replacing the spaces with underscores don't affect readability negatively (or only very little).

adombeck:
> my problem here is how we preserve this policy without having a linter preventing adding tests that don't follow this pattern.

I think if we agree on this pattern, we will quickly get used to using it when adding new tests or reviewing changes to the test names. And it's not a big problem if a change slips through, we can always fix it later when we notice it.

cmd/authd-oidc/daemon/daemon_test.go
```diff
@@ -116,9 +116,9 @@ func TestAppRunFailsOnComponentsCreationAndQuit(t *testing.T) {
 		dataDirBehavior int
 		configBehavior  int
 	}{
-		"Error on existing data dir being a file":    {dataDirBehavior: dirIsFile},
-		"Error on data dir missing parent directory": {dataDirBehavior: noParentDir},
-		"Error on wrong permission on data dir":      {dataDirBehavior: wrongPermission},
+		"Error_on_existing_data_dir_being_a_file":    {dataDirBehavior: dirIsFile},
```
denisonbarbosa, line 119: The whole purpose of using capitalized letters for the test cases was because they were supposed to be "English sentences". Now that we are using a more "computer-like" format, maybe we can also drop the uppercased first letter rule?

denisonbarbosa (approved):
LGTM! Feel free to merge once the conversation on Jira is settled also.

PR #199, opened 2024-11-08
Update golden files env var

adombeck:
Update golden files via the TESTS_UPDATE_GOLDEN env var
    
This allows us to update all golden files in one go, without having to
install the update flag per test package and deal with errors that the
flag is not supported in some packages.

Analogous to https://github.com/ubuntu/authd/pull/252.

UDENG-5173


denisonbarbosa (approved):
Thanks for addressing this!

PR #207, opened 2024-11-12
Improve golden file error messages

adombeck:
UDENG-5195

This changes the error message from:

```
Error:      	Not equal: 
expected: map[string]string{"button":"Request new login code", "code":"user_code", "content":"https://verification_uri.com", "foo":"bar", "label":"Scan the QR code or access \"https://verification_uri.com\" and use the provided login code", "type":"qrcode", "wait":"true"}
actual  : map[string]string{"button":"Request new login code", "code":"user_code", "content":"https://verification_uri.com", "label":"Scan the QR code or access \"https://verification_uri.com\" and use the provided login code", "type":"qrcode", "wait":"true"}

Diff:
--- Expected
+++ Actual
@@ -1,2 +1,2 @@
-(map[string]string) (len=7) {
+(map[string]string) (len=6) {
  (string) (len=6) "button": (string) (len=22) "Request new login code",
@@ -4,3 +4,2 @@
  (string) (len=7) "content": (string) (len=28) "https://verification_uri.com",
- (string) (len=3) "foo": (string) (len=3) "bar",
  (string) (len=5) "label": (string) (len=89) "Scan the QR code or access \"https://verification_uri.com\" and use the provided login code",
Test:       	TestSelectAuthenticationMode/Successfully_select_device_auth_qr
Messages:   	SelectAuthenticationMode should have returned the expected layout
```

to 

```
Error:      	Golden file content mismatch

Expected (golden):
--------------------------------------------------
button: Request new login code
code: user_code
content: https://verification_uri.com
label: Scan the QR code or access "https://verification_uri.com" and use the provided login code
type: qrcode
wait: "true"
foo: bar
--------------------------------------------------

Actual: 
--------------------------------------------------
button: Request new login code
code: user_code
content: https://verification_uri.com
label: Scan the QR code or access "https://verification_uri.com" and use the provided login code
type: qrcode
wait: "true"
--------------------------------------------------

Diff:
--- Expected (golden)
+++ Actual
@@ -4,5 +4,4 @@
 label: Scan the QR code or access "https://verification_uri.com" and use the provided login code
 type: qrcode
 wait: "true"
-foo: bar
 
Test:       	TestSelectAuthenticationMode/Successfully_select_device_auth_qr
Messages:   	Golden file: testdata/TestSelectAuthenticationMode/golden/Successfully_select_device_auth_qr
```

If [`delta`](https://github.com/dandavison/delta) is installed on the system, the diff is also colorized:

image

### DIFF
--- a/internal/testutils/golden.go
+++ b/internal/testutils/golden.go
@@ -32,9 +32,20 @@ func init() {
 	}
 }
 
-// GoldenOptions are options for functions that work with golden files.
-type GoldenOptions struct {
-	Path string
+type goldenOptions struct {
+	path string
+}
+
+// GoldenOption is a supported option reference to change the golden files comparison.
+type GoldenOption func(*goldenOptions)
+
+// WithGoldenPath overrides the default path for golden files used.
+func WithGoldenPath(path string) GoldenOption {
+	return func(o *goldenOptions) {
+		if path != "" {
+			o.path = path
+		}
+	}
 }
 
 func updateGoldenFile(t *testing.T, path string, data []byte) {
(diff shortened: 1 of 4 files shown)


internal/testutils/golden.go
3v1n0, line 1: In a similar way, can't we also get this when using pure `diff --color=always` (or auto)?

In general I'd avoid even checking for such tools when we're not in a tty though.

adombeck: > In general I'd avoid even checking for such tools when we're not in a tty though.

Why? I didn't test it yet (I should probably do that), but I assume that the colored output will also be shown correctly in the GitHub CI, where we're not in a tty.

adombeck: > I didn't test it yet (I should probably do that)

Looks good: https://github.com/ubuntu/authd-oidc-brokers/actions/runs/11806570571/job/32891480199?pr=207#step:5:153

internal/testutils/golden.go
```diff
@@ -28,58 +28,74 @@ func init() {
...
 
-// WithGoldenPath overrides the default path for golden files used.
-func WithGoldenPath(path string) GoldenOption {
-	return func(o *goldenOptions) {
-		if path != "" {
-			o.goldenPath = path
-		}
+// CheckOrUpdateGolden compares the provided string with the content of the golden file. If the update environment
+// variable is set, the golden file is updated with the provided string.
+func CheckOrUpdateGolden(t *testing.T, got string, opts *GoldenOptions) {
```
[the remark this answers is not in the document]
adombeck: I did that in a previous iteration where `CheckOrUpdateGolden` called `LoadWithUpdateFromGolden`, which I found to not work well with variadic options. However, `CheckOrUpdateGolden` doesn't call `LoadWithUpdateFromGolden` anymore, so we can go back to the variadic approach.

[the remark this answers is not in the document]
adombeck: done

adombeck:
According to https://github.com/ubuntu/authd/pull/583#discussion_r1840204513, there was a discussion before about changing the Ubuntu version used in our GitHub workflows from jammy to noble, and it was decided that we want to stay on jammy for some reason. If that is correct, I'll have to revert 79afbac2ae5d7c64bf53178b4e6cee139526a1c2, and then we won't have the nice output from delta in our test output. I could live with that, but I wonder why we want to stay on jammy.

denisonbarbosa (approved):

PR #204, opened 2024-11-11
fix: set AuthURLParam to set client id and secret

shipperizer:
workaround to deal with #golang/oauth2/320

PR #216, opened 2024-11-13
fix: use AuthURLParam to set client id and secret

shipperizer:
workaround to deal with golang/oauth2#320

tldr is that IDP servers tend to not be fully compliant with how client
credentials are passed and have bespoke arrangements so anything goes
this enforces the standard implementation from the RFC and has it working
for any RFC compliant OIDC server

full info here https://github.com/golang/oauth2/issues/320

supersedes #204

### DIFF
--- a/internal/broker/broker.go
+++ b/internal/broker/broker.go
@@ -357,7 +357,20 @@ func (b *Broker) generateUILayout(session *sessionInfo, authModeID string) (map[
 	case authmodes.Device, authmodes.DeviceQr:
 		ctx, cancel := context.WithTimeout(context.Background(), maxRequestDuration)
 		defer cancel()
-		response, err := session.authCfg.oauth.DeviceAuth(ctx)
+
+		var authOpts []oauth2.AuthCodeOption
+
+		// workaround to cater for fully RFC compliant oauth2 server which require this
+		// extra option, public providers tend to have bespoke implementation for passing client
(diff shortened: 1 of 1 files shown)


didrocks:
Thanks for adding the comments + fix the slice. I will give it a try on msentraid and enabled the tests in CI.

internal/broker/broker.go
```diff
@@ -357,7 +357,21 @@ func (b *Broker) generateUILayout(session *sessionInfo, authModeID string) (map[
...
+		// credentials that completely bypass this
+		// full explanation in https://github.com/golang/oauth2/issues/320
+		if secret := session.authCfg.oauth.ClientSecret; secret != "" {
+			// TODO @shipperizer verificationMethod should be a configurable value
+			verificationMethod := "client_post"
+			authOpts = append(authOpts, oauth2.SetAuthURLParam(verificationMethod, secret))
+		}
+
+		response, err := session.authCfg.oauth.DeviceAuth(ctx, authOpts...)
+
```
didrocks, line 374: nitpick: still an extra blank line here.

didrocks (changes requested):

didrocks:
Tests are passing on MSEntraID.

didrocks (approved):
Excellent! Everything’s passing now. Let’s merge. Thanks a lot!

PR #220, opened 2024-11-13
ci: Use gotestfmt to format go test output

adombeck:
### TODO
* [x] Change the branch of the GitHub Action to `main` once https://github.com/canonical/desktop-engineering/pull/42 has been merged

UDENG-5208

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,6 +28,16 @@ jobs:
         with:
           go-version-file: go.mod
 
+      - name: Prepare tests artifacts path
+        run: |
+          set -eu
+
+          artifacts_dir=$(mktemp -d --tmpdir authd-test-artifacts-XXXXXX)
+          echo AUTHD_TEST_ARTIFACTS_DIR="${artifacts_dir}" >> $GITHUB_ENV
+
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -37,22 +47,24 @@ jobs:
           rm -fr "${raw_cov_dir}"
           mkdir -p "${raw_cov_dir}"
 
+          # Print executed commands to ease debugging
+          set -x
```
denisonbarbosa, line 51: I'm still against keeping `set -x` by default. It's something that we need to debug the action, sure, but it should be enabled only when debugging is needed.

denisonbarbosa (changes requested):

.github/workflows/ci.yml
adombeck: Some failures are hard to reproduce, so it's not as easy as adding `set -x` and re-running the job again.  The default output of the CI job should be at a balance between too much output, which makes it hard to find the relevant output, and enough output to allow us debugging failures. I added the `set -x` just before the commands for which it might not be clear where the output is coming from, i.e. the `go test` and `go tool` commands.

denisonbarbosa (approved):

PR #224, opened 2024-11-14
Minor broker refactorings

adombeck:
See commit messages for descriptions of the various changes.

UDENG-5453

### DIFF
--- a/internal/testutils/provider.go
+++ b/internal/testutils/provider.go
@@ -74,14 +74,14 @@ type providerServerOption struct {
 // ProviderServerOption is a function that allows to override default options of the mock provider.
 type ProviderServerOption func(*providerServerOption)
 
-// WithHandler specifies a handler to the requested path in the mock provider.
+// WithHandler returns a ProviderServerOption that adds a handler for a provider endpoint specified by path.
 func WithHandler(path string, handler func(http.ResponseWriter, *http.Request)) ProviderServerOption {
 	return func(o *providerServerOption) {
 		o.handlers[path] = handler
 	}
 }
 
-// StartMockProviderServer starts a new HTTP server to be used as an OpenID Connect provider for tests.
+// StartMockProviderServer starts a new HTTP server to be used as an OIDC provider for tests.
 func StartMockProviderServer(address string, tokenHandlerOpts *TokenHandlerOptions, args ...ProviderServerOption) (string, func()) {
 	servMux := http.NewServeMux()
 	server := httptest.NewUnstartedServer(servMux)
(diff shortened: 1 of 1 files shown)


internal/broker/broker.go
```diff
@@ -57,8 +57,8 @@ type userConfig struct {
 type Broker struct {
 	cfg Config
 
-	providerInfo providers.ProviderInfoer
-	oidcCfg      oidc.Config
+	provider providers.Provider
```
denisonbarbosa, line 60: This is tricky... We would then have an oidcProvider and a Provider, both doing different things... I'm not in love with the name `ProviderInfoer` also, but I think naming the interface only as `Provider` could be annoying (@didrocks, IIRC, we tried that name before, right?)

internal/broker/broker.go
```diff
@@ -76,7 +76,7 @@ type session struct {
 	authModes         []string
 	attemptsPerMode   map[string]int
 
-	oidcProvider          *oidc.Provider
+	oidcServer            *oidc.Provider
```
denisonbarbosa, line 79: Ok, I see how you fixed the provider issue now...

However, is this name "accurate" enough? IIRC, the reason I used provider is because the term was homogeneous along all documentations (OpenID, OAuth, MS Entra, Google, etc.), so it wouldn't confuse external contributors/users.

internal/broker/broker_test.go
```diff
@@ -79,19 +79,19 @@ func TestNewSession(t *testing.T) {
 	t.Parallel()
 
 	tests := map[string]struct {
-		customHandlers map[string]testutils.ProviderHandler
+		customHandlers map[string]testutils.EndpointHandler
```
denisonbarbosa, line 82: This was a good one!

denisonbarbosa (changes requested):
I have some questions, but it's more for debating than change requests right on the go.

internal/broker/broker.go
```diff
@@ -166,7 +166,7 @@ func (b *Broker) NewSession(username, lang, mode string) (sessionID, encryptionK
 	defer decorate.OnError(&err, "could not create new session for user %q", username)
 
 	sessionID = uuid.New().String()
-	session := sessionInfo{
+	session := session{
```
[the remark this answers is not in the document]
adombeck: I renamed the local variable to `s`, which should still be clear enough since this is the `NewSession` function.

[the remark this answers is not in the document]
denisonbarbosa: Well, you solved the issue in one function... But there's a lot of cases of the same issue throughout the file.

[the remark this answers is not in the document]
adombeck: The only other cases I see are function arguments named `session`. I don't see a problem with those, because it's clear that in those functions the `session` identifier refers to the argument, not the type.

internal/broker/broker.go
adombeck: It's not perfect but better than what we had previously IMO. I think ideally, the `*oidc.Provider` should be part of the implementations of our own `Provider` interface, because they do belong together conceptually. That would require much more refactoring though.

I went with `oidcServer` because even the doc comment of the `oidc.Provider` struct says that it represents an OIDC server's configuration: https://github.com/coreos/go-oidc/blob/v3.11.0/oidc/oidc.go#L95-L95

Alternatively, we could name this field just `oidc` (and rename the `oidcCfg` field to `idTokenVerifierCfg`, because that's what it is according to the [doc comment](github.com/coreos/go-oidc/blob/v3.11.0/oidc/verify.go#L78-L78)).

denisonbarbosa (approved):

PR #232, opened 2024-11-19
Move golden files below testdata/golden

adombeck:
We used to store golden files in:

    testdata/TopLevelTestName/golden/SubTestName

We now store them in:

    testdata/golden/TopLevelTestName/SubTestName

That allows to easily copy/paste the full test name (top-level test name + subtest name) into a `go test --run` command.

UDENG-5448


internal/broker/broker_test.go
```diff
@@ -12,6 +12,7 @@ import (
 	"github.com/stretchr/testify/require"
 	"github.com/ubuntu/authd-oidc-brokers/internal/broker"
 	"github.com/ubuntu/authd-oidc-brokers/internal/broker/authmodes"
+	"github.com/ubuntu/authd-oidc-brokers/internal/golden"
```
denisonbarbosa, line 15: I can see the appeal here, but can you make it a subpackage of `testutils`? I don't think this belongs solely on `internal/`. This way, we can still relate to it as `golden.CheckOrUpdate`, but at least it's organized better.

internal/golden/golden.go
```diff
@@ -36,11 +37,11 @@ type goldenOptions struct {
 	path string
 }
 
-// GoldenOption is a supported option reference to change the golden files comparison.
-type GoldenOption func(*goldenOptions)
+// Option is a supported option reference to change the golden files comparison.
+type Option func(*goldenOptions)
```
denisonbarbosa, line 41: Probably rename the `goldenOptions` type to `options` also?

internal/golden/golden.go
```diff
@@ -350,7 +365,10 @@ func addEmptyMarker(p string) error {
 			if err != nil {
 				return err
 			}
-			f.Close()
+			err = f.Close()
+			if err != nil {
+				return err
+			}
```
denisonbarbosa, line 371: :eyes:  (you already know my comment...)

denisonbarbosa (commented):
Regarding the changes, I have some comments but nothing big.

However, I do have 2 observations:
1) Did you try reupdating the golden files to check if everything is ok? (I'm assuming you did, but it doesn't hurt to ask).

2) We need to maintain consistency between our projects. If we agree on this new structure, this also implies applying these changes to `authd`, `adsys`, WSL and so on.

internal/golden/golden.go
adombeck: I renamed it to `config` to avoid name collisions

internal/golden/golden.go
adombeck: fixed

adombeck:
> Did you try reupdating the golden files to check if everything is ok? (I'm assuming you did, but it doesn't hurt to ask).

Yes, all the golden files are updated.

> We need to maintain consistency between our projects. If we agree on this new structure, this also implies applying these changes to authd, adsys, WSL and so on.

Right. My proposal is to change the structure for the reason mentioned [above](https://github.com/ubuntu/authd-oidc-brokers/pull/232#issue-2671503837). If you or @didrocks (should I tag anyone else?) see any problems with that, I can split off the other changes into a separate PR. If I understood @didrocks correctly during the standup, he also prefers this directory structure.

didrocks:
> Right. My proposal is to change the structure for the reason mentioned https://github.com/ubuntu/authd-oidc-brokers/pull/232#issue-2671503837. If you or @didrocks (should I tag anyone else?) see any problems with that, I can split off the other changes into a separate PR. If I understood @didrocks correctly during the standup, he also prefers this directory structure.

Yeah, I think that ought to be 2 PRs, as we discussed in the HO, but that’s ok now. Just ensure there is a Jira card created  on moving most of `testutils/` (including this subpackage) to its own repo and list all the projects to move to it. That will make ofc backport harder in the futures when we can’t land the whole content, but that’s ok for now.

adombeck:
> > Right. My proposal is to change the structure for the reason mentioned [#232 (comment)](https://github.com/ubuntu/authd-oidc-brokers/pull/232#issue-2671503837). If you or @didrocks (should I tag anyone else?) see any problems with that, I can split off the other changes into a separate PR. If I understood @didrocks correctly during the standup, he also prefers this directory structure.
> 
> Yeah, I think that ought to be 2 PRs, as we discussed in the HO, but that’s ok now. Just ensure there is a Jira card created on moving most of `testutils/` (including this subpackage) to its own repo and list all the projects to move to it. That will make ofc backport harder in the futures when we can’t land the whole content, but that’s ok for now.

Tracked in UDENG-5363

internal/testutils/golden/golden.go
```diff
@@ -33,18 +33,18 @@ func init() {
 	}
 }
 
-type goldenOptions struct {
+type config struct {
```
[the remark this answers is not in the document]
adombeck: The type is not just used for the default config but also the config which is modified by the `Option` functions, so I prefer `config` over `defaults` (`default` is a reserved keyword so that can't be used anyway).

jibel (changes requested):
testing gh/jira reviewers sync.

PR #240, opened 2024-12-03
Support drop-in config files

adombeck:
We plan to use drop-in config files for the `allowed_users` and `owner` options (both for migration and generating the config file which sets the owner).

### DIFF
--- a/internal/broker/config.go
+++ b/internal/broker/config.go
@@ -3,6 +3,9 @@ package broker
 import (
 	"errors"
 	"fmt"
+	"io/fs"
+	"os"
+	"path/filepath"
 	"strings"
 
 	"gopkg.in/ini.v1"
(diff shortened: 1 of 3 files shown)


internal/broker/config.go
```diff
@@ -31,7 +34,27 @@ const (
...
+		if err != nil {
+			return cfg, err
+		}
+		for _, file := range files {
+			if file.IsDir() {
+				continue
+			}
+			dropInFiles = append(dropInFiles, filepath.Join(dropInDir, file.Name()))
+		}
+	}
```
3v1n0, line 55:
```suggestion
	files, err := os.ReadDir(dropInDir)
	if err != nil && !errors.Is(fs.ErrNotExist) {
	       return cfg, nil
	}
	var dropInFiles any
	for _, file := range files {
		if file.IsDir() {
			continue
		}
		dropInFiles = append(dropInFiles, filepath.Join(dropInDir, file.Name()))
	}
```

Otherwise instead of `fileutils.FileExists` can just use `os.Stat` and check also for being a dir.

adombeck: > files, err := os.ReadDir(dropInDir)
	if err != nil && !errors.Is(fs.ErrNotExist) {
	       return cfg, nil
	}

That has the effect that we ignore e.g. permission errors. The purpose of 

```
exists, err := fileutils.FileExists(dropInDir)
	if err != nil {
		return cfg, err
	}
```

is to fail in such a case, for example when the directory exists but can't be accessed.

3v1n0: Sure, but wouldn't `os.ReadDir(dropInDir)` fail anyways in such case?

adombeck: I think we don't understand each other. Maybe we can solve this better in a quick call?

internal/broker/config.go
```diff
@@ -27,11 +30,39 @@ const (
...
+	dropInDir := cfgPath + ".d"
+	files, err := os.ReadDir(dropInDir)
+	if errors.Is(err, fs.ErrNotExist) {
+		return nil, nil
+	}
+	if err != nil {
+		return nil, err
+	}
+
+	var dropInFiles []any
```
denisonbarbosa, line 44: Why are using a `[]any` here? `ini.Load` should accept a slice of strings, no?

adombeck: no it does not. the function signature is

```
func Load(source interface{}, others ...interface{}) (*File, error)
```

we would have to convert each element in the string slice into an `interface{}`.

adombeck:
@3v1n0 I added two error cases to the tests. Should I test anything else?

3v1n0 (approved):
Lovely, thanks!

PR #249, opened 2024-12-10
Add device ownership

nsklikas:
[IAM-1248](https://warthogs.atlassian.net/browse/IAM-1248)

Implement device ownership logic as described in `UD084`. This PR:
- Adds the new configuration options
- Implements the user filtering logic
- Implements owner autoregistration, if configured

Some things that I am not sure about are:

- [x] Should the server create the `.d` configuration dir or should that be created by the snap? If the server should create it, are the permissions correct?
- [x] What should the permissions for `20-owner-autoregistration.conf` be? (right now they are `0666`)

[IAM-1248]: https://warthogs.atlassian.net/browse/IAM-1248?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/cmd/authd-oidc/daemon/daemon.go
+++ b/cmd/authd-oidc/daemon/daemon.go
@@ -125,6 +125,12 @@ func (a *App) serve(config daemonConfig) error {
 		}
 	}
 
+	brokerConfigDir := broker.GetDropInDir(config.Paths.BrokerConf)
+	if err := ensureDirWithPerms(brokerConfigDir, 0700, os.Geteuid()); err != nil {
+		close(a.ready)
+		return fmt.Errorf("error initializing broker configuration directory %q: %v", brokerConfigDir, err)
+	}
+
 	b, err := broker.New(broker.Config{
 		ConfigFile:            config.Paths.BrokerConf,
 		DataDir:               config.Paths.DataDir,
--- a/conf/broker.conf
+++ b/conf/broker.conf
@@ -15,3 +15,31 @@ client_id = <CLIENT_ID>
 ## If configured, only users with a suffix in this list are allowed to
 ## log in via SSH. The suffixes must be separated by comma.
 #ssh_allowed_suffixes = @example.com,@anotherexample.com
+
+## 'allowed_users' specifies the users who are permitted to log in after
+## successfully authenticating with the Identity Provider.
+## Values are separated by commas. Supported values:
+## - 'OWNER': Grants access to the user specified in the 'owner' option
+##        	(see below). This is the default.
+## - 'ALL': Grants access to all users who successfully authenticate
+##      	with the Identity Provider.
+## - <username>: Grants access to specific additional users
+##           	(e.g. user1@example.com).
+## Example: allowed_users = OWNER,user1@example.com,admin@example.com
+#allowed_users = OWNER
+
+## 'owner' specifies the user assigned the owner role. This user is
+## permitted to log in if 'OWNER' is included in the 'allowed_users'
+## option.
+##
+## If this option is left unset, the first user to successfully log in
+## via this broker will automatically be assigned the owner role. A
+## drop-in configuration file will be created in broker.conf.d/ to set
+## the 'owner' option.
+##
+## To disable automatic assignment, you can either:
+## 1. Explicitly set this option to an empty value (e.g. owner = "")
+## 2. Remove 'OWNER' from the 'allowed_users' option
+##
+## Example: owner = user2@example.com
+#owner =
--- a/internal/broker/broker.go
+++ b/internal/broker/broker.go
@@ -49,8 +49,6 @@ type Config struct {
 type Broker struct {
 	cfg Config
 
-	setOwnerMutex sync.Mutex
-
 	provider providers.Provider
 	oidcCfg  oidc.Config
 
@@ -615,8 +613,7 @@ func (b *Broker) handleIsAuthenticated(ctx context.Context, session *session, au
 		}
 	}
 
-	err = b.maybeRegisterOwner(b.cfg.ConfigFile, authInfo.UserInfo.Name)
-	if err != nil {
+	if err := b.cfg.registerOwner(b.cfg.ConfigFile, authInfo.UserInfo.Name); err != nil {
 		// The user is not allowed if we fail to create the owner-autoregistration file.
 		// Otherwise the owner might change if the broker is restarted.
 		slog.Error(fmt.Sprintf("Failed to assign the owner role: %v", err))
@@ -643,19 +640,6 @@ func (b *Broker) handleIsAuthenticated(ctx context.Context, session *session, au
 	return AuthGranted, userInfoMessage{UserInfo: authInfo.UserInfo}
 }
 
-func (b *Broker) maybeRegisterOwner(cfgPath string, userName string) error {
-	// We need to lock here to avoid a race condition where two users log in at the same time, causing both to be
-	// considered the owner.
-	b.setOwnerMutex.Lock()
-	defer b.setOwnerMutex.Unlock()
-
-	if !b.cfg.userConfig.shouldRegisterOwner() {
-		return nil
-	}
-
-	return b.cfg.registerOwner(cfgPath, userName)
-}
-
 // userNameIsAllowed checks whether the user's username is allowed to access the machine.
 func (b *Broker) userNameIsAllowed(userName string) bool {
 	normalizedUsername := b.provider.NormalizeUsername(userName)
@@ -669,7 +653,8 @@ func (b *Broker) userNameIsAllowed(userName string) bool {
 	if _, ok := b.cfg.userConfig.allowedUsers[normalizedUsername]; ok {
 		return true
 	}
-	return b.cfg.userConfig.ownerAllowed && b.cfg.userConfig.owner == normalizedUsername
+
+	return b.cfg.isOwnerAllowed(normalizedUsername)
 }
 
 func (b *Broker) startAuthenticate(sessionID string) (context.Context, error) {
--- a/internal/broker/broker.go
+++ b/internal/broker/broker.go
@@ -364,7 +364,6 @@ func (b *Broker) generateUILayout(session *session, authModeID string) (map[stri
 		// Some providers support both of these authentication methods, some implement only one and
 		// some implement neither.
 		// This was tested with the following providers:
-		// - Google: supports client_secret_post
 		// - Ory Hydra: supports client_secret_post
 		// TODO @shipperizer: client_authentication methods should be configurable
 		if secret := session.oauth2Config.ClientSecret; secret != "" {
--- a/internal/broker/broker.go
+++ b/internal/broker/broker.go
@@ -98,15 +98,17 @@ type Option func(*option)
 func New(cfg Config, args ...Option) (b *Broker, err error) {
 	defer decorate.OnError(&err, "could not create broker")
 
+	p := providers.CurrentProvider()
+
 	if cfg.ConfigFile != "" {
-		cfg.userConfig, err = parseConfigFile(cfg.ConfigFile)
+		cfg.userConfig, err = parseConfigFile(cfg.ConfigFile, p)
 		if err != nil {
 			return nil, fmt.Errorf("could not parse config: %v", err)
 		}
 	}
 
 	opts := option{
-		provider: providers.CurrentProvider(),
+		provider: p,
 	}
 	for _, arg := range args {
 		arg(&opts)
@@ -644,6 +646,7 @@ func (b *Broker) handleIsAuthenticated(ctx context.Context, session *session, au
 
 // userNameIsAllowed checks whether the user's username is allowed to access the machine.
 func (b *Broker) userNameIsAllowed(userName string) bool {
+	normalizedUsername := b.provider.NormalizeUsername(userName)
 	// The user is allowed to log in if:
 	// - ALL users are allowed
 	// - the user's name is in the list of allowed_users
@@ -651,14 +654,14 @@ func (b *Broker) userNameIsAllowed(userName string) bool {
 	if b.cfg.userConfig.allUsersAllowed {
 		return true
 	}
-	if _, ok := b.cfg.userConfig.allowedUsers[userName]; ok {
+	if _, ok := b.cfg.userConfig.allowedUsers[normalizedUsername]; ok {
 		return true
 	}
 	if !b.cfg.userConfig.ownerAllowed {
 		return false
 	}
 	// If owner is undefined, then the first user to log in is considered the owner
-	return b.cfg.userConfig.firstUserBecomesOwner || b.cfg.userConfig.owner == userName
+	return b.cfg.userConfig.firstUserBecomesOwner || b.cfg.userConfig.owner == normalizedUsername
 }
 
 func (b *Broker) startAuthenticate(sessionID string) (context.Context, error) {
--- a/internal/broker/broker.go
+++ b/internal/broker/broker.go
@@ -49,6 +49,8 @@ type Config struct {
 type Broker struct {
 	cfg Config
 
+	setOwnerMutex sync.Mutex
+
 	provider providers.Provider
 	oidcCfg  oidc.Config
 
@@ -613,19 +615,16 @@ func (b *Broker) handleIsAuthenticated(ctx context.Context, session *session, au
 		}
 	}
 
-	if !b.userNameIsAllowed(authInfo.UserInfo.Name) {
-		return AuthDenied, errorMessage{Message: "permission denied"}
+	err = b.maybeRegisterOwner(b.cfg.ConfigFile, authInfo.UserInfo.Name)
+	if err != nil {
+		// The user is not allowed if we fail to create the owner-autoregistration file.
+		// Otherwise the owner might change if the broker is restarted.
+		slog.Error(fmt.Sprintf("Failed to assign the owner role: %v", err))
+		return AuthDenied, errorMessage{Message: "could not register the owner"}
 	}
 
-	// If the owner is unset and allowed, we auto-generate a config file with the first
-	// user to log in as the owner.
-	if b.cfg.userConfig.shouldRegisterOwner() {
-		if err := b.cfg.registerOwner(b.cfg.ConfigFile, authInfo.UserInfo.Name); err != nil {
-			// The user is not allowed, if we fail to create the owner-autoregistration file.
-			// Otherwise the owner might change if the broker is restarted.
-			slog.Error(fmt.Sprintf("Failed to assign the owner role: %v", err))
-			return AuthDenied, errorMessage{Message: "could not register the owner"}
-		}
+	if !b.userNameIsAllowed(authInfo.UserInfo.Name) {
+		return AuthDenied, errorMessage{Message: "permission denied"}
 	}
 
 	if session.isOffline {
@@ -644,6 +643,19 @@ func (b *Broker) handleIsAuthenticated(ctx context.Context, session *session, au
 	return AuthGranted, userInfoMessage{UserInfo: authInfo.UserInfo}
 }
 
+func (b *Broker) maybeRegisterOwner(cfgPath string, userName string) error {
+	// We need to lock here to avoid a race condition where two users log in at the same time, causing both to be
+	// considered the owner.
+	b.setOwnerMutex.Lock()
+	defer b.setOwnerMutex.Unlock()
+
+	if !b.cfg.userConfig.shouldRegisterOwner() {
+		return nil
+	}
+
+	return b.cfg.registerOwner(cfgPath, userName)
+}
+
 // userNameIsAllowed checks whether the user's username is allowed to access the machine.
 func (b *Broker) userNameIsAllowed(userName string) bool {
 	normalizedUsername := b.provider.NormalizeUsername(userName)
@@ -657,11 +669,7 @@ func (b *Broker) userNameIsAllowed(userName string) bool {
 	if _, ok := b.cfg.userConfig.allowedUsers[normalizedUsername]; ok {
 		return true
 	}
-	if !b.cfg.userConfig.ownerAllowed {
-		return false
-	}
-	// If owner is undefined, then the first user to log in is considered the owner
-	return b.cfg.userConfig.firstUserBecomesOwner || b.cfg.userConfig.owner == normalizedUsername
+	return b.cfg.userConfig.ownerAllowed && b.cfg.userConfig.owner == normalizedUsername
 }
 
 func (b *Broker) startAuthenticate(sessionID string) (context.Context, error) {
--- a/internal/broker/broker.go
+++ b/internal/broker/broker.go
@@ -611,6 +611,21 @@ func (b *Broker) handleIsAuthenticated(ctx context.Context, session *session, au
 		}
 	}
 
+	if !b.userNameIsAllowed(authInfo.UserInfo.Name) {
+		return AuthDenied, errorMessage{Message: "permission denied"}
+	}
+
+	// If the owner is unset and allowed, we auto-generate a config file with the first
+	// user to log in as the owner.
+	if b.cfg.userConfig.shouldRegisterOwner() {
+		if err := b.cfg.registerOwner(b.cfg.ConfigFile, authInfo.UserInfo.Name); err != nil {
+			// The user is not allowed, if we fail to create the owner-autoregistration file.
+			// Otherwise the owner might change if the broker is restarted.
+			slog.Error(fmt.Sprintf("Failed to assign the owner role: %v", err))
+			return AuthDenied, errorMessage{Message: "could not register the owner"}
+		}
+	}
+
 	if session.isOffline {
 		return AuthGranted, userInfoMessage{UserInfo: authInfo.UserInfo}
 	}
@@ -625,6 +640,25 @@ func (b *Broker) handleIsAuthenticated(ctx context.Context, session *session, au
 	token.CleanupOldEncryptedToken(session.oldEncryptedTokenPath)
 
 	return AuthGranted, userInfoMessage{UserInfo: authInfo.UserInfo}
+}
+
+// userNameIsAllowed checks whether the user's username is allowed to access the machine.
+func (b *Broker) userNameIsAllowed(userName string) bool {
+	// The user is allowed to log in if:
+	// - ALL users are allowed
+	// - the user's name is in the list of allowed_users
+	// - the user is the owner of the machine and OWNER is in the allowed_users list
+	if b.cfg.userConfig.allUsersAllowed {
+		return true
+	}
+	if _, ok := b.cfg.userConfig.allowedUsers[userName]; ok {
+		return true
+	}
+	if !b.cfg.userConfig.ownerAllowed {
+		return false
+	}
+	// If owner is undefined, then the first user to log in is considered the owner
+	return b.cfg.userConfig.firstUserBecomesOwner || b.cfg.userConfig.owner == userName
 }
 
 func (b *Broker) startAuthenticate(sessionID string) (context.Context, error) {
--- a/internal/broker/broker.go
+++ b/internal/broker/broker.go
@@ -46,11 +46,17 @@ type Config struct {
 }
 
 type userConfig struct {
-	clientID           string
-	clientSecret       string
-	issuerURL          string
-	homeBaseDir        string
-	allowedSSHSuffixes []string
+	clientID     string
+	clientSecret string
+	issuerURL    string
+
+	allowedUsers          map[string]struct{}
+	allUsersAllowed       bool
+	ownerAllowed          bool
+	firstUserBecomesOwner bool
+	owner                 string
+	homeBaseDir           string
+	allowedSSHSuffixes    []string
 }
 
 // Broker is the real implementation of the broker to track sessions and process oidc calls.
--- a/internal/broker/broker.go
+++ b/internal/broker/broker.go
@@ -45,20 +45,6 @@ type Config struct {
 	userConfig
 }
 
-type userConfig struct {
-	clientID     string
-	clientSecret string
-	issuerURL    string
-
-	allowedUsers          map[string]struct{}
-	allUsersAllowed       bool
-	ownerAllowed          bool
-	firstUserBecomesOwner bool
-	owner                 string
-	homeBaseDir           string
-	allowedSSHSuffixes    []string
-}
-
 // Broker is the real implementation of the broker to track sessions and process oidc calls.
 type Broker struct {
 	cfg Config
--- a/internal/broker/broker_test.go
+++ b/internal/broker/broker_test.go
@@ -878,6 +878,18 @@ func TestIsAuthenticatedAllowedUsersConfig(t *testing.T) {
 		wantAllowedUsers   []string
 		wantUnallowedUsers []string
 	}{
+		"No_users_allowed": {
+			wantUnallowedUsers: allUsers,
+		},
+		"No_users_allowed_when_owner_is_allowed_but_not_set": {
+			ownerAllowed:       true,
+			wantUnallowedUsers: allUsers,
+		},
+		"No_users_allowed_when_owner_is_set_but_not_allowed": {
+			owner:              u1,
+			wantUnallowedUsers: allUsers,
+		},
+
 		"All_users_are_allowed": {
 			allUsersAllowed:  true,
 			wantAllowedUsers: allUsers,
@@ -887,13 +899,6 @@ func TestIsAuthenticatedAllowedUsersConfig(t *testing.T) {
 			owner:              u1,
 			wantAllowedUsers:   []string{u1},
 			wantUnallowedUsers: []string{u2, u3},
-		},
-		"No_users_allowed": {
-			wantUnallowedUsers: allUsers,
-		},
-		"No_users_allowed_when_owner_is_allowed_but_not_set": {
-			ownerAllowed:       true,
-			wantUnallowedUsers: allUsers,
 		},
 		"Only_first_user_allowed": {
 			ownerAllowed:          true,
@@ -912,10 +917,6 @@ func TestIsAuthenticatedAllowedUsersConfig(t *testing.T) {
 			owner:              u2,
 			wantAllowedUsers:   []string{u1, u2},
 			wantUnallowedUsers: []string{u3},
-		},
-		"Owner_is_set_but_not_allowed": {
-			owner:              u1,
-			wantUnallowedUsers: allUsers,
 		},
 	}
 
--- a/internal/broker/broker_test.go
+++ b/internal/broker/broker_test.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"path/filepath"
 	"slices"
+	"strings"
 	"testing"
 	"time"
 
@@ -860,7 +861,7 @@ func TestIsAuthenticatedAllowedUsersConfig(t *testing.T) {
 
 	u1 := "u1"
 	u2 := "u2"
-	u3 := "u3"
+	u3 := "U3"
 	allUsers := []string{u1, u2, u3}
 
 	idTokenClaims := []map[string]interface{}{}
@@ -917,6 +918,13 @@ func TestIsAuthenticatedAllowedUsersConfig(t *testing.T) {
 			owner:              u2,
 			wantAllowedUsers:   []string{u1, u2},
 			wantUnallowedUsers: []string{u3},
+		},
+		"Usernames_are_normalized": {
+			ownerAllowed:       true,
+			allowedUsers:       map[string]struct{}{u3: {}},
+			owner:              strings.ToLower(u3),
+			wantAllowedUsers:   []string{u3},
+			wantUnallowedUsers: []string{u1, u2},
 		},
 	}
 
--- a/internal/broker/broker_test.go
+++ b/internal/broker/broker_test.go
@@ -857,6 +857,17 @@ func TestConcurrentIsAuthenticated(t *testing.T) {
 
 func TestIsAuthenticatedAllowedUsersConfig(t *testing.T) {
 	t.Parallel()
+
+	u1 := "u1"
+	u2 := "u2"
+	u3 := "u3"
+	allUsers := []string{u1, u2, u3}
+
+	idTokenClaims := []map[string]interface{}{}
+	for _, uname := range allUsers {
+		idTokenClaims = append(idTokenClaims, map[string]interface{}{"sub": "user", "name": "user", "email": uname})
+	}
+
 	tests := map[string]struct {
 		allowedUsers          map[string]struct{}
 		owner                 string
@@ -864,49 +875,47 @@ func TestIsAuthenticatedAllowedUsersConfig(t *testing.T) {
 		allUsersAllowed       bool
 		firstUserBecomesOwner bool
 
-		allowedUsernames   []string
-		unallowedUsernames []string
+		wantAllowedUsers   []string
+		wantUnallowedUsers []string
 	}{
 		"All_users_are_allowed": {
 			allUsersAllowed:  true,
-			allowedUsernames: []string{"u1", "u2"},
+			wantAllowedUsers: allUsers,
 		},
 		"Only_owner_allowed": {
 			ownerAllowed:       true,
-			owner:              "machine_owner",
-			allowedUsernames:   []string{"machine_owner"},
-			unallowedUsernames: []string{"u1", "u2"},
+			owner:              u1,
+			wantAllowedUsers:   []string{u1},
+			wantUnallowedUsers: []string{u2, u3},
 		},
 		"No_users_allowed": {
-			unallowedUsernames: []string{"u1", "u2", "machine_owner", "u3"},
+			wantUnallowedUsers: allUsers,
 		},
 		"No_users_allowed_when_owner_is_allowed_but_not_set": {
 			ownerAllowed:       true,
-			unallowedUsernames: []string{"u1", "u2", "machine_owner", "u3"},
+			wantUnallowedUsers: allUsers,
 		},
 		"Only_first_user_allowed": {
 			ownerAllowed:          true,
 			firstUserBecomesOwner: true,
-			allowedUsernames:      []string{"random_user"},
-			unallowedUsernames:    []string{"u1", "u2"},
+			wantAllowedUsers:      []string{u1},
+			wantUnallowedUsers:    []string{u2, u3},
 		},
 		"Specific_users_allowed": {
-			allowedUsers:       map[string]struct{}{"u1": {}, "u2": {}},
-			allowedUsernames:   []string{"u1", "u2"},
-			unallowedUsernames: []string{"u3"},
+			allowedUsers:       map[string]struct{}{u1: {}, u2: {}},
+			wantAllowedUsers:   []string{u1, u2},
+			wantUnallowedUsers: []string{u3},
 		},
 		"Specific_users_and_owner": {
-			allowedUsers:       map[string]struct{}{"u1": {}, "u2": {}},
 			ownerAllowed:       true,
-			owner:              "machine_owner",
-			allowedUsernames:   []string{"u1", "u2", "machine_owner"},
-			unallowedUsernames: []string{"u3"},
+			allowedUsers:       map[string]struct{}{u1: {}},
+			owner:              u2,
+			wantAllowedUsers:   []string{u1, u2},
+			wantUnallowedUsers: []string{u3},
 		},
-		"Owner_is_disabled_even_when_registered": {
-			allowedUsers:       map[string]struct{}{"u1": {}, "u2": {}},
-			owner:              "machine_owner",
-			allowedUsernames:   []string{"u1", "u2"},
-			unallowedUsernames: []string{"machine_owner", "u3"},
+		"Owner_is_set_but_not_allowed": {
+			owner:              u1,
+			wantUnallowedUsers: allUsers,
 		},
 	}
 
@@ -917,14 +926,6 @@ func TestIsAuthenticatedAllowedUsersConfig(t *testing.T) {
 			dataDir := filepath.Join(outDir, "data")
 			err := os.Mkdir(dataDir, 0700)
 			require.NoError(t, err, "Setup: Mkdir should not have returned an error")
-
-			idTokenClaims := []map[string]interface{}{}
-			for _, uname := range tc.allowedUsernames {
-				idTokenClaims = append(idTokenClaims, map[string]interface{}{"sub": "user", "name": "user", "email": uname})
-			}
-			for _, uname := range tc.unallowedUsernames {
-				idTokenClaims = append(idTokenClaims, map[string]interface{}{"sub": "user", "name": "user", "email": uname})
-			}
 
 			b := newBrokerForTests(t, &brokerForTestConfig{
 				Config:                broker.Config{DataDir: dataDir},
@@ -938,7 +939,7 @@ func TestIsAuthenticatedAllowedUsersConfig(t *testing.T) {
 				},
 			})
 
-			for _, u := range append(tc.allowedUsernames, tc.unallowedUsernames...) {
+			for _, u := range allUsers {
 				sessionID, key := newSessionForTests(t, b, u, "")
 				token := tokenOptions{username: u}
 				generateAndStoreCachedInfo(t, token, b.TokenPathForSession(sessionID))
@@ -952,11 +953,15 @@ func TestIsAuthenticatedAllowedUsersConfig(t *testing.T) {
 				access, data, err := b.IsAuthenticated(sessionID, authData)
 				require.True(t, json.Valid([]byte(data)), "IsAuthenticated returned data must be a valid JSON")
 				require.NoError(t, err)
-				if slices.Contains(tc.allowedUsernames, u) {
+				if slices.Contains(tc.wantAllowedUsers, u) {
 					require.Equal(t, access, broker.AuthGranted, "authentication failed")
-				} else {
-					require.Equal(t, access, broker.AuthDenied, "authentication failed")
+					continue
 				}
+				if slices.Contains(tc.wantUnallowedUsers, u) {
+					require.Equal(t, access, broker.AuthDenied, "authentication failed")
+					continue
+				}
+				t.Fatalf("user %s is not in the allowed or unallowed users list", u)
 			}
 		})
 	}
--- a/internal/broker/broker_test.go
+++ b/internal/broker/broker_test.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"slices"
 	"testing"
 	"time"
 
@@ -536,8 +537,10 @@ func TestIsAuthenticated(t *testing.T) {
 			require.NoError(t, err, "Setup: Mkdir should not have returned an error")
 
 			cfg := &brokerForTestConfig{
-				Config:           broker.Config{DataDir: dataDir},
-				getUserInfoFails: tc.getUserInfoFails,
+				Config:                broker.Config{DataDir: dataDir},
+				getUserInfoFails:      tc.getUserInfoFails,
+				ownerAllowed:          true,
+				firstUserBecomesOwner: true,
 			}
 			if tc.customHandlers == nil {
 				// Use the default provider URL if no custom handlers are provided.
@@ -707,14 +710,36 @@ func TestIsAuthenticated(t *testing.T) {
 // Due to ordering restrictions, this test can not be run in parallel, otherwise the routines would not be ordered as expected.
 func TestConcurrentIsAuthenticated(t *testing.T) {
 	tests := map[string]struct {
-		firstCallDelay  int
-		secondCallDelay int
+		firstCallDelay        int
+		secondCallDelay       int
+		ownerAllowed          bool
+		allUsersAllowed       bool
+		firstUserBecomesOwner bool
 
 		timeBetween time.Duration
 	}{
-		"First_auth_starts_and_finishes_before_second":                  {secondCallDelay: 1, timeBetween: 2 * time.Second},
-		"First_auth_starts_first_but_second_finishes_first":             {firstCallDelay: 3, timeBetween: time.Second},
-		"First_auth_starts_first_then_second_starts_and_first_finishes": {firstCallDelay: 2, secondCallDelay: 3, timeBetween: time.Second},
+		"First_auth_starts_and_finishes_before_second": {
+			secondCallDelay: 1,
+			timeBetween:     2 * time.Second,
+			allUsersAllowed: true,
+		},
+		"First_auth_starts_first_but_second_finishes_first": {
+			firstCallDelay:  3,
+			timeBetween:     time.Second,
+			allUsersAllowed: true,
+		},
+		"First_auth_starts_first_then_second_starts_and_first_finishes": {
+			firstCallDelay:  2,
+			secondCallDelay: 3,
+			timeBetween:     time.Second,
+			allUsersAllowed: true,
+		},
+		"First_auth_starts_first_but_second_finishes_first_and_is_registered_as_the_owner": {
+			firstCallDelay:        3,
+			timeBetween:           time.Second,
+			ownerAllowed:          true,
+			firstUserBecomesOwner: true,
+		},
 	}
 
 	for name, tc := range tests {
@@ -728,9 +753,12 @@ func TestConcurrentIsAuthenticated(t *testing.T) {
 			username2 := "user2@example.com"
 
 			b := newBrokerForTests(t, &brokerForTestConfig{
-				Config:          broker.Config{DataDir: dataDir},
-				firstCallDelay:  tc.firstCallDelay,
-				secondCallDelay: tc.secondCallDelay,
+				Config:                broker.Config{DataDir: dataDir},
+				allUsersAllowed:       tc.allUsersAllowed,
+				ownerAllowed:          tc.ownerAllowed,
+				firstUserBecomesOwner: tc.firstUserBecomesOwner,
+				firstCallDelay:        tc.firstCallDelay,
+				secondCallDelay:       tc.secondCallDelay,
 				tokenHandlerOptions: &testutils.TokenHandlerOptions{
 					IDTokenClaims: []map[string]interface{}{
 						{"sub": "user1", "name": "user1", "email": username1},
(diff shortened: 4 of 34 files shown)


internal/broker/export_test.go
```diff
@@ -19,6 +19,10 @@ func (cfg *Config) SetHomeBaseDir(homeBaseDir string) {
 	cfg.homeBaseDir = homeBaseDir
 }
 
+func (cfg *Config) SetAllowedUsers(allowedUsers map[string]bool) {
```
nsklikas, line 22: It looks like these methods are added to the `Config` struct that is defined and used in `broker.go`, is this intentional?

adombeck: Yes, that is intentional. These methods are defined in `export_test.go` instead of `broker.go` because they should only be used in tests.

adombeck:
Hi @nsklikas! There are some failing tests (https://github.com/ubuntu/authd-oidc-brokers/actions/runs/12256799287/job/34192844158?pr=249) and some linter errors (https://github.com/ubuntu/authd-oidc-brokers/actions/runs/12256799287?pr=249). Please resolve those before marking the PR ready for review. Thanks!

nsklikas:
Hello @adombeck, thank you for the very quick response. Sorry for this, all checks are passing now.

BarcoMasile (changes requested):

cmd/authd-oidc/daemon/daemon.go
```diff
@@ -125,6 +125,16 @@ func (a *App) serve(config daemonConfig) error {
...
+	brokerConfigDir := config.Paths.BrokerConf + ".d"
+	// Should these dirs be created by the server or the snap?
+	// TODO(nsklikas): are these permissions correct?
+	if err := ensureDirWithPerms(brokerConfigDir, 0700, os.Geteuid()); err != nil {
+		if err := ensureDirWithPerms(brokerConfigDir, 0755, os.Geteuid()); err != nil {
+			close(a.ready)
+			return fmt.Errorf("error initializing broker configuration directory %q: %v", brokerConfigDir, err)
+		}
+	}
+
```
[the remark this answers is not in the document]
nsklikas: If this is true then shouldn't [this](https://github.com/ubuntu/authd-oidc-brokers/blob/main/cmd/authd-oidc/daemon/daemon.go#L122) be changed as well?

internal/broker/config.go
```diff
@@ -24,15 +27,53 @@ const (
 
 	// usersSection is the section name in the config file for the users and broker specific configuration.
 	usersSection = "users"
+	// alloweedUsersKey is the key in the config file for the users that are allowed to access the machine.
```
shipperizer, line 30:
```suggestion
	// allowedUsersKey is the key in the config file for the users that are allowed to access the machine.
```

internal/broker/config.go
```diff
@@ -24,15 +27,53 @@ const (
...
+	allowedUsersKey = "allowed_users"
+	// alloweedUsersKey is the key in the config file for the users that are allowed to access the machine.
```
shipperizer, line 32:
```suggestion
	// ownerKey is......
```

internal/broker/config.go
```diff
@@ -53,6 +94,38 @@ func getDropInFiles(cfgPath string) ([]any, error) {
...
+
+	cfg.homeBaseDir = users.Key(homeDirKey).String()
+	cfg.allowedSSHSuffixes = strings.Split(users.Key(sshSuffixesKey).String(), ",")
+	// We need to differentiate unset owner from empty owner.
+	// - Unset means that the owner will be autoregistered
+	// - Empty means that there is no owner
+	if users.HasKey(ownerKey) {
+		o := users.Key(ownerKey).String()
+		cfg.owner = &o
+	}
```
shipperizer, line 113: this feels a bit counterintuitive as we checked already and put OWNER, wondering if we should have a flag instead saying "no owner so first registration needed" and leave the list empty?

just throwing it out there

adombeck (commented):
I'm not completely done with the review, but I'm submitting the comments I made so far

internal/broker/config.go
```diff
@@ -86,11 +159,60 @@ func parseConfigFile(cfgPath string) (userConfig, error) {
...
+	uc.owner = &owner
+}
+
+func (uc *userConfig) PersistOwner(cfgPath, userName string) error {
+	p := filepath.Join(getDropInDir(cfgPath), ownerRegistrationConfigPath)
+
+	templateName := strings.SplitN(ownerRegistrationConfigTemplate, "/", 2)[1]
+	t, err := template.New(templateName).ParseFS(ownerRegistrationConfig, ownerRegistrationConfigTemplate)
+	if err != nil {
+		slog.Error(fmt.Sprintf("Failed to open autoregastration template: %v", err))
```
shipperizer, line 201:
```suggestion
		slog.Error(fmt.Sprintf("Failed to open autoregistration template: %v", err))
```

cmd/authd-oidc/daemon/daemon.go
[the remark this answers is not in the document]
adombeck: > If this is true, then shouldn't [this](https://github.com/ubuntu/authd-oidc-brokers/blob/main/cmd/authd-oidc/daemon/daemon.go#L122) be changed as well?

No. The reason we call `ensureDirWithPerms` there twice is not that it requires different permissions to create a directory with 0700 than with 0755 (it does not), but that we want to create the data directory with 0700 by default but also support it having permissions 0755 (which is the case when it's the [`SNAP_DATA`](https://snapcraft.io/docs/data-locations#heading--system) directory, because that's created with `0755` by snap). We don't need to support that for the `broker.conf.d` directory.

internal/broker/broker.go
```diff
@@ -636,6 +631,40 @@ func (b *Broker) handleIsAuthenticated(ctx context.Context, session *session, au
 	return AuthGranted, userInfoMessage{UserInfo: authInfo.UserInfo}
 }
 
+// userNameIsAllowed checks whether the user's username is allowed to access the machine.
+func (b *Broker) userNameIsAllowed(userName string) bool {
+	// The user is allowed to login if:
+	// - ALL users are allowed
+	// - the user's name is in the list of allowed_users
+	// - the user is the owner of the machine
```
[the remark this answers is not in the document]
adombeck: I disagree, explaining the intend of a function in a comment doesn't only make it easier for reviewers but also for whoever else reads the code in the future.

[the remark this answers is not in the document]
3v1n0: I personally don't care much of comments when commit messages explain the rationale well enough (really, I comment-less code is more than fine when `git blame` can be your friend), but given this is _not_ the case (that implies: please be more elaborated in them), I feel it's better to keep it too.

internal/broker/broker.go
```diff
@@ -638,6 +631,37 @@ func (b *Broker) handleIsAuthenticated(ctx context.Context, session *session, au
...
+	allowed := false
+	if b.cfg.userConfig.AllUsersAllowed() {
+		allowed = true
+	} else if b.cfg.userConfig.IsUserAllowed(userName) {
+		allowed = true
+	} else if b.cfg.userConfig.OwnerUserAllowed() {
+		// If owner is undefined, then the first user to login is considered the owner
+		if b.cfg.userConfig.OwnerIsUnset() || *b.cfg.userConfig.Owner() == userName {
```
3v1n0, line 647: `*b.cfg.userConfig.Owner()` shouldn't we check for `nil` ptr too?

internal/broker/broker.go
```diff
@@ -638,6 +631,37 @@ func (b *Broker) handleIsAuthenticated(ctx context.Context, session *session, au
...
+			allowed = true
+		}
+	}
```
3v1n0, line 650: I feel this code can be put in a single function where you follow the happy-path better with early returns.

internal/broker/broker.go
```diff
@@ -638,6 +631,37 @@ func (b *Broker) handleIsAuthenticated(ctx context.Context, session *session, au
...
+
+	// If the owner is unset and allowed, we auto-generate a config file with the first
+	// user to login as the owner.
+	if b.cfg.userConfig.OwnerUserAllowed() && b.cfg.userConfig.OwnerIsUnset() {
+		if b.cfg.ConfigFile != "" {
+			// TODO(nsklikas): What should we do in case of error?
```
3v1n0, line 656: Opening a discussion here, since this should be handled.

internal/broker/broker_test.go
```diff
@@ -22,6 +22,10 @@ import (
 
 var defaultIssuerURL string
 
+func pointerTo[T ~string](s T) *T {
+	return &s
+}
```
3v1n0, line 27: In `authd` itself we use this instead:

```go
func ptrValue[T any](value T) *T {
	return &value
}
```

So for sake of consistency between our projects, please reuse the same.

internal/broker/broker.go
nsklikas: After our sync today, it was decided that authn should fail in this case. I will update it.

internal/broker/broker.go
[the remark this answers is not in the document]
adombeck: note: meta discussion which is not relevant to this PR

> I personally don't care much of comments when commit messages explain the rationale well enough (really, I comment-less code is more than fine when git blame can be your friend)

I think commit messages should explain the rationale in more detail than code comments, but IMO they don't replace code comments, because it requires significantly more work to view the commit message which originally introduced a change when the code has been modified many times since.

internal/broker/config_test.go
```diff
@@ -150,3 +150,242 @@ func TestParseConfig(t *testing.T) {
...
+[oidc]
+issuer = https://issuer.url.com
+client_id = client_id
+
+[users]
+allowed_users = ALL
+owner = machine_owner
+home_base_dir = /home
+allowed_ssh_suffixes = @issuer.url.com
+			`,
```
adombeck, line 172: For consistency, lets extract the config that's used in the test cases into a map, like it's done [here](https://github.com/nsklikas/authd-oidc-brokers/blob/90573fc10b23e7435dee63b41d73d74cc9efcd3b/internal/broker/config_test.go#L16).

internal/broker/config_test.go
```diff
@@ -150,3 +150,239 @@ func TestParseConfig(t *testing.T) {
 		})
 	}
 }
+
+func TestParseUserConfig(t *testing.T) {
+	t.Parallel()
+	tests := map[string]struct {
+		config string
+		check  func(userConfig)
```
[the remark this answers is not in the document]
nsklikas: Should these persisted to golden files or simply checking them in the golang code is enough?

nsklikas:
I tried to apply all the changes proposed, I did the following manual tests:
- Clean install with default config and no 20-autoregistration file, populates 20-autoregistration file
- Install the version on current `main`, configure, login, install the version from this branch and login without changing any of the config populates the 20-autoregistration file
- Login when `allowed_users=ALL` and no owner set with `nikos.sklikas@canonical.com` succeeds and does not populate 20-autoregistration file
- Login when `allowed_users=ALL,OWNER` and unset owner with `nikos.sklikas@canonical.com` succeeds and populates 20-autoregistration file
- Login when `owner=nikos.sklikas@canonical.com` and `allowed_users=a@canonical.com,OWNER` with `nikos.sklikas@canonical.com` succeeds
- Login when `owner=nikos.sklikas@canonical.com` and `allowed_users=a@canonical.com` with `nikos.sklikas@canonical.com` fails with `authentication failure: permission denied`
- Login when `allowed_users=OWNER` and `owner=` fails for any user

I went with @adombeck's suggestion, but I am not adding the owner in the allowedUsers list because I think that in the future we may want to differentiate between regular users and the owner. Also tried to add more info the commit messages (cc @3v1n0)

internal/broker/config.go
```diff
@@ -1,9 +1,12 @@
 package broker
```
adombeck, line 1: I noticed that [conf/broker.conf](https://github.com/nsklikas/authd-oidc-brokers/blob/c529e64daae63b823d13e58ed1a138fdc111efaf/conf/broker.conf) hasn't been updated. The new options should be added to it in the user section as specified in the spec:

```
[users]
## 'allowed_users' specifies the users who are permitted to log in after
## successfully authenticating with the Identity Provider.
## Values are separated by commas. Supported values:
## - 'OWNER': Grants access to the user specified in the 'owner' option
##        	(see below). This is the default.
## - 'ALL': Grants access to all users who successfully authenticate
##      	with the Identity Provider.
## - <username>: Grants access to specific additional users
##           	(e.g. user1@example.com).
## Example: allowed_users = OWNER,user1@example.com,admin@example.com
#allowed_users = OWNER

## 'owner' specifies the user assigned the owner role. This user is
## permitted to log in if 'OWNER' is included in the 'allowed_users'
## option.
##
## If this option is left unset, the first user to successfully log in
## via this broker will automatically be assigned the owner role. A
## drop-in configuration file will be created in broker.conf.d/ to set
## the 'owner' option.
##
## To disable automatic assignment, you can either:
## 1. Explicitly set this option to an empty value (e.g. owner = "")
## 2. Remove 'OWNER' from the 'allowed_users' option
##
## Example: owner = user2@example.com
#owner =
```

nsklikas: Good catch, thanks!

internal/broker/templates/20-owner-autoregistration.conf.tmpl
```diff
@@ -0,0 +1,13 @@
+## This file was generated automatically by the broker. DO NOT EDIT.
```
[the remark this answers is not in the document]
adombeck: I think the file timestamps are sufficient, we don't need to add one to the content

internal/broker/config.go
```diff
@@ -53,6 +96,45 @@ func getDropInFiles(cfgPath string) ([]any, error) {
...
+				cfg.firstUserBecomesOwner = true
+			}
+		} else {
+			cfg.allowedUsers[user] = true
+		}
+	}
+
+	// We need to read the owner key before we call HasKey, because the key is created
+	// after we call the "Key" function and we can't distinguish between empty and unset.
+	cfg.owner = users.Key(ownerKey).String()
```
adombeck, line 135: I don't understand this comment.

> We need to read the owner key before we call HasKey

`users.HasKey(ownerKey)` is called above, so we don't read the owner key before we call `HasKey`?

> because the key is created after we call the "Key"

What does "created after" mean? Is it created *when* we call the `Key` method or what causes it to be created?

internal/broker/broker.go
```diff
@@ -636,6 +631,44 @@ func (b *Broker) handleIsAuthenticated(ctx context.Context, session *session, au
 	return AuthGranted, userInfoMessage{UserInfo: authInfo.UserInfo}
 }
 
+// userNameIsAllowed checks whether the user's username is allowed to access the machine.
+func (b *Broker) userNameIsAllowed(userName string) bool {
+	// The user is allowed to login if:
```
adombeck, line 636: "login" is the noun, "log in" is the verb

```suggestion
	// The user is allowed to log in if:
```

internal/broker/broker.go
```diff
@@ -636,6 +631,44 @@ func (b *Broker) handleIsAuthenticated(ctx context.Context, session *session, au
...
+		return true
+	} else if _, ok := b.cfg.userConfig.allowedUsers[userName]; ok {
+		return true
+	} else if b.cfg.userConfig.ownerAllowed {
+		// If owner is undefined, then the first user to login is considered the owner
+		if b.cfg.userConfig.firstUserBecomesOwner || b.cfg.userConfig.owner == userName {
+			return true
+		}
+	}
+	return false
```
adombeck, line 650: We tend to avoid the `else` when it's not needed (because of the `return` in the `if` block)

```suggestion
	if b.cfg.userConfig.allUsersAllowed {
		return true
	}

	if _, ok := b.cfg.userConfig.allowedUsers[userName]; ok {
		return true
	}

	if b.cfg.userConfig.ownerAllowed &&
		(b.cfg.userConfig.firstUserBecomesOwner || b.cfg.userConfig.owner == userName) {
		return true
	}

	return false
}
```
```

File: internal/broker/config.go [hunk 19]
```diff
@@ -86,11 +168,50 @@ func parseConfigFile(cfgPath string) (userConfig, error) {
...
-	if users != nil {
-		cfg.homeBaseDir = users.Key(homeDirKey).String()
-		cfg.allowedSSHSuffixes = strings.Split(users.Key(sshSuffixesKey).String(), ",")
-	}
+	populateUsersConfig(&cfg, iniCfg.Section(usersSection))
 
 	return cfg, nil
 }
+
+func (uc *userConfig) ShouldRegisterOwner() bool {
```
adombeck, line 176: This doesn't need to be exported

```suggestion
func (uc *userConfig) shouldRegisterOwner() bool {
```

internal/broker/config.go
```diff
@@ -86,11 +168,50 @@ func parseConfigFile(cfgPath string) (userConfig, error) {
...
+	return uc.ownerAllowed && uc.firstUserBecomesOwner && uc.owner == ""
+}
+
+func (uc *userConfig) RegisterOwner(cfgPath, userName string) error {
```
adombeck, line 180: This also doesn't need to be exported

```suggestion
func (uc *userConfig) registerOwner(cfgPath, userName string) error {
```

internal/broker/broker.go
```diff
@@ -636,6 +631,44 @@ func (b *Broker) handleIsAuthenticated(ctx context.Context, session *session, au
...
+}
+
+// allowUser checks whether the user's username is allowed and registers the owner if necessary.
+func (b *Broker) allowUser(userName string) bool {
+	allowed := b.userNameIsAllowed(userName)
+
+	// If the owner is unset and allowed, we auto-generate a config file with the first
+	// user to login as the owner.
+	if allowed && b.cfg.userConfig.ShouldRegisterOwner() {
+		err := b.cfg.RegisterOwner(b.cfg.ConfigFile, userName)
```
adombeck, line 660: I wouldn't expect a method called `allowUser` to also register the user as the owner. I would prefer to move this to the caller (and then we can remove the `allowUser` method and let the caller call `userNameIsAllowed`  directly):

```diff
diff --git a/internal/broker/broker.go b/internal/broker/broker.go
index d33d794..c86c6bf 100644
--- a/internal/broker/broker.go
+++ b/internal/broker/broker.go
@@ -611,10 +611,22 @@ func (b *Broker) handleIsAuthenticated(ctx context.Context, session *session, au
 		}
 	}
 
-	if !b.allowUser(authInfo.UserInfo.Name) {
+	if !b.userNameIsAllowed(authInfo.UserInfo.Name) {
 		return AuthDenied, errorMessage{Message: "permission denied"}
 	}
 
+	// If the owner is unset and allowed, we auto-generate a config file with the first
+	// user to login as the owner.
+	if b.cfg.userConfig.ShouldRegisterOwner() {
+		err := b.cfg.RegisterOwner(b.cfg.ConfigFile, authInfo.UserInfo.Name)
+		if err != nil {
+			// The user is not allowed, if we fail to create the owner-autoregistration file.
+			// Otherwise the owner might change if the broker is restarted.
+			slog.Error(fmt.Sprintf("Failed to assign the owner role: %v", err))
+			return AuthDenied, errorMessage{Message: "could not assign the owner role"}
+		}
+	}
+
 	if session.isOffline {
 		return AuthGranted, userInfoMessage{UserInfo: authInfo.UserInfo}
 	}
@@ -650,25 +662,6 @@ func (b *Broker) userNameIsAllowed(userName string) bool {
 	return false
 }
 
-// allowUser checks whether the user's username is allowed and registers the owner if necessary.
-func (b *Broker) allowUser(userName string) bool {
-	allowed := b.userNameIsAllowed(userName)
-
-	// If the owner is unset and allowed, we auto-generate a config file with the first
-	// user to login as the owner.
-	if allowed && b.cfg.userConfig.ShouldRegisterOwner() {
-		err := b.cfg.RegisterOwner(b.cfg.ConfigFile, userName)
-		if err != nil {
-			// The user is not allowed, if we fail to create the owner-autoregistration file.
-			// Otherwise the owner might change if the broker is restarted.
-			slog.Error(fmt.Sprintf("Failed to auto-register the owner: %f", err))
-			return false
-		}
-	}
-
-	return allowed
-}
-
 func (b *Broker) startAuthenticate(sessionID string) (context.Context, error) {
 	session, err := b.getSession(sessionID)
 	if err != nil {
```

internal/broker/broker_test.go
```diff
@@ -827,6 +830,166 @@ func TestConcurrentIsAuthenticated(t *testing.T) {
...
+				if _, err := os.Stat(b.TokenPathForSession(sessionID)); err == nil {
+					err := os.WriteFile(b.TokenPathForSession(sessionID), []byte("Definitely an encrypted token"), 0600)
+					require.NoError(t, err, "Teardown: Failed to write generic token file")
+				}
+				passwordPath := b.PasswordFilepathForSession(sessionID)
+				if _, err := os.Stat(passwordPath); err == nil {
+					err := os.WriteFile(passwordPath, []byte("Definitely a hashed password"), 0600)
+					require.NoError(t, err, "Teardown: Failed to write generic password file")
+				}
+			}
```
adombeck, line 975: We can avoid some code duplication here by iterating over the allowed and unallowed users in the same loop:

```diff
diff --git a/internal/broker/broker_test.go b/internal/broker/broker_test.go
index 55de0d0..4c9d017 100644
--- a/internal/broker/broker_test.go
+++ b/internal/broker/broker_test.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"slices"
 	"testing"
 	"time"
 
@@ -891,13 +892,16 @@ func TestIsAuthenticatedAllowedUsersConfig(t *testing.T) {
 			err := os.Mkdir(dataDir, 0700)
 			require.NoError(t, err, "Setup: Mkdir should not have returned an error")
 			var session string
+			var users []string
 
 			idTokenClaims := []map[string]interface{}{}
 			for _, uname := range tc.allowedUsernames {
 				idTokenClaims = append(idTokenClaims, map[string]interface{}{"sub": "user", "name": "user", "email": uname})
+				users = append(users, uname)
 			}
 			for _, uname := range tc.unallowedUsernames {
 				idTokenClaims = append(idTokenClaims, map[string]interface{}{"sub": "user", "name": "user", "email": uname})
+				users = append(users, uname)
 			}
 
 			b := newBrokerForTests(t, &brokerForTestConfig{
@@ -912,7 +916,7 @@ func TestIsAuthenticatedAllowedUsersConfig(t *testing.T) {
 				},
 			})
 
-			for _, u := range tc.allowedUsernames {
+			for _, u := range users {
 				sessionID, key := newSessionForTests(t, b, u, "")
 				session = sessionID
 				token := tokenOptions{username: u}
@@ -927,7 +931,13 @@ func TestIsAuthenticatedAllowedUsersConfig(t *testing.T) {
 				access, data, err := b.IsAuthenticated(sessionID, authData)
 				require.True(t, json.Valid([]byte(data)), "IsAuthenticated returned data must be a valid JSON")
 				require.NoError(t, err)
-				require.NotEqual(t, access, broker.AuthDenied, "authentication failed")
+
+				if slices.Contains(tc.allowedUsernames, u) {
+					require.Equal(t, access, broker.AuthGranted, "authentication failed")
+				} else {
+					require.Equal(t, access, broker.AuthDenied, "authentication failed")
+				}
+
 				got := isAuthenticatedResponse{Access: access, Data: data, Err: fmt.Sprint(err)}
 				out, err := yaml.Marshal(got)
 				require.NoError(t, err, "Failed to marshal first response")
@@ -946,33 +956,6 @@ func TestIsAuthenticatedAllowedUsersConfig(t *testing.T) {
 					require.NoError(t, err, "Teardown: Failed to write generic password file")
 				}
 			}
-			for _, u := range tc.unallowedUsernames {
-				sessionID, key := newSessionForTests(t, b, u, "")
-				session = sessionID
-				token := tokenOptions{username: u}
-				generateAndStoreCachedInfo(t, token, b.TokenPathForSession(sessionID))
-				err = password.HashAndStorePassword("password", b.PasswordFilepathForSession(sessionID))
-				require.NoError(t, err, "Setup: HashAndStorePassword should not have returned an error")
-
-				updateAuthModes(t, b, sessionID, authmodes.Password)
-
-				authData := `{"challenge":"` + encryptChallenge(t, "password", key) + `"}`
-
-				access, data, err := b.IsAuthenticated(sessionID, authData)
-				require.True(t, json.Valid([]byte(data)), "IsAuthenticated returned data must be a valid JSON")
-				require.NoError(t, err)
-				require.Equal(t, access, broker.AuthDenied, "authentication succeeded")
-
-				if _, err := os.Stat(b.TokenPathForSession(sessionID)); err == nil {
-					err := os.WriteFile(b.TokenPathForSession(sessionID), []byte("Definitely an encrypted token"), 0600)
-					require.NoError(t, err, "Teardown: Failed to write generic token file")
-				}
-				passwordPath := b.PasswordFilepathForSession(sessionID)
-				if _, err := os.Stat(passwordPath); err == nil {
-					err := os.WriteFile(passwordPath, []byte("Definitely a hashed password"), 0600)
-					require.NoError(t, err, "Teardown: Failed to write generic password file")
-				}
-			}
 
 			if session != "" {
 				issuerDataDir := filepath.Dir(b.UserDataDirForSession(session))
```

internal/broker/broker.go
nsklikas: The reason for not placing this logic in `handleIsAuthenticated` is that I didn't want to make an already huge function (that IMO is too complicated), even more more complicated. But since that is what you want, I will move it.

internal/broker/broker.go
nsklikas: lol thank you for the clarification, I know the difference. Maybe I am a little dyslexic after all.

internal/broker/config.go
nsklikas: yes, https://github.com/go-ini/ini/blob/main/section.go#L171

internal/broker/config_test.go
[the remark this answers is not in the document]
adombeck: the go code is enough :)

internal/broker/broker.go
adombeck: Sorry, I didn't mean to imply that you don't know that. It's a very common mistake - in fact, we already used it incorrectly in the `broker.conf` file before (https://github.com/ubuntu/authd-oidc-brokers/pull/250).

internal/broker/broker.go
adombeck: I agree that the function is too big, in my opinion it should be refactored (and I plan to do that at some point). However, I think the

```
if b.cfg.userConfig.ShouldRegisterOwner() {
    err := b.cfg.RegisterOwner(b.cfg.ConfigFile, authInfo.UserInfo.Name)
   // error handling
```

already adds as little complexity as possible without losing clarity.

internal/broker/config.go
adombeck: Thanks for updating the comment, it makes more sense now :)

internal/broker/config.go
```diff
@@ -24,15 +27,55 @@ const (
...
+	OwnerUserKeyword = "OWNER"
+
+	// ownerAutoRegistrationConfigPath is the name of the file that will be auto-generated to register the owner.
+	ownerAutoRegistrationConfigPath     = "20-owner-autoregistration.conf"
+	ownerAutoRegistrationConfigTemplate = "templates/20-owner-autoregistration.conf.tmpl"
+)
+
+var (
+	//go:embed templates/20-owner-autoregistration.conf.tmpl
+	ownerRegistrationConfig embed.FS
```
BarcoMasile, line 51: nitpick: missed this one (which I didn't point to before) for renaming to `ownerAutoRegistretionConfig`, feel free to disregard

BarcoMasile (approved):
Great job!

conf/broker.conf
```diff
@@ -14,3 +14,31 @@ client_id = <CLIENT_ID>
...
+## via this broker will automatically be assigned the owner role. A
+## drop-in configuration file will be created in broker.conf.d/ to set
+## the 'owner' option.
+##
+## To disable automatic assignment, you can either:
+## 1. Explicitly set this option to an empty value (e.g. owner = "")
+## 2. Remove 'OWNER' from the 'allowed_users' option
+##
+## Example: owner = user2@example.com
+# owner =
```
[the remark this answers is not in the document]
adombeck: I like about the `##` that there is a clear distinction between the comment and the commented-out option. I'll push a commit which changes the comments of the old options to use that same style.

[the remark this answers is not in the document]
adombeck: I don't have permission to push to your branch, so I pushed it to [a separate branch](https://github.com/ubuntu/authd-oidc-brokers/commit/2de735d62079cca56858ced8e4701be9fc6ade2d), feel free to cherry-chick pick that commit.

cmd/authd-oidc/daemon/daemon.go
```diff
@@ -125,6 +125,12 @@ func (a *App) serve(config daemonConfig) error {
 		}
 	}
 
+	brokerConfigDir := config.Paths.BrokerConf + ".d"
```
3v1n0, line 128: Mybe we can avoid using constants here by refactoring something shared with https://github.com/ubuntu/authd-oidc-brokers/blob/4227a64cd96c11d954ea3f50828835e739263a5f/internal/broker/config.go#L35 ?

internal/broker/broker.go
```diff
@@ -620,6 +611,22 @@ func (b *Broker) handleIsAuthenticated(ctx context.Context, session *session, au
...
 
+	if !b.userNameIsAllowed(authInfo.UserInfo.Name) {
+		return AuthDenied, errorMessage{Message: "permission denied"}
+	}
+
+	// If the owner is unset and allowed, we auto-generate a config file with the first
+	// user to login as the owner.
+	if b.cfg.userConfig.shouldRegisterOwner() {
+		err := b.cfg.registerOwner(b.cfg.ConfigFile, authInfo.UserInfo.Name)
+		if err != nil {
```
3v1n0, line 622:
```suggestion
		if err := b.cfg.registerOwner(b.cfg.ConfigFile, authInfo.UserInfo.Name); err != nil {
```

internal/broker/broker.go
```diff
@@ -636,6 +643,27 @@ func (b *Broker) handleIsAuthenticated(ctx context.Context, session *session, au
...
+	if _, ok := b.cfg.userConfig.allowedUsers[userName]; ok {
+		return true
+	}
+	if b.cfg.userConfig.ownerAllowed {
+		// If owner is undefined, then the first user to login is considered the owner
+		if b.cfg.userConfig.firstUserBecomesOwner || b.cfg.userConfig.owner == userName {
+			return true
+		}
+	}
+	return false
```
3v1n0, line 664:
```suggestion

	if !b.cfg.userConfig.ownerAllowed {
		return false;
	}
        // If owner is undefined, then the first user to login is considered the owner
        return b.cfg.userConfig.firstUserBecomesOwner || b.cfg.userConfig.owner == userName
```

internal/broker/broker_test.go
```diff
@@ -729,6 +732,7 @@ func TestConcurrentIsAuthenticated(t *testing.T) {
 
 			b := newBrokerForTests(t, &brokerForTestConfig{
 				Config:          broker.Config{DataDir: dataDir},
+				allUsersAllowed: true,
```
3v1n0, line 735:
```suggestion
				anylUserAllowed: true,
```
maybe?

internal/broker/broker_test.go
```diff
@@ -827,6 +831,147 @@ func TestConcurrentIsAuthenticated(t *testing.T) {
 	}
 }
 
+func TestIsAuthenticatedAllowedUsersConfig(t *testing.T) {
+	t.Parallel()
+	tests := map[string]struct {
+		allowedUsers          map[string]bool
```
3v1n0, line 837: Since this is just a set, it's better not to use booleans here or it may imply something else:

```suggestion
		allowedUsers          map[string]struct{}
```

internal/broker/config.go
```diff
@@ -53,6 +96,45 @@ func getDropInFiles(cfgPath string) ([]any, error) {
...
+		cfg.ownerAllowed = true
+		cfg.firstUserBecomesOwner = true
+		return
+	}
+
+	cfg.homeBaseDir = users.Key(homeDirKey).String()
+	cfg.allowedSSHSuffixes = strings.Split(users.Key(sshSuffixesKey).String(), ",")
+
+	if cfg.allowedUsers == nil {
+		cfg.allowedUsers = make(map[string]bool)
```
3v1n0, line 111:
```suggestion
		cfg.allowedUsers = make(map[string]struct{})
```

internal/broker/config.go
```diff
@@ -53,6 +96,45 @@ func getDropInFiles(cfgPath string) ([]any, error) {
...
+			cfg.allUsersAllowed = true
+		} else if user == OwnerUserKeyword {
+			cfg.ownerAllowed = true
+			if !users.HasKey(ownerKey) {
+				// If owner is unset, then the first user becomes owner
+				cfg.firstUserBecomesOwner = true
+			}
+		} else {
+			cfg.allowedUsers[user] = true
+		}
```
3v1n0, line 130:
```suggestion
		if user == AllUsersKeyword {
			cfg.allUsersAllowed = true
			continue
		}
		
		if user == OwnerUserKeyword {
			cfg.ownerAllowed = true
			if !users.HasKey(ownerKey) {
				// If owner is unset, then the first user becomes owner
				cfg.firstUserBecomesOwner = true
			}
			continue
		}

		cfg.allowedUsers[user] = struct{}
```

internal/broker/config.go
```diff
@@ -24,15 +27,55 @@ const (
...
+type templateEnv struct {
+	Owner string
+}
+
+type userConfig struct {
+	clientID     string
+	clientSecret string
+	issuerURL    string
+
+	allowedUsers          map[string]bool
```
3v1n0, line 63:
```suggestion
	allowedUsers          map[string]struct{}
```

internal/broker/config.go
```diff
@@ -24,15 +27,55 @@ const (
...
+	allUsersAllowed       bool
```
3v1n0, line 64:
```suggestion
	anyUserAllowed       bool
```
?

internal/broker/config.go
```diff
@@ -24,15 +27,55 @@ const (
...
+	ownerAllowed          bool
+	firstUserBecomesOwner bool
+	owner                 string
+	homeBaseDir           string
+	allowedSSHSuffixes    []string
+}
+
+func getDropInDir(cfgPath string) string {
+	return cfgPath + ".d"
```
3v1n0, line 73: We may have this already, so let's share this logic in the same place for the 3 (now) occurrencies

internal/broker/broker_test.go
nsklikas: I would rather keep a consistent name across the code, unless you want me to change the name of the field in `userConfig` as well. BUT then we would be mapping the `ALL` keyword to `any`, which again feels weird.

3v1n0: Ok, fair enough

internal/broker/config.go
```diff
@@ -86,11 +168,50 @@ func parseConfigFile(cfgPath string) (userConfig, error) {
...
+	if err != nil {
+		slog.Error(fmt.Sprintf("Failed to open autoregistration template: %v", err))
+		return err
+	}
+
+	f, err := os.OpenFile(p, os.O_RDWR|os.O_CREATE|os.O_TRUNC, 0600)
+	if err != nil {
+		slog.Error(fmt.Sprintf("Failed to create owner registration file: %v", err))
+		return err
+	}
```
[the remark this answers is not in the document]
nsklikas: good catch

conf/broker.conf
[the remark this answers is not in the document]
nsklikas: thanks, will rebase

internal/broker/config.go
nsklikas: I will leave the function in this file and make exportable, unless you have any objections with that.

internal/broker/config.go
nsklikas: oops, changed it

internal/providers/msentraid/msentraid.go
```diff
@@ -273,12 +273,17 @@ func (p Provider) CurrentAuthenticationModesOffered(
 	return offeredModes, nil
 }
 
+// NormalizeUsername parses a username into a normalized version.
+func (p Provider) NormalizeUsername(username string) string {
+	return strings.ToLower(username)
```
3v1n0, line 278: Is this a new logic defined somewhere?

I think it's correct, but it wasn't used before, was it?

nsklikas:
I think that I have applied all of your comments. In addition, I introduced the `NormalizeUsername` method as described in the latest version of the spec.

internal/providers/msentraid/msentraid.go
nsklikas: It's defined in the latest version of the [spec](https://docs.google.com/document/d/1GK2yZ71_w2WxyNndap47RheezKGNCC4RyElKKeAmHL0/edit?tab=t.0#heading=h.azt725jx5rb8). Before `strings.EqualFold` was used which is case insensitive.

The reason for adding this method is that we wish to check if the normalized username is in the `allowed_users` list, so we had to extract the normalization logic in a separate function.

adombeck:
> > I introduced the NormalizeUsername method as described in the latest version of the spec.
> 
> This should also close [ubuntu/authd#530](https://github.com/ubuntu/authd/issues/530) and [ubuntu/authd#531](https://github.com/ubuntu/authd/issues/531), no?

No, I don't think those bugs are affected by the change, because `NormalizeUsername` is only used to compare usernames in the broker, and we already used the case-insensitive `VerifyUsername` for that before.

internal/broker/broker.go
```diff
@@ -620,6 +613,21 @@ func (b *Broker) handleIsAuthenticated(ctx context.Context, session *session, au
...
 	}
 
+	if !b.userNameIsAllowed(authInfo.UserInfo.Name) {
+		return AuthDenied, errorMessage{Message: "permission denied"}
+	}
+
+	// If the owner is unset and allowed, we auto-generate a config file with the first
+	// user to login as the owner.
+	if b.cfg.userConfig.shouldRegisterOwner() {
+		if err := b.cfg.registerOwner(b.cfg.ConfigFile, authInfo.UserInfo.Name); err != nil {
```
wood-push-melon, line 623: Just a minor question (feel free to ignore if I am wrong):

Is the check `b.cfg.userConfig.shouldRefisterOwner()` a bit duplicated with the check in the `b.cfg.registerOwner`? Is it okay to remove the check logic in the `registerOwner` method?

nsklikas: It is, wanted to be extra safe that the function won't be called when it shouldn't. As it seems redundant now, I will remove it.

internal/broker/broker.go
```diff
@@ -636,6 +644,25 @@ func (b *Broker) handleIsAuthenticated(ctx context.Context, session *session, au
...
+		return true
+	}
+	if _, ok := b.cfg.userConfig.allowedUsers[userName]; ok {
+		return true
+	}
+	if !b.cfg.userConfig.ownerAllowed {
+		return false
+	}
+	// If owner is undefined, then the first user to login is considered the owner
+	return b.cfg.userConfig.firstUserBecomesOwner || b.cfg.userConfig.owner == userName
```
adombeck, line 663: There's a race condition when two users log in at the same time. I tried to fix that in 9cc259e0d8bc26e125009b4d934292cbdcbdbc55, feel free to cherry-pick.

internal/broker/broker_test.go
```diff
@@ -827,6 +831,147 @@ func TestConcurrentIsAuthenticated(t *testing.T) {
...
+			owner:              "machine_owner",
+			allowedUsernames:   []string{"machine_owner"},
+			unallowedUsernames: []string{"u1", "u2"},
+		},
+		"No_users_allowed": {
+			ownerAllowed:       true,
+			owner:              "",
+			unallowedUsernames: []string{"u1", "u2", "machine_owner", "u3"},
+		},
+		"Only_unset_owner_allowed": {
```
adombeck, line 861: Maybe a bit easier to understand

```suggestion
		"Only_first_user_allowed": {
```

internal/broker/broker_test.go
```diff
@@ -827,6 +831,147 @@ func TestConcurrentIsAuthenticated(t *testing.T) {
...
+		},
+		"Only_owner_allowed": {
+			ownerAllowed:       true,
+			owner:              "machine_owner",
+			allowedUsernames:   []string{"machine_owner"},
+			unallowedUsernames: []string{"u1", "u2"},
+		},
+		"No_users_allowed": {
+			ownerAllowed:       true,
+			owner:              "",
```
adombeck, line 858:
```suggestion

```

adombeck (commented):
I have not yet fully completed this iteration of the review, submitting anyway.

I feel like we're getting close to merge!

internal/broker/broker_test.go
nsklikas: If I remove this it will be like testing the following:
```
owner = 
allowed_users =
```

But I am trying to test:
```
owner = 
allowed_users = OWNER
```
which is equivalent to:
```
owner = 
```

Should I add an extra test to test what you suggest?

internal/providers/msentraid/msentraid.go
```diff
@@ -273,12 +273,17 @@ func (p Provider) CurrentAuthenticationModesOffered(
...
 }
 
-// VerifyUsername checks if the authenticated username matches the requested username and that both are valid.
-func (p Provider) VerifyUsername(requestedUsername, authenticatedUsername string) error {
-	// Microsoft Entra usernames are case-insensitive. We can safely use strings.EqualFold here without worrying about
+// NormalizeUsername parses a username into a normalized version.
+func (p Provider) NormalizeUsername(username string) string {
+	// Microsoft Entra usernames are case-insensitive. We can safely use strings.ToLower here without worrying about
 	// different Unicode characters that fold to the same lowercase letter, because the Microsoft Entra username policy
 	// (which we checked above) ensures that the username only contains ASCII characters.
```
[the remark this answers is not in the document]
nsklikas: Just to be clear, this refers to the check in https://github.com/ubuntu/authd-oidc-brokers/blob/main/internal/providers/msentraid/msentraid.go#L288, right?

internal/broker/broker_test.go
```diff
@@ -827,6 +831,147 @@ func TestConcurrentIsAuthenticated(t *testing.T) {
...
+					if err != nil {
+						require.ErrorIs(t, err, os.ErrNotExist, "Teardown: Failed to rename issuer data directory")
+						t.Logf("Failed to rename issuer data directory: %v", err)
+					}
+				}
+			}
+
+			golden.CheckOrUpdateFileTree(t, outDir)
+		})
+	}
```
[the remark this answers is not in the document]
nsklikas: Great, this will really simplify the test

internal/providers/msentraid/msentraid.go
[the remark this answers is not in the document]
adombeck: yes, exactly

internal/broker/broker_test.go
adombeck: > But I am trying to test:
> ```
> owner = 
> allowed_users = OWNER
> ```

Isn't that already tested by https://github.com/ubuntu/authd-oidc-brokers/blob/9cc259e0d8bc26e125009b4d934292cbdcbdbc55/internal/broker/broker_test.go#L861-L866

adombeck: Ah no it's not, because `firstUserBecomesOwner` is `false` when you explicitly set owner to an empty value.

adombeck: How about we use these two test cases:

```
"No_users_allowed": {
	unallowedUsernames: []string{"u1", "u2", "machine_owner", "u3"},
},
"No_users_allowed_when_owner_is_allowed_but_not_set": {
	ownerAllowed:       true,
	unallowedUsernames: []string{"u1", "u2", "machine_owner", "u3"},
},
```

internal/broker/helper_test.go
```diff
@@ -49,6 +54,23 @@ func newBrokerForTests(t *testing.T, cfg *brokerForTestConfig) (b *broker.Broker
 	if cfg.allowedSSHSuffixes != nil {
 		cfg.SetAllowedSSHSuffixes(cfg.allowedSSHSuffixes)
 	}
+	if cfg.allowedUsers == nil {
+		cfg.SetAllowedUsers(map[string]struct{}{"OWNER": {}})
+	} else {
+		cfg.SetAllowedUsers(cfg.allowedUsers)
+	}
```
adombeck, line 61: We don't add "OWNER" to the `userConfig.allowedUsers` map anymore

```suggestion
	if cfg.allowedUsers != nil {
		cfg.SetAllowedUsers(cfg.allowedUsers)
	}
```

internal/broker/broker_test.go
```diff
@@ -827,6 +831,147 @@ func TestConcurrentIsAuthenticated(t *testing.T) {
...
+			allowedUsers:       map[string]struct{}{"u1": {}, "u2": {}},
+			ownerAllowed:       true,
+			owner:              "machine_owner",
+			allowedUsernames:   []string{"u1", "u2", "machine_owner"},
+			unallowedUsernames: []string{"u3"},
+		},
+		"Owner_is_disabled_even_when_registered": {
+			allowedUsers:       map[string]struct{}{"u1": {}, "u2": {}},
+			owner:              "machine_owner",
+			allowedUsernames:   []string{"u1", "u2"},
```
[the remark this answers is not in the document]
adombeck: Mmh maybe we want to differentiate between `allowedUsers` which we pass to the broker during setup and the list of users which we expect the broker to allow (when checking the authentication result).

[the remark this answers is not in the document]
adombeck: How about this: e37011c81b1999358cc2f4a2437a552649e28a36

internal/broker/broker_test.go
```diff
@@ -827,6 +831,147 @@ func TestConcurrentIsAuthenticated(t *testing.T) {
...
+			unallowedUsernames: []string{"u3"},
+		},
+		"Specific_users_and_owner": {
+			allowedUsers:       map[string]struct{}{"u1": {}, "u2": {}},
+			ownerAllowed:       true,
+			owner:              "machine_owner",
+			allowedUsernames:   []string{"u1", "u2", "machine_owner"},
+			unallowedUsernames: []string{"u3"},
+		},
+		"Owner_is_disabled_even_when_registered": {
```
adombeck, line 879: Maybe a bit easier to understand

```suggestion
		"Owner_is_set_but_not_allowed": {
```

adombeck: We can also name it consistently with the other `No_users_allowed` test cases and group them together: 66996aa01d9f7b6669766b16e374908fa80e9f91

internal/broker/config.go
```diff
@@ -86,11 +176,48 @@ func parseConfigFile(cfgPath string) (userConfig, error) {
...
+	defer f.Close()
+
+	err = t.Execute(f, templateEnv{Owner: userName})
+	if err != nil {
```
3v1n0, line 212:
```suggestion
	if err := t.Execute(f, templateEnv{Owner: userName}); err != nil {
```

internal/broker/config.go
```diff
@@ -86,11 +176,48 @@ func parseConfigFile(cfgPath string) (userConfig, error) {
...
+		return nil
+	}
+
+	p := filepath.Join(GetDropInDir(cfgPath), ownerAutoRegistrationConfigPath)
+
+	templateName := filepath.Base(ownerAutoRegistrationConfigTemplate)
+	t, err := template.New(templateName).ParseFS(ownerAutoRegistrationConfig, ownerAutoRegistrationConfigTemplate)
+	if err != nil {
+		slog.Error(fmt.Sprintf("Failed to open autoregistration template: %v", err))
+		return err
```
3v1n0, line 201: I feel it would be better to just:

```suggestion
		return fmt.Errorf("failed to open autoregistration template: %w", err)
```

Then in the caller function to log in case of error.

internal/broker/config_test.go
```diff
@@ -150,3 +152,216 @@ func TestParseConfig(t *testing.T) {
...
+
+func TestRegisterOwner(t *testing.T) {
+	outDir := t.TempDir()
+	userName := "owner_name"
+	confPath := filepath.Join(outDir, "broker.conf")
+
+	err := os.WriteFile(confPath, []byte(configTypes["valid"]), 0600)
+	require.NoError(t, err, "Setup: Failed to write config file")
+
+	dropInDir := confPath + ".d"
```
3v1n0, line 351: Can reuse `GetDropInDir`?

internal/broker/config_test.go
```diff
@@ -150,3 +152,216 @@ func TestParseConfig(t *testing.T) {
...
+		t.Run(name, func(t *testing.T) {
+			t.Parallel()
+
+			outDir := t.TempDir()
+			confPath := filepath.Join(outDir, "broker.conf")
+
+			err := os.WriteFile(confPath, []byte(testParseUserConfigTypes[name]), 0600)
+			require.NoError(t, err, "Setup: Failed to write config file")
+
+			dropInDir := confPath + ".d"
```
3v1n0, line 319: Same here, can reuse `GetDropInDir`?

internal/broker/helper_test.go
```diff
@@ -49,6 +54,23 @@ func newBrokerForTests(t *testing.T, cfg *brokerForTestConfig) (b *broker.Broker
 	if cfg.allowedSSHSuffixes != nil {
 		cfg.SetAllowedSSHSuffixes(cfg.allowedSSHSuffixes)
 	}
+	if cfg.allowedUsers == nil {
+		cfg.SetAllowedUsers(map[string]struct{}{"OWNER": {}})
+	} else {
+		cfg.SetAllowedUsers(cfg.allowedUsers)
```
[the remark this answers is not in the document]
3v1n0: Can likely ignore as per https://github.com/ubuntu/authd-oidc-brokers/pull/249/files#r1886878243

internal/providers/msentraid/msentraid_test.go
```diff
@@ -52,6 +52,35 @@ func TestCheckTokenScopes(t *testing.T) {
...
+	}
+	for name, tc := range tests {
+		t.Run(name, func(t *testing.T) {
+			t.Parallel()
+
+			p := msentraid.New()
+
+			ret := p.NormalizeUsername(tc.username)
+
+			require.Equal(t, tc.normalizedUsername, ret)
```
3v1n0, line 79: nit:

```suggestion
			p := msentraid.New()
			normalized := p.NormalizeUsername(tc.username)
			require.Equal(t, tc.normalizedUsername, normalized)
```

internal/providers/msentraid/msentraid_test.go
```diff
@@ -52,6 +52,35 @@ func TestCheckTokenScopes(t *testing.T) {
 	}
 }
 
+func TestNormalizeUsername(t *testing.T) {
+	t.Parallel()
+	tests := map[string]struct {
+		username string
+
+		normalizedUsername string
```
3v1n0, line 60:
```suggestion
		wantNormalized string
```

internal/broker/config.go
adombeck: agreed

internal/broker/config_test.go
nsklikas: So I just want to offer my logic behind this, in tests IMO in tests it's better not to use helper functions or constants (e.g. a couple of lines we use `"20-owner-autoregistration.conf"` to generate the path of the file, instead of using the const `ownerAutoRegistrationConfigPath`). The reason for this is to act as a safe guard for typos in the code and to make changing this behavior harder. Also this function is used in the function that we are trying to test, so from a higher level I don't like this.

Of course, this does not always make sense and there are cases where it makes changes to the code harder (but I think that behavior like this is unlikely to change, so this is not the case).

Anyway, this is a nit-pick I would say, I don't think there is any real difference between the 2 approaches and I understand the reason, so I will update it.

internal/broker/broker.go
shipperizer: @adombeck are you able to provide a test for it?

at the same time being conscious of the previous use case where all users are allowed, this is a new functionality that can simply go in a separate PR and we can write a `TODO` that is going to get addressed

internal/providers/msentraid/msentraid_test.go
nsklikas: The newlines were there to separate the 3 steps of the test (setup, execute, assert). I will update it.

internal/broker/broker.go
adombeck: In my opinion we don't necessarily need a test for it, and because it seems to require quite some effort to add one, I'd say lets merge it without the test.

nsklikas:
I tried to apply all the changes requested:
- Cherry picked @adombeck commits (had to change them slightly), thanks for the race condition fix. As discussed on MM, no tests were added.
- Removed golden files from the broker tests
- Fixed a bug where the owner was not normalized on auto-registration
- Applied code style changes

adombeck (approved):
To me this looks good to merge now, great job!

adombeck:
@3v1n0 are you happy as well?

internal/broker/broker.go
```diff
@@ -636,6 +643,35 @@ func (b *Broker) handleIsAuthenticated(ctx context.Context, session *session, au
 	return AuthGranted, userInfoMessage{UserInfo: authInfo.UserInfo}
 }
 
+func (b *Broker) maybeRegisterOwner(cfgPath string, userName string) error {
+	// We need to lock here to avoid a race condition where two users log in at the same time, causing both to be
+	// considered the owner.
+	b.setOwnerMutex.Lock()
+	defer b.setOwnerMutex.Unlock()
```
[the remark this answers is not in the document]
adombeck: what do you mean?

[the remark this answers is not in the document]
3v1n0: I mean the mutex being part of `userConfig` and so moving the ops dealing with it there?

internal/broker/broker.go
```diff
@@ -657,11 +669,7 @@ func (b *Broker) userNameIsAllowed(userName string) bool {
 	if _, ok := b.cfg.userConfig.allowedUsers[normalizedUsername]; ok {
 		return true
 	}
-	if !b.cfg.userConfig.ownerAllowed {
-		return false
-	}
-	// If owner is undefined, then the first user to log in is considered the owner
-	return b.cfg.userConfig.firstUserBecomesOwner || b.cfg.userConfig.owner == normalizedUsername
+	return b.cfg.userConfig.ownerAllowed && b.cfg.userConfig.owner == normalizedUsername
```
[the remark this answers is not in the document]
nsklikas: No, this part of the code will be reached only after the first user logs in (owner is registered) so we don't need to lock it. 

`owner` will be written only once and that is protected by the mutex. All other accesses are reads and they will happen after the owner is written.

internal/broker/broker.go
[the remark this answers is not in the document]
nsklikas: I considered that, but I decided to keep @adombeck's suggestion because i wanted to keep the config object as light as possible.

On one hand it makes sense to keep the mutex as close as possible to the race condition (the writing of the file), so we could move it to the config object. On the other hand, it also fine to say that `registerOwner` is not thread safe and the caller is responsible for implementing that. 

I am fine with either approach (slightly leaning on keeping the current one).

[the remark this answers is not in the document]
adombeck: Ah okay. I don't care if we keep it here or move it to `userConfig`.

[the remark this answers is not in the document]
3v1n0: I meant something more like:

```diff
diff --git a/internal/broker/broker.go b/internal/broker/broker.go
index 4cf8198..6426843 100644
--- a/internal/broker/broker.go
+++ b/internal/broker/broker.go
@@ -49,8 +49,6 @@ type Config struct {
 type Broker struct {
 	cfg Config
 
-	setOwnerMutex sync.Mutex
-
 	provider providers.Provider
 	oidcCfg  oidc.Config
 
@@ -615,7 +613,7 @@ func (b *Broker) handleIsAuthenticated(ctx context.Context, session *session, au
 		}
 	}
 
-	err = b.maybeRegisterOwner(b.cfg.ConfigFile, authInfo.UserInfo.Name)
+	b.cfg.registerOwner(b.cfg.ConfigFile, authInfo.UserInfo.Name)
 	if err != nil {
 		// The user is not allowed if we fail to create the owner-autoregistration file.
 		// Otherwise the owner might change if the broker is restarted.
@@ -643,19 +641,6 @@ func (b *Broker) handleIsAuthenticated(ctx context.Context, session *session, au
 	return AuthGranted, userInfoMessage{UserInfo: authInfo.UserInfo}
 }
 
-func (b *Broker) maybeRegisterOwner(cfgPath string, userName string) error {
-	// We need to lock here to avoid a race condition where two users log in at the same time, causing both to be
-	// considered the owner.
-	b.setOwnerMutex.Lock()
-	defer b.setOwnerMutex.Unlock()
-
-	if !b.cfg.userConfig.shouldRegisterOwner() {
-		return nil
-	}
-
-	return b.cfg.registerOwner(cfgPath, userName)
-}
-
 // userNameIsAllowed checks whether the user's username is allowed to access the machine.
 func (b *Broker) userNameIsAllowed(userName string) bool {
 	normalizedUsername := b.provider.NormalizeUsername(userName)
@@ -669,7 +654,8 @@ func (b *Broker) userNameIsAllowed(userName string) bool {
 	if _, ok := b.cfg.userConfig.allowedUsers[normalizedUsername]; ok {
 		return true
 	}
-	return b.cfg.userConfig.ownerAllowed && b.cfg.userConfig.owner == normalizedUsername
+
+	return b.cfg.isOwnerAllowed(normalizedUsername)
 }
 
 func (b *Broker) startAuthenticate(sessionID string) (context.Context, error) {
diff --git a/internal/broker/config.go b/internal/broker/config.go
index 7070921..c084069 100644
--- a/internal/broker/config.go
+++ b/internal/broker/config.go
@@ -69,11 +69,10 @@ type userConfig struct {
 	ownerAllowed          bool
 	firstUserBecomesOwner bool
 	owner                 string
+	ownerMutex            *sync.RWMutex
 	homeBaseDir           string
 	allowedSSHSuffixes    []string
 
-	setOwnerMutex *sync.Mutex
-
 	provider provider
 }
 
@@ -144,12 +143,14 @@ func populateUsersConfig(cfg *userConfig, users *ini.Section) {
 
 	// We need to read the owner key after we call HasKey, because the key is created
 	// when we call the "Key" function and we can't distinguish between empty and unset.
+	cfg.ownerMutex.Lock()
+	defer cfg.ownerMutex.Unlock()
 	cfg.owner = cfg.provider.NormalizeUsername(users.Key(ownerKey).String())
 }
 
 // parseConfigFile parses the config file and returns a map with the configuration keys and values.
 func parseConfigFile(cfgPath string, p provider) (userConfig, error) {
-	cfg := userConfig{provider: p, setOwnerMutex: &sync.Mutex{}}
+	cfg := userConfig{provider: p, ownerMutex: &sync.RWMutex{}}
 
 	dropInFiles, err := getDropInFiles(cfgPath)
 	if err != nil {
@@ -189,11 +190,22 @@ func (uc *userConfig) shouldRegisterOwner() bool {
 	return uc.ownerAllowed && uc.firstUserBecomesOwner && uc.owner == ""
 }
 
+func (uc *userConfig) isOwnerAllowed(userName string) bool {
+	uc.ownerMutex.RLock()
+	defer uc.ownerMutex.RUnlock()
+
+	return uc.ownerAllowed && uc.owner == userName
+}
+
 func (uc *userConfig) registerOwner(cfgPath, userName string) error {
 	// We need to lock here to avoid a race condition where two users log in at the same time, causing both to be
 	// considered the owner.
-	uc.setOwnerMutex.Lock()
-	defer uc.setOwnerMutex.Unlock()
+	uc.ownerMutex.Lock()
+	defer uc.ownerMutex.Unlock()
+
+	if !uc.shouldRegisterOwner() {
+		return nil
+	}
 
 	if cfgPath == "" {
 		uc.owner = uc.provider.NormalizeUsername(userName)
diff --git a/internal/broker/config_test.go b/internal/broker/config_test.go
index 90f33b1..f20e9d5 100644
--- a/internal/broker/config_test.go
+++ b/internal/broker/config_test.go
@@ -59,7 +59,7 @@ issuer = https://higher-precedence-issuer.url.com
 func TestParseConfig(t *testing.T) {
 	t.Parallel()
 	p := &testutils.MockProvider{}
-	ignoredFields := map[string]struct{}{"provider": {}, "setOwnerMutex": {}}
+	ignoredFields := map[string]struct{}{"provider": {}, "ownerMutex": {}}
 
 	tests := map[string]struct {
 		configType string
@@ -319,7 +319,7 @@ func TestRegisterOwner(t *testing.T) {
 	err = os.Mkdir(dropInDir, 0700)
 	require.NoError(t, err, "Setup: Failed to create drop-in directory")
 
-	cfg := userConfig{firstUserBecomesOwner: true, ownerAllowed: true, provider: p, setOwnerMutex: &sync.Mutex{}}
+	cfg := userConfig{firstUserBecomesOwner: true, ownerAllowed: true, provider: p, ownerMutex: &sync.RWMutex{}}
 	err = cfg.registerOwner(confPath, userName)
 	require.NoError(t, err)
 
diff --git a/internal/broker/export_test.go b/internal/broker/export_test.go
index 1723c93..769c093 100644
--- a/internal/broker/export_test.go
+++ b/internal/broker/export_test.go
@@ -9,7 +9,7 @@ import (
 )
 
 func (cfg *Config) Init() {
-	cfg.setOwnerMutex = &sync.Mutex{}
+	cfg.ownerMutex = &sync.RWMutex{}
 }
 
 func (cfg *Config) SetClientID(clientID string) {
@@ -29,10 +29,14 @@ func (cfg *Config) SetAllowedUsers(allowedUsers map[string]struct{}) {
 }
 
 func (cfg *Config) SetOwner(owner string) {
+	cfg.ownerMutex.Lock()
+	defer cfg.ownerMutex.Unlock()
 	cfg.owner = owner
 }
 
 func (cfg *Config) SetFirstUserBecomesOwner(firstUserBecomesOwner bool) {
+	cfg.ownerMutex.Lock()
+	defer cfg.ownerMutex.Unlock()
 	cfg.firstUserBecomesOwner = firstUserBecomesOwner
 }
 
```

Now the usage of the RWMutex vs Mutex is on whether you want to b able to read an unset value while writing on it.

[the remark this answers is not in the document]
nsklikas: Applied the patch, should be fine now (cc @3v1n0)

internal/broker/broker.go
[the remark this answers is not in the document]
3v1n0: Mh, I see that in this specific case what happens, but in general if we've a critical section then we should ensure that's always the case. At least as clean-code policy.

So, I'd say in this specific case having it write-protected by always readable by a RWMutex is the way to make sure that both `owner` and `ownerAllowed` are always read and written in a critical section.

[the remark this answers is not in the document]
nsklikas: I see your point, I'm fine with that. Anyway, the patch you provided should already take care of that.

PR #250, opened 2024-12-11
Fix spelling in config file

adombeck:
* "login" is the noun, "log in" is the verb
* Write "SSH" instead of "ssh"


adombeck:
Superseded by https://github.com/ubuntu/authd-oidc-brokers/pull/253

PR #251, opened 2024-12-12
Remove incorrect os.Exit(m.Run()) calls

didrocks:
This is not needed since Go 1.15, where the test runtime registers the test result itself.
Also, os.Exit() would skip any defer() being called before, which might pollute the system running the tests.

UDENG-5633


didrocks:
Upstream discussion and reference: https://github.com/golang/go/issues/34129

3v1n0 (approved):
Indeed!!

PR #257, opened 2024-12-13
Device ownership migration

nsklikas:
https://github.com/ubuntu/authd-oidc-brokers/pull/249 introduces a breaking change. The default behavior before was that ALL users were allowed, but now only the owner is allowed.

In order to migrate the config, we set a new snapctl config param. If that is not present, that means that we need to migrate the config.

Performed the following manual test:
1. Packed  and installed the snap from `main`
2. Tagged the latest commit on this branch with `0.2.0`
3. Packed and installed the snap from this branch
4. Validated that  `00-migration-allowed_users.conf` and `10-allowed_users.conf` are created
5. `rm -r  /var/snap/authd-oidc/current/broker.conf.d/`
6. Installed the snap from this branch again
7. Validated that `/var/snap/authd-oidc/current/broker.conf.d/` is empty

The logs can be read with `journalctl -r -t authd-oidc`


nsklikas:
@adombeck @3v1n0 please have a look when you can

snap/hooks/post-refresh
```diff
@@ -0,0 +1,34 @@
+#!/bin/sh
+set -eu
+
+echo "post-refresh hook" >> /tmp/snap.log # TODO: Only for debugging
+
+PREVIOUS_VERSION=$(snapctl get previous-version)
+echo "Previous version: $PREVIOUS_VERSION" >> /tmp/snap.log
```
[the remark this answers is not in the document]
nsklikas: Good idea, changed the file name to `snap-post-refresh` and changed it so that some of the logs are written to the journal (`journalctl -r -t authd-oidc` to get them)

snap/hooks/post-refresh
```diff
@@ -0,0 +1,34 @@
...
+
+should_transition_to_allowed_users() {
+  # Transition to allowed users if:
+  # - previous-version is not set (that means that the previous version is
+  #   older than 0.2.0, i.e. the version where we introduced setting the
+  #   previous-version in the pre-refresh hook).
+  # - previous-version is set, but it is less than 0.2.0. That should never
+  #   happen, but we check it to give an example how the previous-version
+  #   can be used to transition data from older versions.
+  [ -z "${PREVIOUS_VERSION:-}" ] || version_less_than "${PREVIOUS_VERSION:-}" "${INITIAL_ALLOWED_USERS_VERSION:-}"
```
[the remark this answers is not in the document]
nsklikas: tried it without `:-` and it works, thanks

snap/hooks/post-refresh
```diff
@@ -0,0 +1,34 @@
...
+}
+
+transition_to_allowed_users() {
+  echo "Transitioning to allowed users" >> /tmp/snap.log # TODO: Only for debugging
+  mkdir ${SNAP_DATA}/broker.conf.d
+  cp --update=none ${SNAP}/conf/broker.conf.d.orig/{00-migration-allowed_users.conf,10-allowed_users.conf} ${SNAP_DATA}/broker.conf.d/
```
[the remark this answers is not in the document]
nsklikas: That makes sense, updated it

snap/hooks/post-refresh
```diff
@@ -0,0 +1,43 @@
...
+
+PREVIOUS_VERSION=$(snapctl get previous-version)
+INITIAL_ALLOWED_USERS_VERSION="0.2.0"
+
+version_less_than() {
+  [ "$(printf '%s\n' "${@}" | sort -V | head -n1)" = "$1" ]
+}
+
+echo_and_log_to_journal() {
+  echo "${@}"
```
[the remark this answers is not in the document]
nsklikas: This is piped to `/tmp/snap-private-tmp/snap.authd-oidc/tmp/snap-refresh.log`.

snap/hooks/post-refresh
```diff
@@ -0,0 +1,43 @@
...
+  logger -t $SNAP_NAME "${@}"
```
adombeck, line 13: did you test that the messages actually end up in the journal (i.e. that the snap sandbox lets us print to the host's journal)?

snap/hooks/post-refresh
```diff
@@ -0,0 +1,43 @@
...
+  #   previous-version in the pre-refresh hook).
+  # - previous-version is set, but it is less than 0.2.0. That should never
+  #   happen, but we check it to give an example how the previous-version
+  #   can be used to transition data from older versions.
+  [ -z "${PREVIOUS_VERSION}" ] || version_less_than "${PREVIOUS_VERSION}" "${INITIAL_ALLOWED_USERS_VERSION}"
+}
+
+transition_to_allowed_users() {
+  echo_and_log_to_journal "Transitioning to allowed users"
+  mkdir -p ${SNAP_DATA}/broker.conf.d
```
adombeck, line 29:
```suggestion
  mkdir -p --mode=700 ${SNAP_DATA}/broker.conf.d
```

snap/hooks/post-refresh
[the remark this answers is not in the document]
adombeck: no that's not what I meant, I didn't see that you're redirecting stdout, so I was wondering if there is any use in printing to stdout.

snap/hooks/post-refresh
```diff
@@ -0,0 +1,43 @@
...
+  cp --update=none ${SNAP}/conf/migrations/pre-${INITIAL_ALLOWED_USERS_VERSION}/broker.conf.d/* ${SNAP_DATA}/broker.conf.d/
```
3v1n0, line 30: Nice, as per this I think you can just make the logic easier by just copying everything that is under  `${SNAP}/conf/migrations/pre-$....` where the version is something lower than ours no?

I mean, without having  specific `should_transition_to_allowed_users` case I think.

snap/hooks/post-refresh
nsklikas: yes, do you get a different behavior?

adombeck: I did not test it

snap/hooks/post-refresh
adombeck: I don't see how you can get rid of the checks done in `should_transition_to_allowed_users`. can you provide the command(s) you have in mind to do that?

3v1n0: Yeah I was thinking something like that, but let me know if that's not as simple as I thought :)

nsklikas: The problem I see with this is that if you refresh with the same version (or even from 0.2 to 0.3) these files would be created.

We don't want this to happen, because it would changing the default behavior (it would be allow `ALL` instead of allow `OWNER`). We only want to to create those files if we are migrating from `<0.2`, because the default behavior has changed and we don't want to introduce breaking changes.

3v1n0: I see, fair enough...

snap/hooks/post-refresh
nsklikas: Ok, I thought I was seeing ghosts :P

adombeck:
I found an edge case in which the migration doesn't work:

1. Install a version >0.2.0
2. Roll back to a version <0.2.0
3. Update to a version >0.2.0

In step 3, the `previous-version` is still set to the value from step 1, so it doesn't migrate.

I don't see a way to fix that and in my opinion it's acceptable to not handle this edge case.

snap/hooks/post-refresh
adombeck: I noticed that the `broker.conf` is created with the default permissions in https://github.com/nsklikas/authd-oidc-brokers/blob/4e490fdbeabb7c6b2331493775bbe924ccdbacc1/snap/hooks/install#L4. IMO we should be consistent, so either we create both with default permissions or with 700. I'm leaning towards the latter, because the config file can contains secrets (e.g. the `client_secret`).

adombeck: See https://github.com/ubuntu/authd-oidc-brokers/pull/262

nsklikas:
> I found an edge case in which the migration doesn't work:
> 
>     1. Install a version >= 0.2.0
> 
>     2. Roll back to a version < 0.2.0
> 
>     3. Update to a version >= 0.2.0
> 
> 
> In step 3, the `previous-version` is still set to the value from step 1, so it doesn't migrate.
> 
> I don't see a way to fix that and in my opinion it's acceptable to not handle this edge case.

Oh, nice catch. This is an edge case, so I think it could be fine if we don't cater to it. But could we avoid this by unsetting the `previous-version` when the post-refresh hook finishes? I guess even in that case there could be cases were the config would remain set (e.g. in case of an error), but I think that it is good enough.

 I will try to push a fix tomorrow.

adombeck:
> But could we avoid this by unsetting the previous-version when the post-refresh hook finishes? 

Good idea, that should work!

snap/hooks/post-refresh
```diff
@@ -0,0 +1,39 @@
...
+  #   previous-version in the pre-refresh hook).
+  # - previous-version is set, but it is less than 0.2.0. That should never
+  #   happen, but we check it to give an example how the previous-version
+  #   can be used to transition data from older versions.
+  [ -z "${PREVIOUS_VERSION}" ] || version_less_than "${PREVIOUS_VERSION}" "${INITIAL_ALLOWED_USERS_VERSION}"
+}
+
+transition_to_allowed_users() {
+  log "Transitioning to allowed users"
+  install -D --target-directory --mode=0700"${SNAP_DATA}/broker.conf.d" "${SNAP}/conf/migrations/pre-${INITIAL_ALLOWED_USERS_VERSION}/broker.conf.d/"*
```
3v1n0, line 28: nit (and a space fix):

```suggestion
  install -D --target-directory --mode=0700 "${SNAP_DATA}/broker.conf.d" \
    "${SNAP}/conf/migrations/pre-${INITIAL_ALLOWED_USERS_VERSION}/broker.conf.d/"*
```

snap/hooks/pre-refresh
```diff
@@ -0,0 +1,20 @@
+#!/bin/sh
+set -eu
+
+INITIAL_PREVIOUS_VERSION_CONFIG="0.2.0"
+
+if [ -n "${DEBUG:-}" ]; then
```
nsklikas, line 6: @adombeck question: Is `DEBUG` something you can enable in snap or is it for local testing?

adombeck: You can set the environment variable in the `snapcraft.yaml`  before building the snap:

```
environment:
  DEBUG: "true"
```

snap/hooks/pre-refresh
```diff
@@ -1,8 +1,20 @@
...
+version_less_than() {
+  [ "$(printf '%s\n' "${@}" | sort -V | head -n1)" = "$1" ]
+}
+
+# If we are downgrading, we don't set the `previous-version` if the target version is older
+# than the version that introduced this configuration (0.2.0).
+# Otherwise, we will end up with stale data.
+if version_less_than "${SNAP_VERSION}" "${INITIAL_PREVIOUS_VERSION_CONFIG}"; then
+  snapctl set previous-version="${SNAP_VERSION}"
+fi
```
adombeck, line 20: I don't think that's going to work. We only ship this `pre-refresh` hook in 2.0.0, so if it exists, `SNAP_VERSION` will always be >= 2.0.0. I think unsetting the `previous-version` in the `post-refresh` hook, as you suggested yesterday, should work though.

nsklikas: Yep you are right. The problem with what I suggested yesterday is that the pre-refresh hook will run when the downgrade happens and it will set the config. Marking the PR as draft for now, until I test things out.

nsklikas:
I couldn't find a workaround for the case that @adombeck found, but I think it's fine if we leave it as it is. If the user upgraded to 0.2.0, then migrated to 0.1.0, deleted the migration files and then upgraded again, I think we can assume that they know what they are doing. If they don't manually delete the migration files, then we are fine because they will still be there after upgrading.

I pushed a small unrelated fix, if we all agree that we can't/won't solve this bug then I think we can move on.

snap/hooks/pre-refresh
adombeck: >  The problem with what I suggested yesterday is that the pre-refresh hook will run when the downgrade happens and it will set the config.

Ah, right! Too bad.

snap/hooks/post-refresh
```diff
@@ -17,7 +17,7 @@ log() {
 }
 
 version_less_than() {
-  [ "$(printf '%s\n' "${@}" | sort -V | head -n1)" = "$1" ]
+  [ "$1" = "$2" ] && return 1 || [ "$(printf '%s\n' "${@}" | sort -V | head -n1)" = "$1" ]
```
adombeck, line 20: Good catch!

adombeck:
> if we all agree that we can't/won't solve this bug then I think we can move on.

I also don't see a way to solve that edge case with the limited options snap currently provides. I think it's fine to not handle that case.

adombeck (approved):

3v1n0 (approved):
Nice one, thanks!

PR #264, opened 2024-12-19
daemon: Always close the ready channel on error

3v1n0:
We were not closing it in case broker.New failed, so let's just use a deferred function to ensure that this always happens, unless we're done it already explicitly

### DIFF
--- a/cmd/authd-oidc/daemon/daemon.go
+++ b/cmd/authd-oidc/daemon/daemon.go
@@ -116,18 +116,26 @@ func New(name string) *App {
 // serve creates new dbus service on the system bus. This call is blocking until we quit it.
 func (a *App) serve(config daemonConfig) error {
 	ctx := context.Background()
+	// Ensure that the a.ready channel is closed when the function returns, which is what Quit() waits for before exiting.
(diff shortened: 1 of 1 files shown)


cmd/authd-oidc/daemon/daemon.go
```diff
@@ -142,20 +149,18 @@ func (a *App) serve(config daemonConfig) error {
...
 	daemon, err := daemon.New(ctx, s, daemonopts...)
 	if err != nil {
 		_ = s.Stop()
-		close(a.ready)
 		return err
 	}
 
 	a.daemon = daemon
-	close(a.ready)
+	closeFunc()
```
adombeck, line 163: We could just not call it here and let the deferred function call run on return, right? Then we don't even need the trick with checking the pointer and setting it to nil, because the function is always run exactly once.

3v1n0: Yeah, I was thinking about but I didn't want to change the behavior that the `close` happens *before* having finished serving, while by doing that it would happen after.

adombeck (approved):

PR #284, opened 2025-01-07
Use the msgraph host from the OpenID provider metadata

adombeck:
GCC High tenants use a different msgraph host. The hostname is part of the provider metadata that's retrieved during OpenID Connect Discovery via the `.well-known/openid-configuration` path.

This commit uses that hostname from the retrieved provider metadata in order to support GCC High tenants.

Closes https://github.com/ubuntu/authd/issues/708
UDENG-5709

### DIFF
--- a/internal/providers/msentraid/msentraid.go
+++ b/internal/providers/msentraid/msentraid.go
@@ -28,7 +28,11 @@ func init() {
 	pp.ColoringEnabled = false
 }
 
-const localGroupPrefix = "linux-"
+const (
+	localGroupPrefix   = "linux-"
+	defaultMSGraphHost = "graph.microsoft.com"
+	msgraphAPIVersion  = "v1.0"
+)
 
 // Provider is the Microsoft Entra ID provider implementation.
 type Provider struct {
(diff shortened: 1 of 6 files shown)


internal/providers/msentraid/msentraid.go
```diff
@@ -86,14 +90,34 @@ func (p Provider) GetExtraFields(token *oauth2.Token) map[string]interface{} {
...
+	}
+
+	return map[string]interface{}{
+		"msgraph_host": claims.MSGraphHost,
+	}, nil
+}
+
 // GetUserInfo is a no-op when no specific provider is in use.
-func (p Provider) GetUserInfo(ctx context.Context, accessToken *oauth2.Token, idToken *oidc.IDToken) (info.User, error) {
+func (p Provider) GetUserInfo(ctx context.Context, accessToken *oauth2.Token, idToken *oidc.IDToken, providerMetadata map[string]interface{}) (info.User, error) {
```
denisonbarbosa, line 109: Now that we are touching the function signature, maybe we should sneak in an update to the function comment? We missed updating it when adding this code to the `msentraid` package and it deserves a better comment.

denisonbarbosa (approved):
LGTM!

PS.: Added a small suggestion (I know it would be "cheating", but well...) so we can make the codebase better...

internal/providers/msentraid/msentraid.go
adombeck: good point, done

PR #285, opened 2025-01-07
Fix snap install hook

adombeck:
* Create `broker.conf` with permissions 0600
* Fix installation of snap failing with "install: Permission denied"

Closes https://github.com/ubuntu/authd/issues/716
UDENG-5713


didrocks (approved):
As debugged/discussed together.

PR #295, opened 2025-01-07
Hide --config flag

adombeck:
With 7a4d98054cb7d3d719e3a6f0700ccb6dc67e69b6, we automatically generate config files in the `.d` directory of the config file. We don't want to create those in `$PWD/mybroker.conf.d` when `--config mybroker.conf` is used.  

To avoid additional complexity, we decided to keep the current behavior but hide the flag from the usage message, so that it's only used for debugging and testing.  

Also hide the --paths-config flag for the same reason.

UDENG-5653


adombeck:
I tested that after installing a snap built from this branch, `snap run authd-msentraid --help` doesn't list the flags anymore.

didrocks (approved):
Indeed, it seems there is no other way to directly hide a flag object without going to the FlagSet and reference by name. Thanks for tackling this!

PR #298, opened 2025-01-08
Correctly handle pre-release versions in post-refresh hook

adombeck:
* Use SemVer versions
  * Separate the "dirty" annotation with a . instead of +
  * Versions must now have a patch version
*  Make the version comparison in the post-refresh hook SemVer compatible, so that 0.2.0-pre1 is considered lower than 0.2.0
* Set INITIAL_ALLOWED_USERS_VERSION="0.2.0-pre1" in the post-refresh hook
* Add special handling for the pre-release versions which were uploaded to the edge channels and already contain the allowed users config

UDENG-5714

TODO:
* [ ] Tag the commit "Set INITIAL_ALLOWED_USERS_VERSION to 0.2.0-pre1" with `0.2.0-pre1` after this has been merged to main
* [ ] Tag the same commit on the `msentraid` branch
* [ ] Tag the same commit on the `google` branch

### DIFF
--- a/snap/get_version
+++ b/snap/get_version
@@ -5,8 +5,8 @@ set -eu
 
 # When considering a tag, if starting with "<current branch name>-" it will have its prefix removed.
 # For instance:
-# * 0.1 -> 0.1
-# * msentraid-0.1 -> 0.1 on msentraid branch.
+# * 0.1.0 -> 0.1.0
+# * msentraid-0.1.0 -> 0.1.0 on msentraid branch.
 
 # 1. If current commit is tagged, the version is directly the tag name.
 # 2. If current commit is not tagged, the version is:
--- a/snap/get_version
+++ b/snap/get_version
@@ -13,7 +13,7 @@ set -eu
 #    a. <last tag name on current branch>+<commit_sha> for main branch.
 #    b. <last tag name on current branch (not in main)>+<commit_sha>.<last_commit_merged_from_main> for other branches.
 #
-# Any of those version will be annoted with +dirty if there are local changes.
+# Any of those version will be annoted with .dirty if there are local changes.
 
 # set_version will markup the version in the snapcraft.yaml file after amending it with a dirty markup if necessary.
 # $1: version: the version to set.
@@ -32,7 +32,7 @@ annotate_with_dirty() {
     # check if current tree content is dirty.
     is_dirty=$(git -C "${SNAPCRAFT_PART_SRC}" status --porcelain)
     if [ -n "${is_dirty}" ]; then
-        version="${version}+dirty"
+        version="${version}.dirty"
     fi
 
     echo "${version}"
--- a/snap/get_version
+++ b/snap/get_version
@@ -88,4 +88,10 @@ last_commit_on_main=$(git -C "${SNAPCRAFT_PART_SRC}" merge-base main HEAD)
 last_commit_on_main=$(git -C "${SNAPCRAFT_PART_SRC}" rev-parse --short=7 "${last_commit_on_main}")
 version="${version}.${last_commit_on_main}"
 
+# Check if the version is a valid semantic version.
+if ! semver check "${version}"; then
+    echo "Version ${version} is not a valid semantic version."
+    exit 1
+fi
+
 set_version "${version}"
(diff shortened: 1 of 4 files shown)


adombeck:
I see one edge case that's still a problem: When merging a PR which introduces a new migration, launchpad will automatically create and upload a new release on the edge channel, and I don't see a way to tag that commit with `0.3.0-pre1` before it does that. So users who install that version will get the new config file and when they update, it will still be migrated.

snap/get_version
```diff
@@ -88,4 +88,10 @@ last_commit_on_main=$(git -C "${SNAPCRAFT_PART_SRC}" merge-base main HEAD)
 last_commit_on_main=$(git -C "${SNAPCRAFT_PART_SRC}" rev-parse --short=7 "${last_commit_on_main}")
 version="${version}.${last_commit_on_main}"
 
+# Check if the version is a valid semantic version.
+if ! semver check "${version}"; then
```
didrocks, line 92: This will fail for snap build until we tag the new google branch though as it will have "notag"

snap/snapcraft.yaml
```diff
@@ -46,3 +46,11 @@ parts:
     build-packages:
       - git # The script needs Git.
     override-build: ./snap/get_version
```
didrocks, line 48: This steps depends on the semver step to be built, no? Also, does it have access to it in its path?

tools/semver/semver.go
```diff
@@ -0,0 +1,89 @@
...
+	fmt.Fprintf(os.Stderr, `Usage: %s <command> [arguments]
+
+Commands:
+  check <version>           Check if a version is valid
+  compare <ver1> <ver2>     Compare two versions
+
+Examples:
+  %s check 1.2.3           # Prints "valid" or "invalid"
+  %s compare 1.2.3 2.0.0   # Prints "less", "equal", or "greater"
+`, os.Args[0], os.Args[0], os.Args[0])
```
didrocks, line 52: I think you can replace the stuttering of `os.Args[0]` with `%[0]s` for each elements to replace

didrocks (changes requested):
I think some changes in term of build order (try to build the snap in a clean environment for instance) would be needed. Otherwise, a small enhancement suggestions.

On the topic of the initial race: I think this is an ok "edge" case (pun intended) for people running this track, which should be only temporary. So I wouldn’t worry too much about it.

snap/snapcraft.yaml
adombeck: > This steps depends on the semver step to be built, no?

Yes. I'm not sure how snapcraft determines the order of parts which have no dependencies, but apparently it did build it in the order we need, because the build succeeded every time I tried. Still definitely better to make the dependency explicit, so I added

```
    after:
      - semver
```

to the `version` part.

> Also, does it have access to it in its path?

Yes, apparently parts are built in an environment where the output `bin` directory of the other parts is already in the PATH.

snap/get_version
adombeck: That's why I added this TODO to the description:

> Tag the commit "Set INITIAL_ALLOWED_USERS_VERSION to 0.2.0-pre1" with 0.2.0-pre1 after this has been merged to main

Once that's done, both the msentraid and google snap should have `0.2.0-pre1`, right?

didrocks: IIRC, we don’t rely on the tag on main anymore but tag each branch manually with msentraid-<version> and google-<version>, no?

adombeck: Oh, I was not aware that's how we do it. I still think we should document our release process.

I added two more TODOs:
> Tag the same commit on the msentraid branch
> Tag the same commit on the google branch

tools/semver/semver.go
adombeck: done

snap/get_version
didrocks: You were the one convincing me to ignore the tag on the main branch :D

I think we’ll autotag. I’ll draft something for this.

snap/snapcraft.yaml
didrocks: ack!

didrocks (approved):
Thanks for the changes!

PR #301, opened 2025-01-09
Fix get_version script

adombeck:
Building from main or any other branch that's not "msentraid" or "google" failed because it used "notag" which is not a valid semantic version.

It now first tries to use the most recent tag prefixed with the branch name (e.g. "msentraid-0.2.0" when on the msentraid branch) and if there is none, just uses the most recent tag on the branch.


didrocks (approved):
Remember I mentioned the "notag" in our standup? :)

Anyway, I was only thinking of new branches for brokers that we released, but indeed, not ones from random branches we creates. I guess it makes sense. However, maybe you want to restrict this to tags on current branch or main rather than the more recent one?

adombeck:
> However, before merging, maybe you want to restrict this to tags on current branch or main rather than the more recent one? That would make a little bit more sense to me.

It's already restricted to the most recent one on the current branch:

```
tag=$(git tag --sort=-v:refname --merged="${current_branch}" | head -1)
```

adombeck:
> Anyway, I was only thinking of new branches for brokers that we released, but indeed, not ones from random branches we creates. I guess it makes sense.

To handle that, we also need to look for tags on main if we don't find a tag on the current branch. I can implement that.

adombeck:
> To handle that, we also need to look for tags on main if we don't find a tag on the current branch.

Ah actually, any new branch will already have tags from main, so I think it's fine as it is.

adombeck:
> True, but not if you branch from a "google" or "msentraid" branch. In that case, it will be one of those branches where the most recent tag will be taken and as the prefix won’t match, we end up with a non semver version.

Right. I don't see a good way to fix that, do you? Maybe we can ignore that edge case for now?

didrocks:
> Right. I don't see a good way to fix that, do you? Maybe we can ignore that edge case for now?

I think we can probably reuse the logic of things coming from the *current branch only* (which was in my first logic of get_version) or main. It’s a double call but it’s doable through git.

adombeck:
> I think we can probably reuse the logic of things coming from the current branch only (which was in my first logic of get_version) or main

How would that fix the "branched of from msentraid or google" issue? Would you ignore tags which are on more than one branch but not on main? Wouldn't that also break the msentraid/google branch when you branch off of it?

didrocks:
> How would that fix the "branched of from msentraid or google" issue? Would you ignore tags which are on more than one branch but not on main? Wouldn't that also break the msentraid/google branch when you branch off of it?

Because the tagged commit would not come from the "current branch" as soon as you committed to it. So it will only search from the "local" branch, and we can default to "main". This is really the preexisting logic from the first version of this script, which had this fallback.

adombeck:
> Because the tagged commit would not come from the "current branch" as soon as you committed to it.

I don't understand this sentence and you didn't explain which tag you would use/ignore based on whether it's on the current branch and/or main. We can jump into a call if you prefer.

didrocks:
> I don't understand this sentence and you didn't explain which tag you would use/ignore based on whether it's on the current branch and/or main. We can jump into a call if you prefer.

Let’s !

adombeck:
So what I understood in the call is that your proposal is:

1. Use the tag from the current branch *if it's not also on another branch*
2. Else, use the most recent tag from main

There's still trouble with the edge case that a new branch was branched of from msentraid/google, which will result in releases from both the new branch and msentraid/google using the tag from main. With the change from this PR, that edge case would result in the snap failing to build on the new branch (but not the msentraid/google branch) because it would use the `msentraid-0.2.0` tag as the version, which is not a valid SemVer. I didn't realize during the call, but I think the behavior that the msentraid/google branch use the tag from main is actually worse than letting builds on the new branch fail: If you create a new branch based on msentraid/google and push it, the versions of all releases from the msentraid/google branch (including automatic releases to the edge channel) change to the last version on main.

didrocks:
We discussed also fallbacking to latest tag, independently on the branch and make it semver compatible.

I think the use case is not that much of an edge one:
- I have a fix I want to test for Google or MSEntraID
- I want to easily build the snap from it to test it

What I would do:
1. branch google/msentraid
2. do my changes
3. try building the snap so that I have the correct configuration, build instructions in the snap and so on.

Here, 3. would fail because of the semver version. As long as we can make it build and installable, I don’t really care which tag is used as this is for tests mostly. I don’t think that having to tag (and then remove it), just for the sake of having this snap building and tested is a nice experience.

adombeck:
> We discussed also fallbacking to latest tag, independently on the branch and make it semver compatible.

You mean stripping any non-numeric prefix, so that `msentraid-0.2.0` becomes `0.2.0`, even if the branch is not `msentraid`, right? Let's do that then.

snap/get_version
```diff
@@ -38,35 +38,32 @@ annotate_with_dirty() {
...
-# strip_branch_tag_prefix will remove the branch name prefix from the tag name.
-# $1: tag: the tag name to strip the prefix from.
-# $2: current_branch: the branch name to strip from the tag.
+# strip_branch_tag_prefix removes any non-numeric prefix ending with a
+# dash (e.g. "msentraid-") from the tag name. We use this to remove the
+# branch name prefix from the tag name, but we do not just strip the
+# current branch name because we also want to support branching of a new
+# branch and use the latest tag from that branch (for example when
+# branching of the msentraid branch to test a fix, then that branch
+# should still use a valid version). $1: tag: the tag name to strip the
```
didrocks, line 47: nitpick: I would align $1 on a new line.

didrocks (approved):
Just one nitpick, but then, feel free to merge!

snap/get_version
adombeck: Oops, right, that was incorrectly formatted by the Rewrap extension I'm using :)

didrocks (approved):

adombeck:
I found another issue with the solution (sigh)

Since we're using `git tag --sort=-v:refname` to sort the tags, all `msentraid-*` tags will always be listed before any other tags (because `m` is higher than `g` and any numbers). So if I branch off branch `google` from branch `msentraid`, it will always use the last `msentraid-*` tag on the branch instead of the last `google` branch.

I pushed a change which uses a variation of my initial proposal:

```
# If there's a tag prefixed with the current branch name, that tag is
# used and the prefix is stripped. For example:
# * msentraid-0.1.0 -> 0.1.0
#
# Else, the highest version tag that starts with a number (in contrast
# to a prefix like "msentraid-") is used. For example:
# * 0.1.0 -> 0.1.0
```

I hope that works.

didrocks:
interesting idea, let us know!

adombeck:
It works in my tests

didrocks:
I didn’t test it, but I don’t spot anything obviously wrong. We can always revisit it if needed.

PR #302, opened 2025-01-09
Support showing INFO messages without DEBUG messages

adombeck:
Currently, we only show WARN messages when verbose mode is disabled, and when there's at least one "-v" flag, we show both INFO and DEBUG messages.

With this commit, we change that behavior to the same as authd: When there is one "-v" flag, we show INFO messages, when there's at least two, we also show DEBUG messages.


didrocks (approved):
This is the correct behaviour :)

PR #314, opened 2025-01-13
Use lowercase usernames

adombeck:
Broker part of https://github.com/ubuntu/authd/pull/723

Closes https://github.com/ubuntu/authd/issues/530, https://github.com/ubuntu/authd/issues/531
UDENG-4518
UDENG-4519


.github/workflows/ci.yml
```diff
@@ -21,13 +21,19 @@ jobs:
```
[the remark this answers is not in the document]
didrocks: Maybe convert it back to draft to prevent any accident?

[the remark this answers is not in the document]
adombeck: I'm afraid it won't show up on the reviewers' radar then :/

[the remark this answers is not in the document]
didrocks: we can keep the Jira card for reference?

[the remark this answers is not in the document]
adombeck: ok

adombeck:
Same as https://github.com/ubuntu/authd, this PR is also blocked by https://warthogs.atlassian.net/browse/UDENG-5830

snap/hooks/post-refresh
```diff
@@ -67,6 +68,38 @@ transition_to_allowed_users() {
...
+  # - previous-version is not set (that means that the previous version is
+  #   older than 0.2.0, i.e. the version where we introduced setting the
+  #   previous-version in the pre-refresh hook).
+  # - previous-version is less than the version where we introduced lowercase
+  [ -z "${PREVIOUS_VERSION}" ] || version_less_than "${PREVIOUS_VERSION}" "${INITIAL_LOWERCASE_USERNAMES_VERSION}"
+}
+
+transition_to_lowercase_usernames() {
+  log "Transitioning to lowercase usernames"
+    for f in "${SNAP_DATA}"/*/*; do
```
3v1n0, line 82: maybe we can include some more stuff in the wildcards? like `.d` or smth?

internal/broker/broker.go
```diff
@@ -832,8 +832,8 @@ func (b *Broker) fetchUserInfo(ctx context.Context, session *session, t *token.A
 	userInfo.Name = strings.ToLower(userInfo.Name)
 
 	// This means that home was not provided by the claims, so we need to set it to the broker default.
-	if !filepath.IsAbs(userInfo.Home) {
-		userInfo.Home = filepath.Join(b.cfg.homeBaseDir, userInfo.Home)
+	if userInfo.Home == "" {
+		userInfo.Home = filepath.Join(b.cfg.homeBaseDir, userInfo.Name)
```
3v1n0, line 836: I feel this commit deserves a clearer description

internal/broker/broker.go
```diff
@@ -455,6 +455,8 @@ func (b *Broker) IsAuthenticated(sessionID, authenticationData string) (string,
 		return AuthCancelled, string(msg), ctx.Err()
 	}
 
+	log.Debugf(context.Background(), "Authentication result for session %s: %s", sessionID, access)
```
3v1n0, line 458: What if the logging happens as a defer function that tells how we exited (something similar to what we do in pam.go authentication function)?

internal/broker/broker.go
```diff
@@ -154,6 +155,9 @@ func New(cfg Config, args ...Option) (b *Broker, err error) {
 func (b *Broker) NewSession(username, lang, mode string) (sessionID, encryptionKey string, err error) {
 	defer decorate.OnError(&err, "could not create new session for user %q", username)
 
+	// authd usernames are lowercase
+	username = strings.ToLower(username)
+
```
3v1n0, line 160: I feel this is not required anymore, since if we get here, then authd has already called the broker with a lower-case username, no?

At least, an authd version that is new enough... So maybe you wanted to cover partial upgrades here?

internal/broker/broker.go
```diff
@@ -812,6 +816,16 @@ func (b *Broker) fetchUserInfo(ctx context.Context, session *session, t *token.A
 		return info.User{}, fmt.Errorf("username verification failed: %w", err)
 	}
 
+	// authd uses lowercase usernames. To avoid that strings.ToLower folds different Unicode characters to the same
+	// lowercase character, we check that it only contains ASCII characters. If this ever causes issues, we can instead
+	// implement our own ToLower function that doesn't fold different characters to the same lowercase character.
+	//
+	// Note that we keep the Gecos field as it is, as it is not used for authentication.
```
3v1n0, line 828: Is this comment something that applies to authd conversions too?

internal/stringutils/stringutils.go
```diff
@@ -0,0 +1,12 @@
+// Package stringutils provides utility functions for string operations.
+package stringutils
+
+// IsASCII checks if a string contains only ASCII characters.
+func IsASCII(s string) bool {
+	for i := 0; i < len(s); i++ {
+		if s[i] > 127 {
```
3v1n0, line 7: Don't know if we want to mimic `isascii()` here, and in such case we could use the same logic (`(s[i] & ^0x7f) == 0`), however maybe we also want to actually want to mimic `isgraph` instead?

3v1n0 (commented):
Not sure how much of this we need to preserve as per the authd change, since I feel that most of stuff should still work here.

There are stuff we may still want to convert though, such as the configuration option values that mention suffixes or users allowed to log in

PR #315, opened 2025-01-14
Use log package from authd

adombeck:
For consistency and because it's more convenient (especially the `Debugf`, `Infof` etc. methods are more convenient to use than `slog.Debug(fmt.Sprintf(...))`).

UDENG-5773
UDENG-4523

# TODO

* [x] Use log package from `main` instead of `UDENG-5773-allow-importing-log-package` once https://github.com/ubuntu/authd/pull/727 has been merged


adombeck:
The failure of the code sanity job is tracked in https://warthogs.atlassian.net/browse/UDENG-5774

didrocks (approved):
I didn’t spot anything wrong. Nice work! I agree this is easier than `slog.Foo(fmt.Sprintf())`

3v1n0:
Great thing... Indeed it's a more consistent move.

Wondering if even this package could be somewhat shared in a private package... As I don't feel it would be fine in `authd.log` (unless we want to be committed to it...)

adombeck:
> Wondering if even this package could be somewhat shared in a private package...

Yes, @didrocks and I briefly discussed that a few days ago, with the result that we both think it could make sense to move this to a separate repo but we don't want to do that right now.

PR #316, opened 2025-01-15
ci: Install git-delta for nicer diffs in error messages

adombeck:
The diffs of mismatching golden files are passed through git-delta since https://github.com/ubuntu/authd-oidc-brokers/pull/207.

Also makes the test run on on Noble instead of Jammy, because git-delta is not packaged in Jammy.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,6 +28,12 @@ jobs:
         with:
           go-version-file: go.mod
 
+      - name: Install dependencies
+        run: |
+          set -eu
+          sudo apt update
+          sudo apt install -y git-delta
+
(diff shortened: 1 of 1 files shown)


didrocks (commented):
Was https://github.com/ubuntu/authd-oidc-brokers/pull/207 screenshot a CI run?

If no, can you make a run fail so that we see the delta in action rendered by Github CI itself?

adombeck:
There was one here: https://github.com/ubuntu/authd-oidc-brokers/actions/runs/11806570571/job/32891480199?pr=207#step:5:153

didrocks (approved):
Thanks! Excellent rendering! :)

3v1n0 (approved):

PR #319, opened 2025-01-16
Only use Microsoft Entra Security Groups

adombeck:
~~**Based on https://github.com/ubuntu/authd-oidc-brokers/pull/315, review and merge that first.**~~

Other groups might be created by non-admin users.

UDENG-5795


didrocks (approved):
Trusted you on the research of properties for security group. Nothing spotted if I correctly filtered what was in the log PR!

PR #324, opened 2025-01-17
Extract session mode constants

adombeck:
For improved maintainability and code navigation.


internal/broker/sessionmode/consts.go
```diff
@@ -0,0 +1,20 @@
...
+	LoginOld = "auth"
+	// ChangePassword is used when the session is for changing the user password.
+	ChangePassword = "change-password"
+	// ChangePasswordOld is the old name for the change-password session, which is now deprecated but still used by authd
+	// until all broker installations are updated.
+	ChangePasswordOld = "passwd"
+)
+
+// IsChangePasswordMode checks if the given mode is a change-password mode.
+func IsChangePasswordMode(mode string) bool {
```
didrocks, line 18: I would still had a small (3 test cases: each mode + an invalid one) tests. That way we have some clear documentation about what is supported or not in tests for the exported mode.

internal/providers/msentraid/msentraid.go
```diff
@@ -286,7 +287,7 @@ func (p Provider) CurrentAuthenticationModesOffered(
 	log.Debugf(context.Background(), "In CurrentAuthenticationModesOffered: sessionMode=%q, supportedAuthModes=%q, tokenExists=%t, providerReachable=%t, endpoints=%q, currentAuthStep=%d\n", sessionMode, supportedAuthModes, tokenExists, providerReachable, endpoints, currentAuthStep)
 	var offeredModes []string
 	switch sessionMode {
-	case "passwd":
+	case sessionmode.ChangePassword, sessionmode.ChangePasswordOld:
```
didrocks, line 290: see, I find is weird that we have the helper `IsChangePaswordMode`, but still duplicate in the case the need to have both here an in noprovider.

Maybe we should just drop `IsChangePasswordMode` and have the 2 cases listed in `broker.go` too? That way, the day we clean it up would not be only "this is the only place to change", but "I need to look at every places with `ChangePasswordOld`.

(Ofc, you can object that we will remove the const, and so, have to replace the cases there too, but it would mean remove `IsChangePasswordMode` and reintroducing the equality too).

didrocks (changes requested):
I just think we should drop the `IsChangePasswordMode` helper for the reasons I gave in the previous comment. Otherwise, looking good!

internal/providers/msentraid/msentraid.go
adombeck: done

internal/broker/sessionmode/consts.go
adombeck: obsolete

internal/broker/broker_test.go
```diff
@@ -170,17 +171,17 @@ func TestGetAuthenticationModes(t *testing.T) {
...
 
 		"Get_only_password_if_token_exists_and_provider_is_not_available":                {tokenExists: true, providerAddress: "127.0.0.1:31310", unavailableProvider: true},
 		"Get_only_password_if_token_exists_and_provider_does_not_support_device_auth_qr": {tokenExists: true, providerAddress: "127.0.0.1:31311", deviceAuthUnsupported: true},
 
-		// Passwd Session
-		"Get_only_password_if_token_exists_and_session_is_passwd":                      {sessionMode: "passwd", tokenExists: true},
-		"Get_newpassword_if_already_authenticated_with_password_and_session_is_passwd": {sessionMode: "passwd", tokenExists: true, secondAuthStep: true},
+		// Change password session
+		"Get_only_password_if_token_exists_and_session_is_passwd":                      {sessionMode: sessionmode.ChangePassword, tokenExists: true},
+		"Get_newpassword_if_already_authenticated_with_password_and_session_is_passwd": {sessionMode: sessionmode.ChangePassword, tokenExists: true, secondAuthStep: true},
```
didrocks, line 184: shouldn’t we have a test case with the old string there too?

didrocks (changes requested):

internal/providers/msentraid/msentraid.go
didrocks: thx!

internal/broker/sessionmode/consts.go
didrocks: Yeah, was just mentioning in case you didn’t want to remove that function :)

adombeck: Sure, I got that! The only reason I wrote that is to make it clear (both to you and me if I look at the PR again) that I consider this thread as done / ready to be resolved. Sorry if it came across as harsh, wasn't meant that way :)

internal/broker/broker_test.go
adombeck: good point! done

didrocks (approved):
Perfect, thanks!

PR #325, opened 2025-01-17
Automatic device authentication when refresh token is expired

adombeck:
~~**This is based on https://github.com/ubuntu/authd-oidc-brokers/pull/349, please review and merge that first**~~

When refreshing the token fails because the refresh token is expired, we should automatically start device authentication.

Closes https://github.com/ubuntu/authd/issues/575
UDENG-4911


adombeck:
I tested this by applying this patch:

```diff
commit 49c9c3aeb2d58d7b68cebd816ebb887befb080ca (HEAD -> 575-automatic-device-auth-on-expired-refresh-token)
Author: Adrian Dombeck <adrian.dombeck@canonical.com>
Date:   Mon Feb 3 15:36:36 2025 +0100

    XXX: Test expired refresh token

diff --git a/internal/broker/broker.go b/internal/broker/broker.go
index 93d0d82..193e2c8 100644
--- a/internal/broker/broker.go
+++ b/internal/broker/broker.go
@@ -810,6 +810,9 @@ func (b *Broker) updateSession(sessionID string, session session) error {
 func (b *Broker) refreshToken(ctx context.Context, oauth2Config oauth2.Config, oldToken token.AuthCachedInfo) (token.AuthCachedInfo, error) {
        timeoutCtx, cancel := context.WithTimeout(ctx, maxRequestDuration)
        defer cancel()
+
+       return token.AuthCachedInfo{}, &oauth2.RetrieveError{ErrorCode: "invalid_grant", ErrorDescription: "AADSTS50173: Refresh token is expired"}
+
        // set cached token expiry time to one hour in the past
        // this makes sure the token is refreshed even if it has not 'actually' expired
        oldToken.Token.Expiry = time.Now().Add(-time.Hour)

```

adombeck:
I created https://github.com/ubuntu/authd/issues/774 to track the "show a message" part.

internal/broker/broker_test.go
```diff
@@ -165,24 +166,25 @@ func TestGetAuthenticationModes(t *testing.T) {
...
+		"Get_device_auth_qr_if_next_auth_mode_is_device_qr": {nextAuthMode: authmodes.DeviceQr},
 
 		"Get_only_password_if_token_exists_and_provider_is_not_available":                {tokenExists: true, providerAddress: "127.0.0.1:31310", unavailableProvider: true},
 		"Get_only_password_if_token_exists_and_provider_does_not_support_device_auth_qr": {tokenExists: true, providerAddress: "127.0.0.1:31311", deviceAuthUnsupported: true},
 
 		// Change password session
-		"Get_only_password_if_token_exists_and_session_is_for_changing_password":                      {sessionMode: sessionmode.ChangePassword, tokenExists: true},
-		"Get_newpassword_if_already_authenticated_with_password_and_session_is_for_changing_password": {sessionMode: sessionmode.ChangePassword, tokenExists: true, secondAuthStep: true},
-		"Get_only_password_if_token_exists_and_session_mode_is_the_old_passwd_value":                  {sessionMode: sessionmode.ChangePasswordOld, tokenExists: true},
+		"Get_only_password_if_token_exists_and_session_for_changing_password":                   {sessionMode: sessionmode.ChangePassword, tokenExists: true},
```
denisonbarbosa, line 185: I don't understand this change...

internal/broker/broker_test.go
```diff
@@ -1183,6 +1198,8 @@ func TestUserPreCheck(t *testing.T) {
 }
 
 func TestMain(m *testing.M) {
+	log.SetLevel(log.DebugLevel)
+
```
denisonbarbosa, line 1202: I think it's better to only enable this if we run the tests with `-v`, no?

denisonbarbosa (commented):
Some considerations, but nothing big...

internal/broker/broker_test.go
adombeck: I'm interested in the debug output of failed tests, even if `-v` is not set

internal/broker/broker_test.go
adombeck: >  I will fix that.

done

denisonbarbosa: Yeah, it was just the line... Thanks for fixing it!

internal/broker/broker_test.go
denisonbarbosa: Yeah, but this will make tests always print debug logs from the source code even when not run with the verbosity flag, no?

adombeck: The `go test -v` flag does not control log verbosity, it controls whether logs are printed for succeeded tests. If not `go test` is run without `-v` flag, logs are only printed for failed tests. Does that clear things up? If not, please elaborate why you think it is undesirable behavior.

denisonbarbosa: My reasoning is basically: When you run just `go test ./...`, you should expect a simple and clear output of what passed and what failed. If some tests fail and need debugging, then you can run the test(s) that need debugging directly in order to interact with them (or run all again with `-v`, in cases where the test case fails only when executed with others). 

By always enabling debug, you take away the option of having a clean output which is not a good thing. I can see the benefits of having the source logs, but that should be conditional to a flag instead of enforced. Does that make sense?

adombeck: We do the same for several other tests already btw, search for `log.SetLevel(log.DebugLevel)`

adombeck: > I probably missed it in other reviews (or didn't spot the "issue" beforehand), I'm sorry for that.

No reason to be sorry, you were probably not the reviewer of the PRs which added debug levels to the other tests.

> Sure, but my point is that it's better to have control over that. I'm not questioning the usefulness of the logs, my issue here is that this forces the debug level even if you didn't ask for it. We can attach this behavior to go test ./... -v or even to a custom flag that (i.e. go test ./... -- --log_level=debug) and, since you prefer it this way, you can always run the tests with the flag enabled. Some failures can happen due to simple stuff or an issue with a test helper, so having debug messages from the source code always being printed could end up being detrimental to spotting the issue (mainly due to cluttering).

That's optimizing based on speculation, and it adds friction for those of us who want to see the debug logs. I enabled the debug logs to solve a real problem (not having enough information to debug the test failures). The error message is still printed at the bottom of the test, so I doubt that the debug logs will make it harder to understand the error if it's simple. I suggest that, if you (or anyone else) encounters the real problem that the debug logs make it harder to debug test failures, then we revisit the idea to add a custom flag for it.

denisonbarbosa: > I suggest that, if you (or anyone else) encounters the real problem that the debug logs make it harder to debug test failures, then we revisit the idea to add a custom flag for it.

This should be the other way around, though... It's better to add this constrainedly at first and then, if we find ourselves running the tests more frequently with the flags than without it, we can turn it into default behavior.

And I still prefer having the default `go test` behavior of a simple and clean output, even for failures, just telling me what failed, what was expected, what we got, and what should have happened. 

Also, as a side note, did you check if the log messages won't get all tangled up when running tests in parallel? Tangling up stdout and stderr is a very common annoyance of Go parallel testing...

adombeck: > This should be the other way around, though... It's better to add this constrainedly at first and then, if we find ourselves running the tests more frequently with the flags than without it, we can turn it into default behavior.

Look, I'm the only one working a lot on this code since months, and it doesn't look like that's going to change soon. I find myself repeatedly in the situation that I need the debug logs of a test to make the output more useful. Why are you insisting on making that harder for me?

adombeck: I don't want to put more time and energy into this discussion. Let's ask the rest of the team what they prefer and go with that. @didrocks, @3v1n0: Do you prefer having debug logs enabled (and seeing them printed for failing tests, both locally and the CI) or not? If not, I'll add something like this to the `TestMain` of the tests which I want debug logs for:

```
logLevel := os.Getenv("TESTS_LOG_LEVEL")
if logLevel != "" {
    log.SetLevel(log.LevelFromString(logLevel))
}
```

and put a wrapper for the `go` executable in my PATH which sets that environment variable.

adombeck: > The logs are not attached to the current failing tests. 

Sure, that's one of the bigger pain points with testing in Go. It's also true for all other log output (except when using `t.Log`) , not only the debug logs, so I don't really see why it should block enabling debug output.

> I think in the future, we should go back to attach the logger to the context, which will allow a per test debug logs, even under CI.

Did we have a solution for that before? Must have been before my time. I'm not sure what you mean with "attach the logger to the context". The only way I know of to have per test debug logs is to use `t.Log`. At a previous job, I solved that by writing an `io.Writer` which would call `t.Log` and could be used as the output writer of log functions. I could try to find that code.

> the default Go behaviour of "everything is passing" -> no log output, is preserved?

Yes. The commit only changes the log level, it doesn't change anything regarding when logs are printed in tests.

> if a test is failing, how much logs do we have printed? I think it’s not the only failing tests due to that, but all parallel tests in the current package, right? I can see myself living with this, but just want to understand where we stand.

The logs of all tests in the package containing the failing test are printed.

> I didn’t look at the PR, but I think those changes won’t interfere with the logs level tests, correct?

It won't.

adombeck: > I didn’t look at the PR

Btw the change in question is this single line: https://github.com/ubuntu/authd-oidc-brokers/pull/325/commits/7d26df35e705a68534f83b1500a9d4abee68e5ae

adombeck: > I’m not a big fan of the idea of the idea of t.Log and having non test code having that knowledge. What we could do, is:
> 

> 1. attaching a logger to a context that we pass through functions and calls that from the `log` package to write the logs after extracting it from the context.
> 2. in the production code, it’s a real logger that we instantiate and attach to the top level context (or any gRPC/dbus entries).
> 3. in the test code, we create a `logTestWriter` instance, which embeds a `testing.T`, and for each `Write()`, we call `t.Log()`.


Ok, that actually sounds quite similar to the approach I went with back then, which also attached a writer to the context, but an `io.Writer` and IIUC, you would attach a new interface 

```golang
type logger interface {
   Log(string)
}
```
?

That's seems like a fine approach and maybe even better. But that's a discussion for another ticket :)

denisonbarbosa: Yeah. I think it we will need to come back to this discussion later, probably once we finally decide to move `testutils` to its own thing. Let's unblock this and discuss later.

internal/broker/broker.go
```diff
@@ -211,102 +209,140 @@ func (b *Broker) connectToOIDCServer(ctx context.Context) (*oidc.Provider, error
...
+}
 
+func authModeIsAvailable(session session, authMode string) bool {
+	switch authMode {
+	case authmodes.Password:
+		return tokenExists(session)
+	case authmodes.NewPassword:
+		return true
+	case authmodes.Device, authmodes.DeviceQr:
+		return session.oidcServer != nil && session.oidcServer.Endpoint().DeviceAuthURL != "" && !session.isOffline
```
adombeck, line 289: @denisonbarbosa Regarding https://github.com/ubuntu/authd-oidc-brokers/pull/349#discussion_r1959880851, do you think we should extract this line into a method of the `Provider` interface?

adombeck: Great, I fully agree. Anything else missing on this PR for your approval?

PR #334, opened 2025-01-24
Add log messages

adombeck:
Adds some log messages, see commit messages for details.

UDENG-5876

### DIFF
--- a/internal/broker/broker.go
+++ b/internal/broker/broker.go
@@ -741,7 +741,7 @@ func (b *Broker) UserPreCheck(username string) (string, error) {
 	}
 
 	if !found {
-		return "", errors.New("username does not match the allowed suffixes")
+		return "", fmt.Errorf("username %q does not match any allowed suffix", username)
 	}
 
 	u := info.NewUser(username, filepath.Join(b.cfg.homeBaseDir, username), "", "", "", nil)
--- a/internal/broker/broker.go
+++ b/internal/broker/broker.go
@@ -634,6 +634,7 @@ func (b *Broker) handleIsAuthenticated(ctx context.Context, session *session, au
 	}
 
 	if !b.userNameIsAllowed(authInfo.UserInfo.Name) {
+		log.Warningf(context.Background(), "User %q is not in the list of allowed users", authInfo.UserInfo.Name)
 		return AuthDenied, errorMessage{Message: "permission denied"}
 	}
 
--- a/internal/dbusservice/methods.go
+++ b/internal/dbusservice/methods.go
@@ -1,7 +1,10 @@
 package dbusservice
 
 import (
+	"context"
+
 	"github.com/godbus/dbus/v5"
+	"github.com/ubuntu/authd/log"
 )
 
 // NewSession is the method through which the broker and the daemon will communicate once dbusInterface.NewSession is called.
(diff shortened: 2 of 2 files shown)


adombeck:
@didrocks @3v1n0 I would like to include this in the release, so I would be appreciate a quick review :)

internal/broker/broker.go
```diff
@@ -634,6 +634,7 @@ func (b *Broker) handleIsAuthenticated(ctx context.Context, session *session, au
 	}
 
 	if !b.userNameIsAllowed(authInfo.UserInfo.Name) {
+		log.Errorf(context.Background(), "User %q is not in the list of allowed users", authInfo.UserInfo.Name)
```
3v1n0, line 637: This is the only one that I'm unsure about... As IIRC we didn't want to give much information (neither in log form) whether someone was or not managed by the broker.

But I feel that's not a too big issue, given that normally users don't have access to the journal... That said, I'd leave the decision to @didrocks.

adombeck: As I mentioned in the commit message, it's considered a security risk to *not* log failed login attempts:

> To 1. allow the user/admin to figure out why authentication is denied,
and 2. log failed login attempts, which is security critical (security
logging failure is in the OWASP top 10, see
https://owasp.org/Top10/A09_2021-Security_Logging_and_Monitoring_Failures).

adombeck: @didrocks @3v1n0: I would like to include that in the upcoming release, so let's please try to move this forward

didrocks: yeah, I think this is not a big issue by itself. I would though downgrade the log to a warning: this is not an error of the service itself (something malfunctionning), but something to alert the sysadmin on if looking at reasons why something didn’t go to completion.

internal/dbusservice/methods.go
```diff
@@ -35,8 +38,10 @@ func (s *Service) SelectAuthenticationMode(sessionID, authenticationModeName str
 func (s *Service) IsAuthenticated(sessionID, authenticationData string) (access, data string, dbusErr *dbus.Error) {
 	access, data, err := s.broker.IsAuthenticated(sessionID, authenticationData)
 	if err != nil {
+		log.Errorf(context.Background(), "IsAuthenticated error: %v", err)
```
didrocks, line 41: ditto

didrocks (changes requested):
I would like we downgrade the log level, but otherwise, good!

internal/broker/broker.go
adombeck: Ah, good point! I see that `pam_unix` prints it's authentication failure messages with priority `notice`, which is between `info` and `warning`:

```
sudo[33720]: pam_unix(sudo:auth): authentication failure; logname=user uid=1000 euid=0 tty=/dev/pts/8 ruser=user rhost=  user=user
sshd[33337]: pam_unix(sshd:auth): authentication failure; logname= uid=0 euid=0 tty=ssh ruser= rhost=127.0.0.1  user=user
```

I'll do the same if you agree.

adombeck: It requires supporting the notice level in the `log` package in authd. I'll do that in a follow-up and use log level `warning` for now.

adombeck: done

internal/dbusservice/methods.go
adombeck: done

internal/broker/broker.go
adombeck: > I'll do that in a follow-up and use log level warning for now.

tracked in https://warthogs.atlassian.net/browse/UDENG-5879

didrocks (approved):

PR #336, opened 2025-01-24
Avoid repetition in syslog identifier (google)

adombeck:
The default syslog identifier set by snapd includes both the snap name and the service name. Both are the same for this snap, so it doesn't add any value and just clutters the logs.

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -17,6 +17,8 @@ apps:
     plugs:
       - network
     restart-condition: always
+    environment:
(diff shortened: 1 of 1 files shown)


adombeck:
@3v1n0, @didrocks: The [analogous PR for msentraid](https://github.com/ubuntu/authd-oidc-brokers/pull/335) was already approved and merged, so for consistency between our brokers, I would like to merge this one too before the upcoming release,

didrocks (approved):
Let’s merge it then!

PR #348, opened 2025-01-28
Fix post-refresh hook failing if broker.conf.d is empty

adombeck:
In that case it failed with:

```
cp: cannot stat '/snap/authd-msentraid/x1/conf/migrations/pre-0.2.0-pre1/broker.conf.d/*': No such file or directory
```


adombeck:
Converting to draft because I'm not sure we want to silently ignore the case that the `conf/migrations/pre-0.2.0-pre1` directory is empty

PR #350, opened 2025-01-29
Support changing the password if refreshing token or user info fails

adombeck:
We don't need to refresh the token and user info if the session is for changing the password.


didrocks (commented):
shouldn’t that be backed by some tests, or at least, should have impacted some?

adombeck:
I added a test

didrocks (approved):
Thanks for adding the tests!

PR #363, opened 2025-02-03
Log authentication failures caused by an incorrect local password

adombeck:
Unsuccessful authentication attempts are security relevant, so we should make sure to log all cases of those. See also
https://owasp.org/Top10/A09_2021-Security_Logging_and_Monitoring_Failures/

### DIFF
--- a/internal/broker/broker.go
+++ b/internal/broker/broker.go
@@ -574,6 +574,7 @@ func (b *Broker) handleIsAuthenticated(ctx context.Context, session *session, au
 				return AuthDenied, errorMessage{Message: "could not check password"}
 			}
 			if !ok {
+				log.Warningf(context.Background(), "Authentication failure: incorrect local password for user %q", session.username)
 				return AuthRetry, errorMessage{Message: "incorrect password"}
 			}
 


adombeck:
> However, I would like it to wrap those errors, and log them at the "edge" (either authd or the broker directly), by wrapping it. Does it make sense?

I don't think that works well with my proposal to use different log levels for these errors (see https://warthogs.atlassian.net/browse/UDENG-5921), for example log level `NOTICE` for [incorrect password attempts](https://github.com/ubuntu/authd-oidc-brokers/blob/9752d47124a3769383afb5e2a7a696783bb5af2f/internal/broker/broker.go#L577-L577) and log level `WARNING` when the [maximum number of attempts is exceeded](https://github.com/ubuntu/authd-oidc-brokers/blob/9752d47124a3769383afb5e2a7a696783bb5af2f/internal/broker/broker.go#L460-L460).

We could include the log level in the returned message struct or add special-case handling for specific messages on the "edge", but is it worth it? What's the benefit of moving the logging to the edge?

didrocks:
> We could include the log level in the returned message struct or add special-case handling for specific messages on the "edge", but is it worth it? What's the benefit of moving the logging to the edge?

The benefit I would see is that we would log everything that needs to be logged, by explicitely setting it and treating that in a single place.

Indeed, with the use cases you are mentioning, it means that we would need to put some kind of log level in the typed error too.

As you wish, I’m not fully decided or have strong opinions about those. I just wanted to raise it for thoughts. I’m happy to approve the PR if you feel this is unecessary

adombeck:
Ok, then let's keep it as it is for now.

didrocks (approved):

PR #376, opened 2025-02-10
Bump Go toolchain version to 1.23.6

adombeck:
govulncheck reports the following vulnerability in nistec@go1.23.5:

```
Vulnerability #1: GO-2025-3447
    Timing sidechannel for P-256 on ppc64le in crypto/internal/nistec
  More info: https://pkg.go.dev/vuln/GO-2025-3447
  Standard library
    Found in: crypto/internal/nistec@go1.23.5
    Fixed in: crypto/internal/nistec@go1.23.6
    Platforms: ppc64le
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -2,7 +2,7 @@ module github.com/ubuntu/authd-oidc-brokers
 
 go 1.23.0
 
-toolchain go1.23.5
+toolchain go1.23.6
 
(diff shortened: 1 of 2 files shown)


denisonbarbosa (approved):
Thanks for handling it fast!

PR #385, opened 2025-02-11
cmd/authd-oidc/daemon: Warn on no config file

MggMuggins:
Warn for "Non-critical entries that deserve eyes." A missing config file can cause the broker to fail to connect to the provider.

Default level is `2`, `Infof` is log level `0` and `Warningf` is `4`.

### DIFF
--- a/cmd/authd-oidc/daemon/config.go
+++ b/cmd/authd-oidc/daemon/config.go
@@ -49,7 +49,7 @@ func initViperConfig(name string, cmd *cobra.Command, vip *viper.Viper) (err err
 	if err := vip.ReadInConfig(); err != nil {
 		var e viper.ConfigFileNotFoundError
 		if errors.As(err, &e) {
-			log.Infof(context.Background(), "No configuration file: %v.\nWe will only use the defaults, env variables or flags.", e)
+			log.Warningf(context.Background(), "No configuration file: %v.\nWe will only use the defaults, env variables or flags.", e)
 		} else {
 			return fmt.Errorf("invalid configuration file: %w", err)
 		}


3v1n0 (approved):

adombeck:
Hi @MggMuggins, thanks for the contribution!

> A missing config file can cause the broker to fail to connect to the provider.

I think there's a misunderstanding here. There are three different config files:
1. `/etc/authd/authd.yaml` - That's the authd config file which the above log message refers to. It configures authd's log verbosity and the UID and GID ranges it uses for the users/groups it creates. I don't think it warrants a warning if this file is missing, log level INFO seems appropriate here.
2. `/etc/authd/brokers.d/msentraid.conf` (or `/etc/authd/brokers.d/google.conf` if you use the google broker) - Those tell authd which brokers it should use. If there is none, it won't use any broker, which is not an error but might indeed be interesting for the admin. I propose we log a message with log level NOTICE for that (which is for "Conditions that are not error conditions, but that may require special handling", see https://pubs.opengroup.org/onlinepubs/009695399/functions/syslog.html).
3. `/var/snap/authd-msentraid/current/broker.conf` (or `/var/snap/authd-google/current/broker.conf`) - That's the broker config file. That's the file for which what you describe is correct: If it's missing, the broker can't connect to the provider (because it doesn't know which issuer and client ID to use). The broker actually requires this file to exist, if it doesn't, it fails to start the and the journal contains these error messages:

```
Feb 11 17:19:58 ubuntu authd-msentraid[14841]: could not create broker: could not parse config: open /var/snap/authd-msentraid/137/broker.conf: no such file or directory
Feb 11 17:19:58 ubuntu systemd[1]: snap.authd-msentraid.authd-msentraid.service: Main process exited, code=exited, status=1/FAILURE
Feb 11 17:19:58 ubuntu systemd[1]: snap.authd-msentraid.authd-msentraid.service: Failed with result 'exit-code'.
```

MggMuggins:
> I think there's a misunderstanding here.

That's likely :slightly_smiling_face: 

Thanks for the feedback, this is helpful. I'm seeing this message even when `broker.conf` exists (ofc). Would it be possible to clarify which config file this log message refers to?

Likely the reason for my confusion: I don't have access to the environment I'm working with, just logs.  I was a bit hasty opening this PR; this is a red herring.

MggMuggins:
Adding the message for posterity:
```
No configuration file: Config File "authd-msentraid" Not Found in "[/var/snap/authd-msentraid/137 /root/snap/authd-msentraid/137 /etc/authd-msentraid /snap/authd-msentraid/137/bin]".
```

PR #386, opened 2025-02-11
internal/broker: Warn when switching to offline mode

MggMuggins:
If a user gives an invalid issuer URL in broker.conf, go-oidc will return an error with the HTTP response code:
```
Could not connect to the provider: 404 Not Found: . Starting session in offline mode.
```

We shouldn't leave users guessing as to the online/offline state of the broker, especially not the reason for that state choice.

I would have used `Noticef` here instead of `Warningf` but authd 0.4.1 doesn't have `Noticef`.

### DIFF
--- a/internal/broker/broker.go
+++ b/internal/broker/broker.go
@@ -183,7 +183,7 @@ func (b *Broker) NewSession(username, lang, mode string) (sessionID, encryptionK
 	// Construct an OIDC provider via OIDC discovery.
 	s.oidcServer, err = b.connectToOIDCServer(context.Background())
 	if err != nil {
-		log.Debugf(context.Background(), "Could not connect to the provider: %v. Starting session in offline mode.", err)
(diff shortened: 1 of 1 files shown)


adombeck:
> We shouldn't leave users guessing as to the online/offline state of the broker, especially not the reason for that state choice.

Makes sense to me!

> I would have used Noticef here instead of Warningf but authd 0.4.1 doesn't have Noticef.

True, but `main` has (since f8343ba416e286c698dc001f944a7ba38beaccdd) :) I pushed a commit which uses `Noticef` instead.

adombeck (approved):

MggMuggins:
Thanks, missed that 0.4.2 was in use in the gomod. Rebased this and added you in the `Reviewed-by`

adombeck:
Oh, I forgot about this PR. Sorry!

MggMuggins:
That's alright; I was intending to fix the tests but haven't gotten around to it. Looks like they're fixed anyway. Thanks!

PR #387, opened 2025-02-12
Ignore duplicate groups returned by the Microsoft Graph API

adombeck:
It's not clear in which case the Graph API returns multiple groups with the same name (or the same group twice), but we've seen it happen in #789.

Closes #789
UDENG-6030

### DIFF
--- a/internal/providers/msentraid/msentraid.go
+++ b/internal/providers/msentraid/msentraid.go
@@ -186,7 +186,10 @@ func (p Provider) getGroups(token *oauth2.Token, msgraphHost string) ([]info.Gro
 	}
 
 	var groups []info.Group
+	var msGroupNames []string
 	for _, msGroup := range graphGroups {
+		var group info.Group
+
 		idPtr := msGroup.GetId()
 		if idPtr == nil {
 			log.Warning(context.Background(), pp.Sprintf("Could not get ID for group: %v", msGroup))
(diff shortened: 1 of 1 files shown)


adombeck:
AFAICT, we don't have any tests which test the `GetUserInfo` implementation of the `msentraid` provider (and the `getGroup` function called by it). I created a Jira issue and added it to the next pulse: https://warthogs.atlassian.net/browse/UDENG-6033

internal/providers/msentraid/msentraid.go
```diff
@@ -186,35 +186,75 @@ func (p Provider) getGroups(token *oauth2.Token, msgraphHost string) ([]info.Gro
 	}
 
 	var groups []info.Group
+	var msGroupNames []string
```
3v1n0, line 189: Wouldn't be better to use a set for this `e.g. a map[string]struct{}`? And maybe use `unicode.SimpleFold` to add both cases there?

adombeck: To be able to include the unfolded group names in the log message, we need to store it unfolded. We could use a `map[string]string` for that, but tbh I'm not sure the performance benefit is worth the effort

adombeck: > Wouldn't be better to use a set for this e.g. a map[string]struct{}?

Or do you want to store the unfolded name in that map? Then we would still need to iterate over it so there is no performance benefit

adombeck:
@3v1n0 ping?

denisonbarbosa (approved):

internal/providers/msentraid/msentraid.go
3v1n0: Ok, fair enough.

3v1n0 (approved):

PR #406, opened 2025-02-26
Rename `NoProvider` to `GenericProvider`

denisonbarbosa:
After some discussion, we decided that `GenericProvider` is a better naming for the basic provider implementation, so let's go with that instead of using `NoProvider`.

UDENG-4958

### DIFF
--- a/internal/providers/default.go
+++ b/internal/providers/default.go
@@ -3,10 +3,10 @@
 package providers
 
 import (
-	"github.com/ubuntu/authd-oidc-brokers/internal/providers/noprovider"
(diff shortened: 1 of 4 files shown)


adombeck:
I would prefer if you would squash the two commits into one because the first commit doesn't build, which is annoying when you use `git bisect` to find a commit which broke something.

denisonbarbosa:
> I would prefer if you would squash the two commits into one because the first commit doesn't build, which is annoying when you use `git bisect` to find a commit which broke something.

I will. I did it separately to facilitate the review (to avoid the diff showing as a new file and so on...)

adombeck (approved):

PR #408, opened 2025-02-26
Fix device auth being the default when local password is available

adombeck:
Commit 8ded61bf39b64dabd7f9543aee9e47874177c60e changed the order of the
available modes, putting password authentication last.

The order is actually important, because authd picks the first one.

This PR fixes the order and adds a comment explaining why the order
is important.

Closes https://github.com/ubuntu/authd/issues/812
UDENG-6176

### DIFF
--- a/internal/broker/broker.go
+++ b/internal/broker/broker.go
@@ -261,15 +261,18 @@ func (b *Broker) availableAuthModes(session session) (availableModes []string, e
 
 	switch session.mode {
 	case sessionmode.ChangePassword, sessionmode.ChangePasswordOld:
-		// session is for changing the password
+		// Session is for changing the password.
 		if !tokenExists(session) {
 			return nil, errors.New("user has no cached token")
 		}
 		return []string{authmodes.Password}, nil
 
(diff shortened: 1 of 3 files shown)


adombeck:
There is already the test case `TestGetAuthenticationModes/Get_password_and_device_auth_qr_if_token_exists` which tests this, but the golden file was also updated by 8ded61bf39b64dabd7f9543aee9e47874177c60e, so the test passed. That's one issue I repeatedly encounter with the golden files: They don't make it explicit which parts of the content are relevant for the test case, which puts a lot of effort on the developers to consider whether golden file updates are correct or not, and can lead to mistakes like the one here.

adombeck:
I added a commit which explicitly defines the expected first auth mode returned by `GetAuthenticationModes`.

3v1n0 (approved):

denisonbarbosa (approved):
Thanks for also adding tests for it! Indeed the order was incorrect and I didn't spot that in the previous PR, but now we should be covered. Well done!

PR #415, opened 2025-02-28
Add tests for EntraID's specific GetUserInfo

denisonbarbosa:
UDENG-6033


internal/providers/msentraid/msgraphmock_test.go
```diff
@@ -0,0 +1,93 @@
...
+	if handler == nil {
+		handler = simpleGroupHandler
+	}
+	mockServer := httptest.NewServer(handler)
+	return mockServer.URL, mockServer.Close
+}
+
+// simpleGroupHandler simulates a successful response with a list of groups.
+func simpleGroupHandler(w http.ResponseWriter, r *http.Request) {
+	response := map[string]interface{}{
```
3v1n0, line 20: nit:

`any` here too?

internal/providers/msentraid/msentraid_test.go
```diff
@@ -115,3 +118,63 @@ func TestVerifyUsername(t *testing.T) {
...
+		groupEndpointHandler http.HandlerFunc
+
+		wantErr bool
+	}{
+		"Successfully_get_user_info":                               {},
+		"Successfully_get_user_info_with_local_groups":             {groupEndpointHandler: localGroupHandler},
+		"Successfully_get_user_info_with_mixed_groups":             {groupEndpointHandler: mixedGroupHandler},
+		"Successfully_get_user_info_filtering_non_security_groups": {groupEndpointHandler: nonSecurityGroupHandler},
+
+		"Error_when_connecting_to_server":                {providerMetadata: map[string]any{"msgraph_host": "invalid"}, wantErr: true},
```
adombeck, line 139:
```suggestion
		"Error_when_msgraph_host_is_invalid":                {providerMetadata: map[string]any{"msgraph_host": "invalid"}, wantErr: true},
```

internal/providers/msentraid/msentraid_test.go
```diff
@@ -115,3 +118,63 @@ func TestVerifyUsername(t *testing.T) {
...
+		"Error_when_getting_user_claims":                 {invalidIDToken: true, wantErr: true},
```
adombeck, line 140:
```suggestion
		"Error_when_id_token_claims_are_invalid":                 {invalidIDToken: true, wantErr: true},
```

internal/providers/msentraid/msentraid_test.go
```diff
@@ -115,3 +118,63 @@ func TestVerifyUsername(t *testing.T) {
...
+		"Error_when_getting_token_scopes":                {tokenScopes: map[string]any{"scope": struct{ notAString int }{10}}, wantErr: true},
```
adombeck, line 141:
```suggestion
		"Error_when_token_scopes_have_incorrect_type":                {tokenScopes: map[string]any{"scope": struct{ notAString int }{10}}, wantErr: true},
```

internal/providers/msentraid/msentraid_test.go
```diff
@@ -115,3 +118,63 @@ func TestVerifyUsername(t *testing.T) {
...
+		"Error_when_token_does_not_have_required_scopes": {tokenScopes: map[string]any{"scope": "not the required scopes"}, wantErr: true},
+		"Error_when_getting_user_groups":                 {groupEndpointHandler: errorGroupHandler, wantErr: true},
```
adombeck, line 143:
```suggestion
		"Error_when_getting_user_groups_fails":                 {groupEndpointHandler: errorGroupHandler, wantErr: true},
```

internal/providers/msentraid/msentraid_test.go
```diff
@@ -115,3 +118,63 @@ func TestVerifyUsername(t *testing.T) {
...
+			p := msentraid.New()
+			got, err := p.GetUserInfo(context.Background(), accessToken, idToken, tc.providerMetadata)
+			if tc.wantErr {
+				require.Error(t, err, "GetUserInfo should return an error")
+				return
+			}
+			require.NoError(t, err, "GetUserInfo should not return an error")
+
+			want := golden.LoadWithUpdateYAML(t, got)
+			require.Equal(t, want, got, "GetUserInfo should return the expected user info")
```
adombeck, line 177: We have `CheckOrUpdateYAML` now, which does the equality check for us (and displays the diff with `delta` if available).

```suggestion
			golden.CheckOrUpdateYAML(t, got)
```

internal/providers/info/info.go
```diff
@@ -42,3 +42,8 @@ func NewUser(name, home, uuid, shell, gecos string, groups []Group) User {
 
 	return u
 }
+
+// Claimer is an interface that defines a method to extract the claims from the ID token.
+type Claimer interface {
```
adombeck, line 47: I don't like this interface name, in this context "Claim" is a noun, the name "Claimer" makes it sound like it can claim something.

How about we just name the interface "IDToken"?

adombeck (commented):
Excellent work!

internal/providers/info/info.go
denisonbarbosa: Although I also don't think that the `Claimer` name is perfect, I think we should keep the conventions [defined by Go](https://go.dev/doc/effective_go):

> By convention, one-method interfaces are named by the method name plus an -er suffix or similar modification to construct an agent noun: Reader, Writer, Formatter, CloseNotifier etc.

Sometimes this ends up in some weird names (like `Stringer`) but it does indicate quite well the method that the type must implement in order to satisfy the interface.

adombeck: I know the convention, but here it results in a misleading name. `Claim` can be both a noun and a verb. In this context it's a noun, but naming the interface `Claimer` gives the impression that the method name is a verb, i.e. that it can claim something. Anyway, if you still disagree, let's keep it as is.

adombeck (approved):

PR #437, opened 2025-03-14
Snap description refresh

didrocks:
Update the snap description and icon to match the one existing on the store.

UDENG-6199


snap/snapcraft.yaml
```diff
@@ -1,7 +1,27 @@
 name: authd-google
-summary: Google Broker for authd
+summary: Authd - Google IAM Cloud Authentication for Ubuntu
 description: |
-  Broker that enables OIDC authentication with Google for authd.
+  Authd is a powerful authentication service for Ubuntu, designed to integrate with cloud identity providers like
+  Google IAM. It delivers a secure, flexible solution for organizations and individuals who are transitioning
+  to cloud-based identity management on Ubuntu workstations and servers.
+  Authd uses the Oauth Device Authorization Grant and ensures a consistent and secure login experience across
```
adombeck, line 7: Nitpick: It's spelled "OAuth" not "Oauth". I see that the description in the snap store also uses Oauth, so I think we should also change it there.

```suggestion
  Authd uses the OAuth Device Authorization Grant and ensures a consistent and secure login experience across
```

snap/snapcraft.yaml
```diff
@@ -1,7 +1,27 @@
...
+  Ubuntu Desktop and Server — whether through GDM, SSH, or network services like NFS and Samba.
+
+  Key Features
+  * Cloud identity provider Integration: Connects with Google IAM
+  * Secure Login: Authd leverages the Oauth Device Authorization Grant RFC 8628-compliant workflows for reliability
```
adombeck, line 12:
```suggestion
  * Secure Login: Authd leverages the OAuth Device Authorization Grant RFC 8628-compliant workflows for reliability
```

adombeck (approved):

snap/snapcraft.yaml
didrocks: Thanks for spotting it! Normally, once we merge this and I enable "override from snapcraft file", the store description should be synced too (but I will do that after manual  testing).

PR #449, opened 2025-03-17
Fix checkGroupIsDuplicate

adombeck:
It returned false if there was a duplicate and true if there was none.


denisonbarbosa (approved):
Thanks for fixing it quickly!

PR #459, opened 2025-03-18
Add option to force token refresh on login

adombeck:
This option forces refreshing the token on login, so that user accounts which were disabled or removed from the identity provider can't log in anymore.

Closes #813
UDENG-6177

# TODO
* [x] Add a test case to `internal/broker/config_test.go`
  * The implementation depends on which config section we add the new option to, so let's decide that first.

### DIFF
--- a/conf/broker.conf
+++ b/conf/broker.conf
@@ -6,6 +6,18 @@ client_id = <CLIENT_ID>
 ## client secret to authenticate with the provider.
 #client_secret = <CLIENT_SECRET>
 
+## Always authenticate with the identity provider during login, even if
+## a local method (e.g. local password) is used.
+## This works by forcing a token refresh at login, which fails if the
+## user does not have the necessary permissions in the identity provider.
+##
+## If set to false (the default), authentication with the identity
+## provider is only done if the provider is reachable during login.
+##
+## Important: Enabling this option prevents authd users from logging in
+## if the identity provider is unreachable (e.g. due to network issues).
+#force_provider_authentication = false
+
 [users]
 ## The directory where the home directories of new users are created.
 ## Existing users will keep their current home directory.
--- a/internal/broker/broker.go
+++ b/internal/broker/broker.go
@@ -626,8 +626,12 @@ func (b *Broker) handleIsAuthenticated(ctx context.Context, session *session, au
 			return AuthNext, nil
 		}
 
+		if b.cfg.forceProviderAuthentication && session.isOffline {
+			return AuthDenied, errorMessage{Message: "could not refresh token: provider is not reachable"}
+		}
+
 		// Refresh the token if we're online even if the token has not expired
-		if !session.isOffline {
+		if b.cfg.forceProviderAuthentication || !session.isOffline {
 			authInfo, err = b.refreshToken(ctx, session.oauth2Config, authInfo)
 			var retrieveErr *oauth2.RetrieveError
 			if errors.As(err, &retrieveErr) && b.provider.IsTokenExpiredError(*retrieveErr) {
--- a/internal/broker/broker.go
+++ b/internal/broker/broker.go
@@ -95,8 +95,6 @@ type Option func(*option)
 
 // New returns a new oidc Broker with the providers listed in the configuration file.
 func New(cfg Config, args ...Option) (b *Broker, err error) {
-	defer decorate.OnError(&err, "could not create broker")
-
 	p := providers.CurrentProvider()
 
 	if cfg.ConfigFile != "" {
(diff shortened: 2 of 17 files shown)


conf/broker.conf
```diff
@@ -1,3 +1,14 @@
+[general]
+## Force a token refresh on login with the local password. This has the
+## effect that the user is only allowed to log in if they still have
+## the required permissions in the identity provider.
+## If this option is set to false (the default), the token is only
+## refreshed if the provider is reachable.
+## Important: If this option is enabled, authd users won't be able to
+## log in with the local password if the identitiy provider is
+## unreachable (e.g. because there is no working internet connection).
+#force_token_refresh_on_login = false
```
adombeck, line 10: To the reviewer(s): I felt like the option didn't really fit in the existing sections (`oidc` and `users`) so I added this new `general` section. I'm open for other proposals.

adombeck, line 10: I've chosen `force_token_refresh_on_login` as the option name because it most accurately describes what it does, but I'm open to other suggestions. Some other ideas that have been mentioned on https://github.com/ubuntu/authd/issues/813:
* `force_online_check`
* `force_users_online_check`
* `force_provider_access_check`

3v1n0: Yeah, I agree with this, naming of the section can be discussed, but this one looks good to me

3v1n0: `force_token_refresh_on_authentication`? However this also applies to the password change case, right?

Assuming that's done for an admin user-A that changes the password for B? (as A passwd for themselves is just implicit in the initial auth).

adombeck: > Assuming that's done for an admin user-A that changes the password for B? (as A passwd for themselves is just implicit in the initial auth).

Ah, that's a case we don't support yet, right? AFAICT, we only support a user changing their own password.

internal/broker/config.go
```diff
@@ -175,6 +183,11 @@ func parseConfigFile(cfgPath string, p provider) (userConfig, error) {
 		return cfg, fmt.Errorf("config file has invalid values, did you edit the file %q?\n%w", cfgPath, err)
 	}
 
+	general := iniCfg.Section(generalSection)
+	if general != nil {
+		cfg.forceTokenRefresh = general.Key(forceTokenRefreshOnLoginKey).MustBool(false)
```
3v1n0, line 188: Why not using Bool and returning the error in case?

conf/broker.conf
3v1n0: I think we do... At least once your're super user passwd just works as expected for anyone else.

didrocks: We will when you change your password as you have to authenticate through PAM.

I like `force_online_check` or `force_online_authentication` maybe? (or something with `provider` as you mentioned). I think both the `online` only part + the fact that we check the validity of the user should be conveyed in the name. I wouldn’t go to the "token refresh" technical part IMHO.

Similarly, I wouldn’t restrict to "local password" the first line of the comment:
`Always check for user validity check when authenticating.` or something along that line. When we will add more authentication method, even local (think about FIDO keys for instance), this would still be valid. We can still write something along the line of `even when authentication method is purely local (password, …)` if you want to illustrate this. Does it make sense?

adombeck: How about `force_provider_authentication`? That feels less ambiguous to me than `force_online_check` and `force_online_authentication` and IMO still conveys that the provider needs to be reachable.

didrocks: I like `force_provider_authentication`!

adombeck: @didrocks are you also fine with this new `general` section?

adombeck: now that we have `provider` in the option name (`force_provider_authentication`), we could also consider moving it to the `oidc` section, because it's related to the OIDC provider.

didrocks: I think it’s fine to move to the oidc section to avoid creating a new one, wdyt?

internal/broker/broker.go
```diff
@@ -95,14 +95,12 @@ type Option func(*option)
...
 func New(cfg Config, args ...Option) (b *Broker, err error) {
-	defer decorate.OnError(&err, "could not create broker")
-
 	p := providers.CurrentProvider()
 
 	if cfg.ConfigFile != "" {
 		cfg.userConfig, err = parseConfigFile(cfg.ConfigFile, p)
 		if err != nil {
-			return nil, fmt.Errorf("could not parse config: %v", err)
+			return nil, fmt.Errorf("could not parse config file '%s': %v", cfg.ConfigFile, err)
```
denisonbarbosa, line 103: Nitpick: you can just use `%q` instead of `'%s'`

3v1n0 (approved):

PR #472, opened 2025-03-26
Fix invalid msgraph host

adombeck:
Since commit 5fc98520c45294ffb85bb27a81929e2ec1b89fcb, we store the full
URL to the msgraph host in the provider metadata. However, the metadata
on disk is not updated, so it can still contain only the domain name of
the msgraph host, resulting in a failure to get the groups via msgraph:

    Could not fetch user info: could not get user info: failed to get user groups: failed to get user groups: Get "graph.microsoft.com/me/transitiveMemberOf/graph.group": unsupported protocol scheme "". Using cached user info.

This PR fixes that by supporting both the old and new formats of the
msgraph host field in the provider metadata.

Closes https://github.com/ubuntu/authd/issues/858
UDENG-6537


denisonbarbosa (approved):
Thanks for addressing this quickly! LGTM!


PS.: Sorry for having missed that 😅

PR #514, opened 2025-04-16
 Don't overwrite shell when refreshing user info

adombeck:
The user's shell should be configurable locally and not overwritten when the user info is being refreshed.

Closes https://github.com/ubuntu/authd/issues/880

### DIFF
--- a/internal/broker/broker.go
+++ b/internal/broker/broker.go
@@ -644,7 +644,7 @@ func (b *Broker) handleIsAuthenticated(ctx context.Context, session *session, au
 		}
 
 		// Try to refresh the user info
-		userInfo, err := b.fetchUserInfo(ctx, session, &authInfo)
+		userInfo, err := b.refreshUserInfo(ctx, session, &authInfo, authInfo.UserInfo)
 		if err != nil && authInfo.UserInfo.Name == "" {
 			// We don't have a valid user info, so we can't proceed.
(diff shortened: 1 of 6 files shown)


internal/broker/broker.go
```diff
@@ -838,6 +838,19 @@ func (b *Broker) refreshToken(ctx context.Context, oauth2Config oauth2.Config, o
...
 
+func (b *Broker) refreshUserInfo(ctx context.Context, session *session, t *token.AuthCachedInfo, oldUserInfo info.User) (info.User, error) {
+	userInfo, err := b.fetchUserInfo(ctx, session, t)
+	if err != nil {
+		return info.User{}, fmt.Errorf("could not refresh user info: %w", err)
+	}
+
+	// Don't overwrite the shell and home directory, because those should be configurable locally.
+	userInfo.Shell = oldUserInfo.Shell
+	userInfo.Home = oldUserInfo.Home
```
3v1n0, line 849: This works fine with the case in which the home changes (gets moved) because the broker prefix has?

adombeck:
This should actually be handled in authd instead of the broker, because it should depend on whether the user is already present in the authd database, not whether the broker has cached user info or not. So I'm closing this in favor of https://github.com/ubuntu/authd-oidc-brokers/pull/514.

PR #521, opened 2025-05-09
Support specifying extra groups in broker config file

adombeck:
Multiple users asked for a feature to automatically add users to a list of local groups via local configuration (instead of configuration in the identity provider, which is currently only supported for Microsoft Entra ID).

This PR adds two new options, `extra_groups` and `owner_extra_groups`.

`extra_groups` can be set to a comma-separated list of local groups which each authd user will be added to upon login.

`owner_extra_groups` is the same, except that only the owner is added to these groups upon login.


conf/broker.conf
```diff
@@ -1,5 +1,5 @@
 [oidc]
-issuer = https://<ISSUER_URL>
+issuer = https://login.microsoftonline.com/<ISSUER_ID>/v2.0
```
denisonbarbosa, line 2: Forgot to remove this after testing?

conf/broker.conf
```diff
@@ -56,3 +56,13 @@ client_id = <CLIENT_ID>
...
+
+## A comma-separated list of local groups which authd users will be
+## added to upon login.
+## Example: extra_groups = users
+#extra_groups =
+
+## Like 'extra_groups', but only the owner user will be added to these
+## groups.
+## Example: owner_extra_groups = sudo,lpadmin
+#owner_extra_groups =
```
denisonbarbosa, line 68: Maybe we should be more specific about what "owner" means?

denisonbarbosa (commented):
Source changes look good to me. I do have some questions on the config example, though...

denisonbarbosa (approved):
LGTM!

PR #546, opened 2025-06-12
Store directory extension attributes in token.json

adombeck:
To support the use case that groups are assigned additional data via directory extensions, which should be accessible to admins of the local system, we now store the directory extension attributes of a group in the token.json file in the broker's data directory.

This is on request of a user who wants to have shared user and group IDs on all systems using authd, and wants to achieve that by setting IDs via directory extensions, then read those once the user logged in and use authctl to change the user/group ID to the one from the directory extension (once authctl supports that).

The directory extension attributes of user objects can already be stored in the token.json file, by emitting them as claims in the ID token (which is stored in JWT format in the RawIDToken field of token.json): https://learn.microsoft.com/en-us/entra/identity-platform/schema-extensions

Closes https://github.com/ubuntu/authd/issues/963
UDENG-7195

### DIFF
--- a/internal/providers/info/info.go
+++ b/internal/providers/info/info.go
@@ -3,8 +3,9 @@ package info
 
 // Group represents the group information that is fetched by the broker.
 type Group struct {
-	Name string `json:"name"`
-	UGID string `json:"ugid"`
+	Name        string         `json:"name"`
+	UGID        string         `json:"ugid"`
+	ExtraFields map[string]any `json:"extra_fields,omitempty"`
 }
 
 // User represents the user information obtained from the provider.
--- a/internal/providers/msentraid/msentraid.go
+++ b/internal/providers/msentraid/msentraid.go
@@ -220,6 +220,18 @@ func (p Provider) getGroups(token *oauth2.Token, msgraphHost string) ([]info.Gro
 			continue
 		}
 
+		// Store directory extension attributes, if any.
+		additionalData := msGroup.GetAdditionalData()
+		if additionalData != nil {
+			group.ExtraFields = make(map[string]any)
+			for k, v := range additionalData {
+				// Directory extension attributes start with "extension_"
+				if strings.HasPrefix(k, "extension_") && v != nil {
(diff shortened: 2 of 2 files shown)


ajm370:
> @ajm370, you can install the authd-msentraid broker built from this branch via `sudo snap install authd-msentraid --channel edge/963-group-extra-fields` (or, if you already have a version of authd-msentraid installed, update it via `sudo snap refresh authd-msentraid --channel edge/963-group-extra-fields`).
> 
> The directory extension attributes of the user object which are configured to be emitted as claims on the token are available in the JWT stored in the `RawIDToken` field of the token.json. It can be parsed via jq like this:
> 
> ```shell
> jq -r '
>   .RawIDToken
>   | split(".")[1]
>   | gsub("-"; "+") | gsub("_"; "/") |
>     . + (["", "==", "="][(length % 4)])
>   | @base64d
>   | fromjson
> ' /var/snap/authd-msentraid/current/<issuer>/<user>/token.json
> ```
> 
> The directory extension attributes of the group object are stored in the `UserInfo`, like this:
> 
> ```
>   "UserInfo": {
>     [...]
>     "groups": [
>       {
>         "name": "test-group",
>         "ugid": "69228f9d-410c-4e5e-8ec0-e1aef8288ba0",
>         "extra_fields": {
>           "extension_f7128cd7b23a4086bf5c7d4ba710e707_Linux_GID": 1010
>         }
>     ]
>   }
> ```

Many thanks again for putting this together; in terms of testing, is authctl also bundled within this edge channel? Cheers

ajm370:
@adombeck, that looks to be working and gives us a way to consistently enumerate GID's across systems (once authctl is available). Thanks!

Regarding UID's, would it be possible to implement a similar function that pulls down the on-premises custom extension attribute for the user in question? I'm not sure if this could look something like the following request+response:
[https://learn.microsoft.com/en-us/graph/extensibility-overview?tabs=javascript#request](https://learn.microsoft.com/en-us/graph/extensibility-overview?tabs=javascript#request)

PR #560, opened 2025-06-27
Update behavior for `ssh_allowed_suffixes` setting

denisonbarbosa:
We used to allow every user to authenticate for the first time through SSH if this setting was not configured. This behavior could be misleading and result in unexpected situations for the users.

To improve this, we now consider " " (empty string) to mean 'disallow all' and have a new "*" value to represent the 'allow all' option, in case users still want to keep this.

UDENG-7132

### DIFF
--- a/conf/broker.conf
+++ b/conf/broker.conf
@@ -25,9 +25,26 @@ client_id = <CLIENT_ID>
 ## The home directories are created in the format <home_base_dir>/<username>
 #home_base_dir = /home
 
+## By default, SSH only allows logins from users that already exist on the
+## system.
+## New authd users (who have never logged in before) are *not* allowed to log
+## in for the first time via SSH unless this option is configured.
+##
 ## If configured, only users with a suffix in this list are allowed to
-## log in via SSH. The suffixes must be separated by comma.
-#ssh_allowed_suffixes = @example.com,@anotherexample.com
+## authenticate for the first time directly through SSH.
+## Note that this does not affect users that already authenticated for
+## the first time and already exist on the system.
+##
+## Suffixes must be comma-separated (e.g., '@example.com,@example.org').
+## To allow all suffixes, use a single asterisk ('*').
+##
+## Example:
+##     ssh_allowed_suffixes_first_auth = @example.com,@anotherexample.org
+##
+## Example (allow all):
+##     ssh_allowed_suffixes_first_auth = *
+##
+#ssh_allowed_suffixes_first_auth =
 
 ## 'allowed_users' specifies the users who are permitted to log in after
 ## successfully authenticating with the identity provider.
--- a/internal/broker/config.go
+++ b/internal/broker/config.go
@@ -36,13 +36,14 @@ const (
 	ownerKey = "owner"
 	// homeDirKey is the key in the config file for the home directory prefix.
 	homeDirKey = "home_base_dir"
-	// SSHSuffixKey is the key in the config file for the SSH allowed suffixes.
-	sshSuffixesKey = "ssh_allowed_suffixes"
+	// sshSuffixesKey is the key in the config file for the SSH allowed suffixes.
+	sshSuffixesKey = "ssh_allowed_suffixes_first_auth"
+	// sshSuffixesKeyOld is the old key in the config file for the SSH allowed suffixes. It should be removed later.
+	sshSuffixesKeyOld = "ssh_allowed_suffixes"
 	// extraGroupsKey is the key in the config file for the extra groups to add to each authd user.
 	extraGroupsKey = "extra_groups"
 	// ownerExtraGroupsKey is the key in the config file for the extra groups to add to the owner.
 	ownerExtraGroupsKey = "owner_extra_groups"
-
 	// allUsersKeyword is the keyword for the `allowed_users` key that allows access to all users.
 	allUsersKeyword = "ALL"
 	// ownerUserKeyword is the keyword for the `allowed_users` key that allows access to the owner.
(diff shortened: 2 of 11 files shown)


internal/broker/broker.go
```diff
@@ -778,7 +778,10 @@ func (b *Broker) CancelIsAuthenticated(sessionID string) {
 func (b *Broker) UserPreCheck(username string) (string, error) {
 	found := false
 	for _, suffix := range b.cfg.allowedSSHSuffixes {
-		if strings.HasSuffix(username, suffix) {
+		if suffix == "" {
+			continue
+		}
+		if suffix == "*" || strings.HasSuffix(username, suffix) {
```
adombeck, line 784: One thing that was brought up when we discussed this in yesterday's standup is that if we support the special value `*`, users will assume that they can also use `*@foo.org` to match all usernames ending with `@foo.org`.

We could use  https://github.com/gobwas/glob to also support that syntax.

adombeck: mmh, if we support globs, it would make more sense to name the option `allowed_ssh_users` than `allowed_ssh_suffixes`, because you can configure patterns which do not only match suffixes

conf/broker.conf
```diff
@@ -25,9 +25,22 @@ client_id = <CLIENT_ID>
...
+## authenticate for the first time directly through SSH.
+## Note that this does not affect users that already authenticated for
+## the first time and already exist on the system.
+##
+## Example:
+##     ssh_allowed_suffixes = @example.com,@anotherexample.com
+##
+## The suffixes must be separated by comma. In order to allow all suffixes
+## to be able to authenticate, '*' can be used (e.g ssh_allowed_suffixes = *)
+#ssh_allowed_suffixes =
```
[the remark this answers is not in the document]
denisonbarbosa: I'm not sure it's worth the change, tbh. It would create more variants in our code and require us to handle more migrations for the configuration file. Since it's just a configuration key and the explanation is already improved by quite a bit, maybe we can avoid renaming this to keep our lives a bit simpler?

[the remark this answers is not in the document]
denisonbarbosa: Good point. I'll update it!

internal/broker/broker.go
adombeck: Definitely in favor of keeping it simple. Full glob support (and renaming the setting accordingly) now also seems too much to me. However, supporting a leading `*`, i.e. `*@domain.com` in addition to `@domain.com`, seems simple enough to me that I would prefer to support it (via the small diff I pasted).

conf/broker.conf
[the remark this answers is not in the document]
3v1n0: Mhmh... IMHO migrationg would be better, so we avoid confusion, but also it opens room for adding an actual `ssh_allowed_suffixes` option that just may in future, just check all the users

internal/broker/broker.go
denisonbarbosa: I don't like the idea of having both `*`, `*@domain.com`, and `@domain.com` as valid values. I think it might be more harmful than helpful for the understanding here. What if we stop naming it as `suffixes` and use something like `allowed_domains` instead?

adombeck: What do you see as potentially harmful in supporting both `*@domain.com` and `@domain.com`? I genuinely want to understand.

In case that there's a misunderstanding, I only want to support the case that the prefix *starts* with a single `*`. I don't want to support the case that `*` appears anywhere else but the start. I wouldn't even document `*@domain.com`, it's just meant to gracefully handle the case that users (understandably, IMO) use `*@domain.com` if they see that `*` is supported.

conf/broker.conf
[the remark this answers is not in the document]
denisonbarbosa: I agree that migrating is probably not worth it as well. However, the naming is still up for debate as well. Do you guys think that the better explanation is not enough?

internal/broker/config.go
```diff
@@ -121,7 +124,12 @@ func (uc *userConfig) populateUsersConfig(users *ini.Section) {
 	}
 
 	uc.homeBaseDir = users.Key(homeDirKey).String()
+
 	uc.allowedSSHSuffixes = strings.Split(users.Key(sshSuffixesKey).String(), ",")
+	// If the value is empty, we should also try reading the old key as well.
+	if len(uc.allowedSSHSuffixes) == 1 && uc.allowedSSHSuffixes[0] == "" {
+		uc.allowedSSHSuffixes = strings.Split(users.Key(sshSuffixesKeyOld).String(), ",")
+	}
```
adombeck, line 132: This will use the old key even if the new key is present but set to the empty value. I think we should instead use `users.HasKey` instead to determine if the new is present, and only fall back to the old key if it is not:

```diff
diff --git a/internal/broker/config.go b/internal/broker/config.go
index 663abeb..01644bc 100644
--- a/internal/broker/config.go
+++ b/internal/broker/config.go
@@ -130,7 +130,11 @@ func (uc *userConfig) populateUsersConfig(users *ini.Section) {
 	}
 
 	uc.homeBaseDir = users.Key(homeDirKey).String()
-	uc.allowedSSHSuffixes = strings.Split(users.Key(sshSuffixesKey).String(), ",")
+	if users.HasKey(sshSuffixesKey) {
+		uc.allowedSSHSuffixes = strings.Split(users.Key(sshSuffixesKey).String(), ",")
+	} else {
+		uc.allowedSSHSuffixes = strings.Split(users.Key(sshSuffixesKeyOld).String(), ",")
+	}
 
 	if uc.allowedUsers == nil {
 		uc.allowedUsers = make(map[string]struct{})
```

PR #567, opened 2025-06-29
Fix UserPreCheck D-Bus interface

adombeck:
The D-Bus interface definition of the `UserPreCheck` method was missing the out parameter for the user info it returns.


3v1n0:
Oh... Right... Since you're there can you also fix it in authd's `examplebroker/com.ubuntu.auth.ExampleBroker.xml`?

denisonbarbosa (approved):

PR #568, opened 2025-06-29
UserPreCheck: Don't return error if user is not permitted

adombeck:
Lets make it easier for the caller to differentiate between the expected error that the user is not permitted to log in and an unexpected error, by returning an empty string in the first case and an error in the latter.


adombeck:
> In order to check this in the tests we currently have, and to keep the behavior consistent, I feel we should also adapt the example broker and adjust the integration tests on authd side.

Right! I planned to do that but then forgot.

3v1n0 (approved):

PR #570, opened 2025-06-30
msentraid: Do not mention client secret in broker.conf

adombeck:
Entra applications used for authd-msentraid can't use a client secret (see https://github.com/ubuntu/authd/issues/1005), so we should not document it in the broker.conf file shipped with authd-msentraid, to avoid that users are misled into incorrectly using a client secret.

Closes https://github.com/ubuntu/authd/issues/1005 
UDENG-7295


denisonbarbosa (approved):
Completely agreed!

PR #574, opened 2025-07-10
Return a custom error when a IsAuthenticated call was canceled

adombeck:
This allows the caller to detect this case and handle it differently than unexpected error (by not logging an error message).

### DIFF
--- a/internal/dbusservice/methods.go
+++ b/internal/dbusservice/methods.go
@@ -2,6 +2,7 @@ package dbusservice
 
 import (
 	"context"
+	"errors"
 
 	"github.com/godbus/dbus/v5"
 	"github.com/ubuntu/authd/log"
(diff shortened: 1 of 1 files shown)


internal/dbusservice/methods.go
```diff
@@ -68,3 +72,8 @@ func (s *Service) UserPreCheck(username string) (userinfo string, dbusErr *dbus.
 	}
 	return userinfo, nil
 }
+
+// makeCanceledError creates a dbus.Error for a canceled operation.
+func makeCanceledError() *dbus.Error {
+	return &dbus.Error{Name: "com.ubuntu.authd.Canceled"}
```
3v1n0, line 78: Nice, maybe we want to be even more specific?


```suggestion
	return &dbus.Error{Name: "com.ubuntu.authd.IsAuthenticated.Canceled"}
```

3v1n0 (approved):

internal/dbusservice/methods.go
3v1n0: Yeah, I had similar thinking, just mentioned in case... But you can go with this one

PR #575, opened 2025-07-12
snap: Run `go mod vendor` as part of the broker build

adombeck:
I had a snap build fail because I didn't run `go mod vendor` before running `snapcraft build`. That should really be done as part of the snap build.

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -56,6 +56,7 @@ parts:
       - go
     override-build: |
       go mod download all
+      go mod vendor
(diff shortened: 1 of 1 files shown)


denisonbarbosa (approved):
Looks good to me. Do you have any idea why this started to happen?

adombeck:
> Do you have any idea why this started to happen?

It happens for example if you change to a branch which has changes to the go.mod. I don't think this behavior is new. I have a `build-msentraid-snap` script which runs `go mod vendor` (and a git pre-push hook which also runs `go mod vendor`), which might be why I didn't encounter it that often before.

PR #585, opened 2025-07-14
msentraid: Support registering device upon login using libhimmelblau

adombeck:
Device registration enables Entra administrators to monitor which users have logged in on which devices. It’s also a prerequisite for Single Sign-On (SSO), which we plan to support in the future.

This change adds a `register_device` setting to the authd configuration, which controls automatic device registration upon user login.

If the device is disabled in Entra ID, user login is denied.

Closes https://github.com/ubuntu/authd/issues/697
UDENG-7385

## TODO
* [x] Decide if setting should be moved to a different section
* [ ] ~~Refactor broker to not have Entra-ID-specific code in the general broker implementation~~ To be done in a follow-up, tracked in https://warthogs.atlassian.net/browse/UDENG-8324
* [x] Test with a tenant without P2 license

### DIFF
--- a/internal/broker/broker.go
+++ b/internal/broker/broker.go
@@ -782,6 +782,11 @@ func (b *Broker) passwordAuth(ctx context.Context, session *session, secret stri
 		log.Errorf(context.Background(), "Login failed: %s", err)
 		return AuthDenied, errorMessage{Message: "This device is disabled in Microsoft Entra ID, please contact your administrator."}
 	}
+	if errors.Is(err, himmelblau.ErrInvalidRedirectURI) {
+		// Deny login if the redirect URI is invalid, so that users and administrators are aware of the issue.
+		log.Errorf(context.Background(), "Login failed: %s", err)
+		return AuthDenied, errorMessageForDisplay(err, "Invalid redirect URI")
+	}
 	var tokenAcquisitionError himmelblau.TokenAcquisitionError
 	if errors.As(err, &tokenAcquisitionError) {
 		log.Errorf(context.Background(), "Token acquisition failed: %s. Try again using device authentication.", err)


internal/providers/msentraid/himmelblau.go
```diff
@@ -0,0 +1,199 @@
...
+		machineKey,
+		&userToken,
+	)
+	if ret != C.SUCCESS {
+		return "", fmt.Errorf("failed to acquire token by refresh token: %d", int(ret))
+	}
+
+	var accessToken *C.char
+	defer C.free(unsafe.Pointer(accessToken))
+	ret = C.user_token_access_token(userToken, &accessToken)
```
dmulder, line 191: FYI, this response token has a PRT attached, which you appear to have thrown away (IIUC). You might want to be caching that PRT here (it will be useful for SSO, etc).

adombeck: It's correct that we don't store the full `UserToken` returned by `broker_acquire_token_by_refresh_token` yet. That's because to perform device registration, we only need the access token from that user token.

We do plan to look into SSO via the PRT obtained here in the future. I'll add a comment explaining that we will need to cache the full `UserToken` for that.

dmulder: This isn't how you'd obtain a token for enrollment. You need to call one of `acquire_token_interactive_for_device_enrollment`, `initiate_acquire_token_by_mfa_flow_for_device_enrollment`/`acquire_token_by_mfa_flow`, or `initiate_device_flow_for_device_enrollment`/`acquire_token_by_device_flow` to get that token.

dmulder: Or if you already have an authenticated token, you could use `acquire_token_by_refresh_token_for_device_enrollment`.

internal/providers/msentraid/himmelblau.go
```diff
@@ -0,0 +1,199 @@
...
+		nil,
+		// We have to use the edge browser client ID here, else the GET request to the
+		// https://login.microsoftonline.com/<tenant-ID>/oauth2/v2.0/authorize endpoint
+		// fails with "AADSTS500113: no reply address is registered for the application".
+		// Note: Alternatively, we could use `nil` here if we also use `nil` as the client ID
+		// in the `broker_init` call, which means that the user doesn't even have to register
+		// an OIDC app in Entra. However, that has the effect that we can't fetch the groups
+		// of the user.
+		// TODO: The above comment is outdated, we now use the client ID of the broker app
+		C.CString(clientID),
```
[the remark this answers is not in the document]
adombeck: I see how the comment can be confusing. Sorry about that, this code is *not* in a state meant for review yet, I only pushed it to see if the tests pass in the CI (that's why it's still marked as a draft).

I am indeed referring to a custom app registration. Currently, the authd Entra ID broker requires users to register a custom app with the `GroupMember.Read.All` scope: https://documentation.ubuntu.com/authd/latest/howto/configure-authd/#application-registration. That's the app we are using here, which allows us to also fetch the user's groups via the graph API with the token we obtain.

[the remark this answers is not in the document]
adombeck: > Was just seeing how I could help :)

That's much appreciated :)

internal/providers/msentraid/himmelblau.go
```diff
@@ -0,0 +1,188 @@
...
+var registered = false
+
+func (p *Provider) registerDevice(ctx context.Context, token *oauth2.Token, tenantID, domain string) error {
+	if registered {
+		return nil
+	}
+
+	var err *C.MSAL_ERROR
+
+	authority := C.CString("https://login.microsoftonline.com/" + tenantID)
```
3v1n0, line 68:
```suggestion
	authority := C.CString("https://login.microsoftonline.com/" + tenantID)
	defer C.free(unsafe.Pointer(authority))
```

`broker_init` is not taking the ownership of it, right?

internal/providers/msentraid/himmelblau.go
```diff
@@ -0,0 +1,188 @@
...
+		nil, /* cert_key */
+		&brokerClientApp,
+	)
+	if err != nil {
+		panic(fmt.Sprintf("failed to initialize BrokerClientApplication: %v", C.GoString(err.msg)))
+	}
+
+	var attrs *C.EnrollAttrs
+
+	cDomain := C.CString(domain)
```
3v1n0, line 82: Leak?

internal/providers/msentraid/himmelblau.go
```diff
@@ -0,0 +1,188 @@
...
+
+	err = C.enroll_attrs_init(
+		cDomain,
+		C.CString(hostname()),
```
3v1n0, line 86: Leak?

internal/providers/msentraid/himmelblau.go
```diff
@@ -0,0 +1,188 @@
...
+		nil, /* device_type - default is "Linux" */
+		0,   /* join_type - 0: Azure AD join */
+		C.CString(OSVersion()),
```
3v1n0, line 89: Leak?

internal/providers/msentraid/himmelblau.go
```diff
@@ -0,0 +1,188 @@
...
+
+	var transportKey *C.LoadableMsOapxbcRsaKey
+	defer C.loadable_ms_oapxbc_rsa_key_free(transportKey)
+	var certKey *C.LoadableMsDeviceEnrolmentKey
+	defer C.loadable_ms_device_enrollment_key_free(certKey)
+	var deviceID *C.char
+	defer C.free(unsafe.Pointer(deviceID))
+	err = C.broker_enroll_device(
+		brokerClientApp,
+		C.CString(token.RefreshToken),
```
3v1n0, line 104: Leak?

internal/providers/msentraid/himmelblau.go
3v1n0, line 86: Why not failing on error here?

internal/providers/msentraid/himmelblau.go
```diff
@@ -0,0 +1,188 @@
...
+	"golang.org/x/oauth2"
+)
+
+var tpm *C.BoxedDynTpm
+var authValue *C.char
+var loadableMachineKey *C.LoadableMachineKey
+var machineKey *C.MachineKey
+var brokerClientApp *C.BrokerClientApplication
+
+func init() {
```
3v1n0, line 28: Wouldn't be better to move this to a init version that is controlled manually and we can handle the error for?

internal/providers/msentraid/himmelblau.go
```diff
@@ -0,0 +1,188 @@
...
+		panic(fmt.Sprintf("failed to create machine key: %v", C.GoString(err.msg)))
+	}
+
+	err = C.tpm_machine_key_load(tpm, authValue, loadableMachineKey, &machineKey)
+	if err != nil {
+		panic(fmt.Sprintf("failed to load machine key: %v", C.GoString(err.msg)))
+	}
+}
+
+var registered = false
```
3v1n0, line 59: We should protect this with a mutex (atomic logic is not enough, unless you use two variables).

internal/providers/msentraid/himmelblau.go
```diff
@@ -0,0 +1,188 @@
...
+		&certKey,
+		&deviceID,
+	)
+	if err != nil {
+		return fmt.Errorf("failed to enroll device: %v", C.GoString(err.msg))
+	}
+
+	log.Infof(ctx, "Enrolled device with ID: %v", C.GoString(deviceID))
+
+	registered = true
```
3v1n0, line 118: Maybe do this as a defer function initially if `err` is not `nil`?

internal/providers/msentraid/himmelblau.go
```diff
@@ -0,0 +1,188 @@
...
+func hostname() string {
+	name, err := os.Hostname()
+	if err != nil {
+		return "unknown"
+	}
+	return name
+}
+
+// OSVersion gets the pretty name of the OS release from the system.
+func OSVersion() string {
```
3v1n0, line 132: In authd we use gorbe.io/go/osrelease for this, maybe it's better to rely on something else?

internal/providers/msentraid/himmelblau.go
```diff
@@ -0,0 +1,188 @@
...
+		brokerClientApp,
+		C.CString(token.RefreshToken),
+		scopesPtr,
+		C.int(len(scopes)),
+		nil, /* request_resource */
+		// We could use `nil` here instead of the client ID if we also use `nil` as the client ID
+		// in the `broker_init` call, which means that the user doesn't even have to register
+		// an OIDC app in Entra. However, that has the effect that we can't fetch the groups
+		// of the user.
+		C.CString(clientID),
```
3v1n0, line 168: Leak?

.github/workflows/ci.yml
```diff
@@ -85,15 +103,64 @@ jobs:
...
 
+      - name: Run msentraid tests (with Address Sanitizer)
+        if: matrix.test == 'asan'
+        env:
+          # Do not optimize, keep debug symbols and frame pointer for better
+          # stack trace information in case of ASAN errors.
+          CGO_CFLAGS: "-O0 -g3 -fno-omit-frame-pointer"
+          G_DEBUG: "fatal-criticals"
+          GO_TESTS_TIMEOUT: 30m
+          AUTHD_TESTS_SLEEP_MULTIPLIER: 1.5
```
3v1n0, line 114: IIRC unused in this branch too

internal/providers/msentraid/generate-himmelblau.sh
```diff
@@ -0,0 +1,29 @@
...
+TARGET_TRIPLE=$(rustc -vV | awk '/host:/ {print $2}')
+
+mkdir -p himmelblau
+"${CARGO_HOME:-$HOME/.cargo}"/bin/cbindgen --config ./cbindgen.toml > himmelblau/himmelblau.h
+cargo cbuild --release --lib --features=broker,changepassword,on_behalf_of
+
+# Copy header and shared library
+cp "target/${TARGET_TRIPLE}/release/himmelblau.h" "${SCRIPT_DIR}/"
+cp "target/${TARGET_TRIPLE}/release/libhimmelblau.so" "${SCRIPT_DIR}/"
+ln -sf "libhimmelblau.so" "${SCRIPT_DIR}/libhimmelblau.so.0"
```
3v1n0, line 29: As before:


```suggestion
cp "target/${TARGET_TRIPLE}/release/libhimmelblau.so" "${SCRIPT_DIR}/libhimmelblau.so.1"
ln -sf "${SCRIPT_DIR}/libhimmelblau.so.1" "libhimmelblau.so"
```

internal/providers/msentraid/himmelblau.go
adombeck: I'm against pulling in dependencies for trivial stuff like this, because every additional dependency increases the risk of a supply chain attack

internal/providers/msentraid/himmelblau.go
```diff
@@ -0,0 +1,188 @@
...
+
+func hostname() string {
+	name, err := os.Hostname()
+	if err != nil {
+		return "unknown"
+	}
+	return name
+}
+
+// OSVersion gets the pretty name of the OS release from the system.
```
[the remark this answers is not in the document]
adombeck: added a comment

internal/providers/msentraid/generate-himmelblau.sh
adombeck: is there a reason why you used `.1` instead of `.0`?

internal/providers/msentraid/himmelblau.go
```diff
@@ -0,0 +1,372 @@
...
+// Since we're running in a snap, this returns the version of the core base snap
+// (which is not that helpful when it's shown as the device's OS in Entra, so
+// might want to change this in the future, to somehow get the host's OS version).
+func OSVersion() string {
+	data, err := os.ReadFile("/etc/os-release")
+	if err != nil {
+		return "unknown"
```
3v1n0, line 278: Maybe warn

internal/providers/msentraid/himmelblau.go
```diff
@@ -0,0 +1,372 @@
...
+	}
+
+	for _, line := range strings.Split(string(data), "\n") {
+		if strings.HasPrefix(line, "PRETTY_NAME=") {
+			return strings.Trim(strings.TrimPrefix(line, "PRETTY_NAME="), "\"")
+		}
```
3v1n0, line 284:
```suggestion
		if name, found := strings.CutPrefix(line, "PRETTY_NAME=")); found {
			return name
		}
```

Then if you want to use the happy path, would be in an extra line.

internal/providers/msentraid/himmelblau.go
adombeck: To finally follow up on this:

> FYI, this response token has a PRT attached, which you appear to have thrown away

We are now caching more things than when you wrote this comment, but still not the PRT (which is part of the "user token" if I understand correctly). We're caching the data that is needed to load the TPM machine key (`exported_key` and `auth_value`), to initialize the broker app (`transport_key` and `cert_key`), the device ID, and the refresh token we get via OIDC from the Microsoft Authentication Broker.

We use that refresh token in the call to `broker_enroll_device` and then again in a call to `broker_acquire_token_by_refresh_token` to get a user token, from which we use the access  token (retrieved via `user_token_access_token`) to access the Microsoft Graph API (to fetch the user's groups).

In the future (e.g. when we add support for SSO), we might also store the user token (along with the PRT contained in it), but for now I don't see any use case for storing it.

Do you see any issues with that approach?

internal/providers/msentraid/himmelblau.go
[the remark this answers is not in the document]
adombeck: The PR is now ready for review, in case you still feel like taking a look :) 

By the way, I’d also like to thank you for reviewing and merging my upstream PRs so quickly!

dmulder (commented):
I've left a few comments/suggestions, but otherwise this is looking good.

internal/providers/msentraid/himmelblau.go
adombeck: >  FYI, take a look at https://github.com/himmelblau-idm/identity_dbus_broker when you look into implementing SSO. 

Thanks for the pointer, noted!

internal/providers/msentraid/generate-himmelblau.sh
```diff
@@ -0,0 +1,27 @@
...
+cargo install cargo-c
+
+cd "${GIT_DIR}/third_party/libhimmelblau"
+
+TARGET_TRIPLE=$(rustc -vV | awk '/host:/ {print $2}')
+
+cargo cbuild --release --lib --features=broker,changepassword,on_behalf_of
+
+# Copy header and shared library
+cp "target/${TARGET_TRIPLE}/release/himmelblau.h" "${SCRIPT_DIR}/"
```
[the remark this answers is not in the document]
adombeck: It is already in `.gitignore`: https://github.com/ubuntu/authd-oidc-brokers/blob/cb0e76ccfb39ce6f27d00e6d9cddc7f5cfad01b8/internal/providers/msentraid/.gitignore

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "third_party/libhimmelblau"]
+	path = third_party/libhimmelblau
+	url = https://gitlab.com/adombeck/libhimmelblau.git
```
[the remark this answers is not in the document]
adombeck: Good question. The fork allowed to me try out my patches before submitting them upstream, but I don't think I'll need that anymore once https://gitlab.com/samba-team/libhimmelblau/-/merge_requests/115 was merged (and if I do, I can temporarily change the remote to my fork again).

internal/broker/broker.go
```diff
@@ -270,22 +283,73 @@ func (b *Broker) availableAuthModes(session session) (availableModes []string, e
...
-		return tokenExists(session) && passwordFileExists(session)
+		if !tokenExists(session) {
+			log.Debugf(context.Background(), "Token does not exist for user %q, so local password authentication is not available", session.username)
+			return false
+		}
+
+		if !passwordFileExists(session) {
+			log.Debugf(context.Background(), "Password file does not exist for user %q, so local password authentication is not available", session.username)
+			return false
+		}
```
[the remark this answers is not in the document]
adombeck: > I'll just comment here that I strongly disagree with your local password approach.

Oh, interesting! I would like to understand your concern better. Why do you disagree with the approach? Is it because we're not securing the password and credentials with a TPM, or that we could improve the UX by using a Hello key?

> (a short lived refresh_token, I assume?)

Yes, we store the refresh token we get via OIDC and fetch a new set of tokens during login (if the IdP can be reached). It's not that short lived though, it expires after 3 months.

> You already have the TPM object, after all.

We do have a TPM object now. Unfortunately, for now, I don't think we can actually use a hardware TPM as its backend, because of current snap limitations (see https://github.com/ubuntu/authd/issues/472#issuecomment-2298484647). I'll double check that, maybe it's just `systemd-creds` which is affected by the limitation and the way `kanidm_hsm_crypto` uses the TPM is supported. And a soft TPM does not have any actual security benefit as far as I know.

internal/providers/msentraid/himmelblau.go
adombeck: done

internal/providers/msentraid/msentraid.go
```diff
@@ -368,19 +418,80 @@ func (p Provider) VerifyUsername(requestedUsername, authenticatedUsername string
...
+
+	// Extract the tenant ID from the issuer URL
+	// https://login.microsoftonline.com/8de88d99-6d0f-44d7-a8a5-925b012e5940/v2.0
+	// -> 8de88d99-6d0f-44d7-a8a5-925b012e5940
+	tenantID := strings.Split(strings.TrimPrefix(issuerURL, "https://login.microsoftonline.com/"), "/")[0]
+	nameParts := strings.Split(username, "@")
+	if len(nameParts) != 2 {
+		return nil, fmt.Errorf("invalid username format: %s, expected format is 'username@domain'", username)
+	}
+	domain := nameParts[1]
```
3v1n0, line 468: since it's the standard library, maybe `mail.ParseAddress` too?

.github/workflows/ci.yml
```diff
@@ -13,6 +13,11 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v5
+
+      - name: Build libhimmelblau
+        # The code sanity check fails if himmelblau.h does not exist, so we generate it first.
+        run: go generate ./internal/providers/msentraid/...
```
3v1n0, line 21: Maybe move this commit earlier, when the library is added (and before we use it in code)

internal/providers/msentraid/himmelblau.go
```diff
@@ -34,23 +35,49 @@ var (
...
+		filter := "warn"
+		logLevel := log.GetLevel()
+		if logLevel <= log.DebugLevel {
+			log.Debug(context.Background(), "Setting libhimmelblau tracing level to DEBUG")
+			filter += ",himmelblau=debug"
+		} else if logLevel <= log.InfoLevel {
+			filter += ",himmelblau=info"
+		}
+
+		msalErr := C.set_module_tracing_filter(C.CString(filter))
```
3v1n0: Does this allows us to drop the `RUST_LOG` thingy?

internal/providers/msentraid/himmelblau.go
```diff
@@ -34,23 +35,49 @@ var (
...
+		// An optional TPM Transmission Interface. If this parameter is NULL, a soft TPM is initialized.
+		var tctiName *C.char
+		msalErr = C.tpm_init(tctiName, &tpm)
+		if msalErr != nil {
+			err = fmt.Errorf("failed to initialize TPM: %v", C.GoString(msalErr.msg))
+			return
+		}
+	})
+
+	return err
```
3v1n0, line 49: Same here, we may return `nil` after the first time

internal/broker/broker.go
```diff
@@ -710,10 +712,9 @@ func (b *Broker) handleIsAuthenticated(ctx context.Context, session *session, au
 			// One possible reason is that the device was deleted by an administrator in Entra ID.
 			// In this case, the user can perform device authentication again to get a new token
 			// and register the device again, allowing the user to log in.
-			msg := "Authentication failed due to a token issue. Please try again using device authentication."
 			// TODO: Check what happens if we're in offline mode
 			session.nextAuthModes = []string{authmodes.Device, authmodes.DeviceQr}
-			return AuthNext, errorMessage{Message: msg}
+			return AuthNext, errorMessage{Message: "Authentication failed due to a token issue. Please try again using device authentication."}
```
3v1n0, line 765: multiline?

internal/providers/msentraid/generate-himmelblau.sh
```diff
@@ -17,13 +17,21 @@ cargo install cargo-c cbindgen
...
+# Print executed commands to ease debugging
+set -x
+
 "${CARGO_HOME:-$HOME/.cargo}"/bin/cbindgen --config ./cbindgen.toml > himmelblau/himmelblau.h
-cargo cbuild --release --lib --features=broker,changepassword,on_behalf_of
+
+FEATURES="broker,changepassword,on_behalf_of"
+if [ -n "${ENABLE_TEST_FEATURES:-}" ]; then
+  FEATURES="${FEATURES},custom_oidc_discovery_url"
+fi
```
3v1n0, line 30: If you want to use bash arrays I also like them :)

internal/providers/msentraid/himmelblau.go
```diff
@@ -33,6 +33,8 @@ var (
 
 	// ErrDeviceDisabled is returned when the device is disabled in Microsoft Entra ID.
 	ErrDeviceDisabled = fmt.Errorf("device is disabled in Microsoft Entra ID")
+
+	authorityBaseURL = "https://login.microsoftonline.com"
```
3v1n0: Ok, ignore my previous comment then (I'm going through commits...)

internal/providers/msentraid/himmelblau.go
```diff
@@ -81,7 +83,7 @@ func ensureBrokerClientAppInitialized(tenantID string, data *deviceRegistrationD
 	brokerClientAppInitOnce.Do(func() {
 		var msalErr *C.MSAL_ERROR
 
-		authority := C.CString("https://login.microsoftonline.com/" + tenantID)
+		authority := C.CString(authorityBaseURL + "/" + tenantID)
```
3v1n0: `fmt.Sprintf` is clearer in these cases IMHO.

internal/providers/msentraid/msentraid_test.go
```diff
@@ -205,6 +207,35 @@ func TestIsTokenForDeviceRegistration(t *testing.T) {
...
+	accessTokenStr, err := accessToken.SignedString(testutils.MockKey)
+	require.NoError(t, err, "Failed to sign access token")
+	token := &oauth2.Token{
+		AccessToken:  accessTokenStr,
+		RefreshToken: "refreshtoken",
+		Expiry:       time.Now().Add(1000 * time.Hour),
+	}
+
+	tenantID := "8de88d99-6d0f-44d7-a8a5-925b012e5940"
+	issuerURL := fmt.Sprintf("https://login.microsoftonline.com/%s/v2.0", tenantID)
```
3v1n0, line 254: Use defined value?

internal/providers/msentraid/msentraid_test.go
```diff
@@ -205,6 +207,35 @@ func TestIsTokenForDeviceRegistration(t *testing.T) {
...
+
+	mockMS, err := newMockMS(tenantID)
+	require.NoError(t, err, "Failed to start mock MS server")
+	t.Cleanup(mockMS.Close)
+
+	msentraid.SetAuthorityBaseURL(mockMS.Server.URL)
+	err = os.Setenv("HIMMELBLAU_DISCOVERY_URL", mockMS.Server.URL)
```
3v1n0: This makes the test `not parallel`.

internal/providers/msentraid/msentraid_test.go
```diff
@@ -205,6 +207,35 @@ func TestIsTokenForDeviceRegistration(t *testing.T) {
 	}
 }
 
+func TestMaybeRegisterDevice(t *testing.T) {
+	t.Parallel()
```
3v1n0, line 229: Drop this as it's acting on global things

internal/providers/msentraid/msmock_test.go
```diff
@@ -0,0 +1,180 @@
...
+				"oid":  "00000000-0000-0000-0000-000000000000",
+				"tid":  "11111111-1111-1111-1111-111111111111",
+			},
+			"client_info": map[string]string{
+				"uid":  "11111111-1111-1111-1111-111111111111",
+				"utid": "22222222-2222-2222-2222-222222222222",
+			},
+		}
+		w.Header().Set("Content-Type", "application/json")
+		_ = json.NewEncoder(w).Encode(resp)
```
3v1n0: Maybe just panic on error

internal/providers/msentraid/msmock_test.go
```diff
@@ -0,0 +1,180 @@
...
+
+	case "srv_challenge":
+		fmt.Fprintf(os.Stderr, "Mock MS responding with nonce\n")
```
3v1n0: `Fprint`

internal/broker/broker.go
```diff
@@ -698,6 +712,21 @@ func (b *Broker) handleIsAuthenticated(ctx context.Context, session *session, au
 			}
 		}
 
+		// If device registration is enabled, ensure that the device is registered.
+		if b.provider.SupportsDeviceRegistration() && b.cfg.registerDevice {
+			authInfo.DeviceRegistrationData, err = b.provider.MaybeRegisterDevice(ctx, authInfo.Token, session.username, b.cfg.issuerURL, authInfo.DeviceRegistrationData)
```
3v1n0, line 729: looong line :)

internal/broker/broker.go
```diff
@@ -310,6 +310,12 @@ func (b *Broker) authModeIsAvailable(session session, authMode string) bool {
 			return true
 		}
 
+		if session.isOffline {
+			// If the session is in offline mode, we can't register the device anyway,
+			// so we can allow the user to use local password authentication.
+			return true
```
3v1n0, line 316: This does not imply that the user can access if the force-online option has been set, right?

internal/broker/broker.go
```diff
@@ -725,7 +725,7 @@ func (b *Broker) handleIsAuthenticated(ctx context.Context, session *session, au
 		}
 
 		// If device registration is enabled, ensure that the device is registered.
-		if b.provider.SupportsDeviceRegistration() && b.cfg.registerDevice {
+		if b.provider.SupportsDeviceRegistration() && !session.isOffline && b.cfg.registerDevice {
```
3v1n0, line 728: Shouldn't we return with an error and handle it in a specific way here?

internal/providers/msentraid/himmelblau_c.go
3v1n0, line 1: Ok, nicer! :)

internal/providers/msentraid/himmelblau_c.go
```diff
@@ -0,0 +1,330 @@
...
+	return key, cleanup, nil
+}
+
+func enrollDevice(broker *brokerClientApplication, refreshToken string, attrs *C.EnrollAttrs, tpm *boxedDynTPM, machineKey *C.MachineKey) (data *deviceRegistrationData, err error) {
+	cRefreshToken := C.CString(refreshToken)
+	defer C.free(unsafe.Pointer(cRefreshToken))
+
+	var cTransportKey *C.LoadableMsOapxbcRsaKey
+	var cCertKey *C.LoadableMsDeviceEnrolmentKey
+	var cDeviceID *C.char
```
3v1n0, line 180: Needs to be free'd, `loadable_ms_oapxbc_rsa_key_free`, `loadable_ms_device_enrollment_key_free` and `free` I guess

internal/providers/msentraid/msentraid_test.go
```diff
@@ -208,8 +219,28 @@ func TestIsTokenForDeviceRegistration(t *testing.T) {
 }
 
 func TestMaybeRegisterDevice(t *testing.T) {
+	// Registering a device is not thread-safe, because libhimmelblau sets BrokerClientApplication.cert_key
+	// after successful device registration, and uses the same field in subsequent calls,
+	// e.g. in the acquire_token_by_refresh_token method. If the cert_key does not match the expected value,
+	// libhimmelblau returns "TPM error: Failed to load IdentityKey: Aes256GcmDecrypt".
+	mockMSServerRegisterDeviceMu.Lock()
```
3v1n0, line 226: I think we should lock in the himberblau side as mentioned earlier

internal/providers/msentraid/himmelblau.go
```diff
@@ -0,0 +1,342 @@
...
+	if err != nil {
+		panic(fmt.Sprintf("failed to initialize TPM: %v", C.GoString(err.msg)))
+	}
+}
+
+func ensureBrokerClientAppInitialized(tenantID string, data *deviceRegistrationData) (err error) {
+	brokerClientAppInitOnce.Do(func() {
+		var msalErr *C.MSAL_ERROR
+
+		authority := C.CString("https://login.microsoftonline.com/" + tenantID)
```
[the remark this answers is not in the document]
adombeck: already done in later commit

internal/providers/msentraid/himmelblau.go
```diff
@@ -0,0 +1,342 @@
...
+				(*C.uint8_t)(unsafe.Pointer(&data.CertKey[0])),
+				C.size_t(len(data.CertKey)),
+				&cCertKey,
+			)
+			if msalErr != nil {
+				err = fmt.Errorf("failed to deserialize device enrollment key: %v", C.GoString(msalErr.msg))
+				return
+			}
+		} else {
+			cCertKey = nil // No cert key provided, will be generated during enrollment.
```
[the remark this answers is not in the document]
adombeck: obsolete

internal/providers/msentraid/himmelblau.go
```diff
@@ -0,0 +1,342 @@
...
+		}
+
+		var cTransportKey *C.LoadableMsOapxbcRsaKey
+		if data != nil && len(data.TransportKey) > 0 {
+			msalErr = C.deserialize_loadable_ms_oapxbc_rsa_key(
+				(*C.uint8_t)(unsafe.Pointer(&data.TransportKey[0])),
+				C.size_t(len(data.TransportKey)),
+				&cTransportKey,
```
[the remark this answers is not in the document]
adombeck: done

internal/providers/msentraid/himmelblau.go
```diff
@@ -0,0 +1,342 @@
...
+
+// registerDevice registers the device with Microsoft Entra ID.
+// It returns the device registration data as JSON.
+func (p *Provider) registerDevice(ctx context.Context, token *oauth2.Token, tenantID, domain string) ([]byte, error) {
+	if err := ensureBrokerClientAppInitialized(tenantID, nil); err != nil {
+		return nil, fmt.Errorf("failed to initialize broker client application: %v", err)
+	}
+
+	var cAuthValue *C.char
+	msalErr := C.auth_value_generate(&cAuthValue)
```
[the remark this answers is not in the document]
adombeck: already fixed in a later commit

internal/providers/msentraid/himmelblau.go
```diff
@@ -0,0 +1,342 @@
...
+	if msalErr != nil {
+		return nil, fmt.Errorf("failed to generate auth value: %v", C.GoString(msalErr.msg))
+	}
+
+	var loadableMachineKey *C.LoadableMachineKey
+	msalErr = C.tpm_machine_key_create(tpm, cAuthValue, &loadableMachineKey)
```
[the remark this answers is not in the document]
adombeck: already fixed in a later commit

internal/providers/msentraid/himmelblau.go
```diff
@@ -0,0 +1,342 @@
...
+		nil, /* device_type - default is "Linux" */
+		0,   /* join_type - 0: Azure AD join */
+		cOSVersion,
+		&attrs,
+	)
+	if msalErr != nil {
+		return nil, fmt.Errorf("failed to initialize enroll attributes: %v", C.GoString(msalErr.msg))
+	}
+
+	var machineKey *C.MachineKey
```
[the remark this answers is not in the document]
adombeck: already fixed in a later commit

internal/providers/msentraid/himmelblau.go
```diff
@@ -0,0 +1,342 @@
...
+		return "unknown"
+	}
+
+	for _, line := range strings.Split(string(data), "\n") {
+		if name, found := strings.CutPrefix(line, "PRETTY_NAME="); found && name != "" {
+			return name
+		}
+	}
+
+	log.Warningf(context.Background(), "PRETTY_NAME not found in /etc/os-release")
```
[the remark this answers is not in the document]
adombeck: yeah I don't think we need to handle the case that `PRETTY_NAME` is not found better because that's already very unusual, right?

internal/providers/msentraid/himmelblau.go
```diff
@@ -0,0 +1,342 @@
...
+
+	if err := ensureBrokerClientAppInitialized(tenantID, &data); err != nil {
+		return "", fmt.Errorf("failed to initialize broker client application: %v", err)
+	}
+
+	var loadableMachineKey *C.LoadableMachineKey
+	msalErr := C.deserialize_loadable_machine_key(
+		(*C.uint8_t)(unsafe.Pointer(&data.TPMMachineKey[0])),
+		C.size_t(len(data.TPMMachineKey)),
+		&loadableMachineKey,
```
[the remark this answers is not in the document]
adombeck: already fixed in a later commit

internal/providers/msentraid/msentraid.go
```diff
@@ -26,6 +30,12 @@ import (
 
 func init() {
 	pp.ColoringEnabled = false
+
+	// Enable debug logging for libhimmelblau
+	err := os.Setenv("RUST_LOG", "debug")
+	if err != nil {
+		log.Warning(context.Background(), fmt.Sprintf("Failed to set RUST_LOG: %v", err))
```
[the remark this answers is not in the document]
adombeck: done

.github/workflows/ci.yml
```diff
@@ -29,6 +29,10 @@ jobs:
   go-tests:
     name: "Go: Tests"
     runs-on: ubuntu-24.04 # ubuntu-latest-runner
+    strategy:
+      fail-fast: false
+      matrix:
+        test: [ "coverage","asan" ]
```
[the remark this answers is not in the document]
adombeck: done

internal/providers/msentraid/himmelblau.go
adombeck: uh I disagree, I find `authorityBaseURL + "/" + tenantID` easier to read and more idiomatic than `fmt.Sprintf("%s/%s", authorityBaseURL, tenantID)`. I also assume that a simple string concatenation is more efficient than string interpolation.

maybe we can avoid arguing about the two approaches and instead use `url.JoinPath(authorityBaseURL, tenantID)`

internal/broker/broker.go
adombeck: done

internal/broker/broker.go
adombeck: no, I think we want to allow offline login even if device registration is enabled

internal/providers/msentraid/himmelblau_c.go
adombeck: fixed

internal/providers/msentraid/himmelblau.go
```diff
@@ -0,0 +1,342 @@
+package msentraid
```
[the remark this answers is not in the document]
adombeck: I was also considering that but decided against it because then consequent next step would be to also add unit tests for the functions exported by that package, which would be `RegisterDevice` and `AcquireAccessTokenForGraphAPI`. Those two are already tested by `TestMaybeRegisterDevice` and `TestGetUserInfo` in the `msentraid` package, so new tests would only duplicate code and require that the test helpers like `mockMSServer` can be imported by multiple packages, so they would also need to go into a separate package. All that seems like a lot of overhead for little gain.

I would be fine with extracting a sub package if we can agree that we *don't* add additional tests for that package.

internal/providers/msentraid/himmelblau.go
adombeck: indeed, I forgot to drop `RUST_LOG`. did that now. thanks!

internal/providers/msentraid/msentraid.go
```diff
@@ -36,62 +46,66 @@ const (
...
+	skipAccessTokenForGraphAPI bool
+	// Used as the token scopes of the access token for the Microsoft Graph API in tests.
+	tokenScopesForGraphAPI []string
 }
 
 // New returns a new MSEntraID provider.
-func New() Provider {
-	return Provider{
+func New() *Provider {
+	return &Provider{
```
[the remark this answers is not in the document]
adombeck: > These changes would have been nicer in another commit to explain the reasoning.

agreed. I can extract it into a separate commit if we agree that we want this change. I tried to figure out why I did made that change and I think it's because of these new methods which modify fields of the provider, so they need to use a pointer receiver:

```go
func (p *Provider) SetNeedsAccessTokenForGraphAPI(value bool) {
	p.needsAccessTokenForGraphAPI = value
}

// SetTokenScopesForGraphAPI can be used in tests to set the scopes for the Microsoft Graph API access token.
func (p *Provider) SetTokenScopesForGraphAPI(scopes []string) {
	p.tokenScopesForGraphAPI = scopes
}
```

> However, this opens the room for potential races, so we should see if it may require some locking in some cases

how is it affected by races?

internal/providers/msentraid/msentraid.go
```diff
@@ -268,12 +311,19 @@ func removeNonSecurityGroups(groups []msgraphmodels.Groupable) []msgraphmodels.G
...
+			} else if uniqueName := group.GetUniqueName(); uniqueName != nil {
+				s = *uniqueName
+			}
+			if s == "" {
 				log.Debugf(context.Background(), "Removing unnamed non-security group")
-				continue
+			} else {
+				log.Debugf(context.Background(), "Removing non-security group %s", s)
 			}
-			log.Debugf(context.Background(), "Removing non-security group %s", *groupNamePtr)
```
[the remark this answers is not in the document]
adombeck: done. I could also extract it into a separate PR if you prefer, because it seems completely unrelated

internal/providers/msentraid/msentraid.go
adombeck: the `Address` struct returned by `mail.ParseAddress` doesn't even expose the domain field, so what would we gain by using it?

.github/workflows/ci.yml
adombeck: done

internal/providers/msentraid/himmelblau.go
```diff
@@ -0,0 +1,342 @@
...
+		0,   /* join_type - 0: Azure AD join */
+		cOSVersion,
+		&attrs,
+	)
+	if msalErr != nil {
+		return nil, fmt.Errorf("failed to initialize enroll attributes: %v", C.GoString(msalErr.msg))
+	}
+
+	var machineKey *C.MachineKey
+	msalErr = C.tpm_machine_key_load(tpm, cAuthValue, loadableMachineKey, &machineKey)
```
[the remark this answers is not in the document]
adombeck: The only function I found to modify state is `broker_enroll_device` which sets [`BrokerClientApplication.transport_key`](https://gitlab.com/samba-team/libhimmelblau/blob/c3ce26da9b5a729d63f0d332b621cd31f482fad9/src/auth.rs#L3983-L3983) and [`BrokerClientApplication.cert_key`](https://gitlab.com/samba-team/libhimmelblau/blob/c3ce26da9b5a729d63f0d332b621cd31f482fad9/src/auth.rs#L4030-L4030). Holding a lock in `registerDevice`  doesn't help with that, because those keys are used by `user_token_access_token` which we call in the separate `accessTokenFromUserToken` method. So the caller must ensure that `registerDevice` is not called again before it called `accessTokenFromUserToken`. However, I think we're safe because we ensure that only one authentication session can run at a time, and in the tests I'm using locks to ensure that only one test which uses device registration runs at a time.

internal/broker/broker.go
```diff
@@ -182,12 +188,17 @@ func (b *Broker) NewSession(username, lang, mode string) (sessionID, encryptionK
 		s.isOffline = true
 	}
 
+	scopes := append(consts.DefaultScopes, b.provider.AdditionalScopes()...)
+	if b.provider.SupportsDeviceRegistration() && b.cfg.registerDevice {
+		scopes = consts.MicrosoftBrokerAppScopes
+	}
```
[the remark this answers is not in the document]
adombeck: That's part of the TODOs in the PR description. Something for a follow-up though, since it will be too much work to land this cycle.

conf/broker.conf
```diff
@@ -19,6 +19,11 @@ client_id = <CLIENT_ID>
 ## if the identity provider is unreachable (e.g. due to network issues).
 #force_provider_authentication = false
 
+## If enabled, automatically registers the device with Microsoft Entra
+## upon user login.
+## XXX: Should this go into a separate section?
+#register_device = true
```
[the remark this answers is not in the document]
adombeck: done in https://github.com/ubuntu/authd-oidc-brokers/pull/586/commits/f82b0af7f93d16e4f092dbb216e8294397f79234 (that's on PR https://github.com/ubuntu/authd-oidc-brokers/pull/586 which targets the `msentraid` branch, so that the setting is not added to the `broker.conf` of other brokers).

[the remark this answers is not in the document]
adombeck: On the other hand, we already use `msentraid` in the snap name, which is part of the path to the broker config file (`/var/snap/authd-msentraid/current/broker.conf`). So the user is already exposed to the "msentraid" string, which I still find suboptimal, but I would be fine with also using it in the config file itself. What do you think, @3v1n0?

internal/providers/msentraid/msentraid.go
3v1n0: It's to ensure the value *is* an email, I also was wondering if it could provide us more, but at least it's a formal check on the format we expect.

internal/providers/msentraid/himmelblau.go
3v1n0: I would have probably used `url.JoinPath(authorityBaseURL, tenantID)` :)

internal/providers/msentraid/himmelblau.go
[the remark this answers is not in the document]
adombeck: > but I also think we should ensure it the himmelblau code, and not rely on the upper layer of the stack to do the right thing

so should I implement the cleanup function to unlock the mutex after calling `accessTokenFromUserToken` or do you see another way?

internal/providers/msentraid/himmelblau.go
```diff
@@ -0,0 +1,188 @@
+package msentraid
+
+//go:generate ./generate-himmelblau.sh
+
+/*
+#cgo LDFLAGS: -L${SRCDIR} -Wl,-rpath=${SRCDIR} -lhimmelblau
+#include "himmelblau.h"
```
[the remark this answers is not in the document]
3v1n0: Yeah, I was wondering that too... Not a big deal, it avoids having `LD_LIBRARY_PATH`'s around, so fair enough for testing purposes.

internal/providers/msentraid/himmelblau.go
```diff
@@ -0,0 +1,342 @@
...
+		&loadableMachineKey,
+	)
+	if msalErr != nil {
+		return "", fmt.Errorf("failed to deserialize TPM machine key: %v", C.GoString(msalErr.msg))
+	}
+
+	var machineKey *C.MachineKey
+	cAuthValue := C.CString(data.AuthValue)
+	defer C.free(unsafe.Pointer(cAuthValue))
+	msalErr = C.tpm_machine_key_load(tpm, cAuthValue, loadableMachineKey, &machineKey)
```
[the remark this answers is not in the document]
adombeck: >  isn't tpm         *boxedDynTPM a shared variable?

yes, but `tpm_machine_key_load` doesn't write to it AFAICT and we ensure that it's initialized before it's used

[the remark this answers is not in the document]
adombeck: > tpm_machine_key_load doesn't write to it AFAICT 

mmh maybe it could write to it, at least it's passed as a mutable in the Rust code: https://gitlab.com/samba-team/libhimmelblau/blob/2bd265ed8fb38b79955d01f8fb987d38291ac319/src/capi.rs#L360-L360

internal/providers/msentraid/msentraid_test.go
```diff
@@ -177,3 +167,108 @@ func TestGetUserInfo(t *testing.T) {
...
+	defer mockMSServerRegisterDeviceMu.Unlock()
+
+	t.Parallel()
+
+	maybeRegisterDevice(t)
```
3v1n0, line 231: It would be nice if you could add some concurrent tests here, since I feel this is not playing well in such case

3v1n0: Or well, once the mock is setup, try to use the machinery to register multiple devices in parallel and see that we're not failing

internal/providers/msentraid/himmelblau.go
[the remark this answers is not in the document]
3v1n0: Yeah, we don't know but I see it as an implementation detail, in the sense that we should assume that none of these functions acting on a state (and that are not clear getters) may potentially have a state that we should not break through threads.

Then @dmulder may tell us better, but I'd consider that each registration should be protected for the whole flow when acting on potential per-library data.

internal/providers/msentraid/himmelblau.go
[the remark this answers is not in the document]
adombeck: > Well, or we have a global mutex here

the caller of `registerDevice` would have to hold that mutex. a cleanup function is be better for that, because it makes it clear to the caller that they need to call it.

> or we have per registration mutex that has to take care of potential races, also the ones that are in himmelblau

that doesn't work because libhimmelblau's `broker_enroll_device` sets global state (`BrokerClientApplication.transport_key` and `BrokerClientApplication.cert_key`) as explained above

internal/providers/msentraid/himmelblau.go
adombeck: done

conf/broker.conf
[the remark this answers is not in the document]
adombeck: done

internal/providers/msentraid/msentraid.go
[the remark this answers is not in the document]
adombeck: the only methods which write fields of the `Provider` struct are the two methods above. those are indeed prone to races, but the tests which call them make sure that they are not called in parallel. Their only usage is [here](https://github.com/ubuntu/authd-oidc-brokers/blob/d853622356cff0f44930331acf988eb51f427be0/internal/providers/msentraid/msentraid_test.go#L155-L157), and there `p` is only used in the same goroutine which initializes it.

internal/providers/msentraid/himmelblau.go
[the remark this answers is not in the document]
adombeck: done

internal/providers/msentraid/himmelblau.go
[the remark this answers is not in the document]
adombeck: > what we could do is to make registerDevice (and MaybeRegisterDevice) return a cleanup function to unlock a mutex, which the caller can use after calling accessTokenFromUserToken (or once it knows that it won't need that function).

done

adombeck:
@3v1n0 I fixed the issue I mentioned with by splitting `GetUserInfo` and `GetGroups` (0fad402ed726e896d8c523bc83293ff18cb8bf49)

adombeck:
I did everything I wanted to do before merging, so from my side we could merge this now.

3v1n0 (approved):
See if you can factorize some code in `MaybeRegisterDevice` calls, otherwise LGTM.

Thanks a lot for handling this!

PR #586, opened 2025-07-14
msentraid: Build libhimmelblau during snap build

adombeck:
This is the portion of ubuntu/authd#585 which should target the `msentraid` branch, because we don't want to build libhimmelblau for other snaps than the `authd-msentraid-broker` snap.

### DIFF
--- a/conf/broker.conf
+++ b/conf/broker.conf
@@ -15,6 +15,16 @@ client_id = <CLIENT_ID>
 ## if the identity provider is unreachable (e.g. due to network issues).
 #force_provider_authentication = false
 
+[msentraid]
+## Enable automatic device registration with Microsoft Entra ID
+## when a user logs in through this broker.
+##
+## If set to true, authd will attempt to register the local machine
+## as a device in Entra ID upon successful login.
+##
+## If set to false (the default), device registration will be skipped.
+#register_device = false
+
 [users]
 ## The directory where the home directories of new users are created.
 ## Existing users will keep their current home directory.


snap/snapcraft.yaml
```diff
@@ -47,21 +47,40 @@ slots:
...
+    build-packages:
+      - libssl-dev
+    override-build: |      
+      cargo install cargo-c
+      cargo cbuild --release --lib --features=broker,changepassword,on_behalf_of
+      echo "Installing libhimmelblau.so and himmelblau.h"       
+      TARGET_TRIPLE=$(rustc -vV | awk '/host:/ {print $2}') 
+      install -D target/${TARGET_TRIPLE}/release/libhimmelblau.so ${CRAFT_PART_INSTALL}/lib/libhimmelblau.so
+      ln -s libhimmelblau.so ${CRAFT_PART_INSTALL}/lib/libhimmelblau.so.0
+      install -D target/${TARGET_TRIPLE}/release/himmelblau.h ${CRAFT_PART_INSTALL}/include/himmelblau.h
```
3v1n0, line 62: This should be also excluded from the final snap, once compilation is done.

snap/snapcraft.yaml
```diff
@@ -47,21 +47,40 @@ slots:
...
       - go
+    after: 
+      - libhimmelblau
+      - version
     override-build: |
-      go mod download all
-      go mod vendor
+      # Add libhimmelblau to the CFLAGS and LDFLAGS so that it can be found by the Go linker.
+      export CGO_CFLAGS="-I${CRAFT_STAGE}/include"
+      export CGO_LDFLAGS="-L${CRAFT_STAGE}/lib -lhimmelblau"
```
3v1n0, line 76: use `build-environment` instead?

snap/snapcraft.yaml
```diff
@@ -47,21 +47,40 @@ slots:
...
+    plugin: rust
+    build-packages:
+      - libssl-dev
+    override-build: |      
+      cargo install cargo-c
+      cargo cbuild --release --lib --features=broker,changepassword,on_behalf_of
+      echo "Installing libhimmelblau.so and himmelblau.h"       
+      TARGET_TRIPLE=$(rustc -vV | awk '/host:/ {print $2}') 
+      install -D target/${TARGET_TRIPLE}/release/libhimmelblau.so ${CRAFT_PART_INSTALL}/lib/libhimmelblau.so
+      ln -s libhimmelblau.so ${CRAFT_PART_INSTALL}/lib/libhimmelblau.so.0
```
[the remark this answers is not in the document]
adombeck: > Normally the .so only file is a development file, and that is symlinked to the so-versioned one, so we should do the same IMHO

done

> and likely exclude the .so file from the final release once compilation is done

we need the .so, it's linked dynamically

snap/snapcraft.yaml
adombeck: done

snap/snapcraft.yaml
adombeck: I tried to do that via

```diff
diff --git a/snap/snapcraft.yaml b/snap/snapcraft.yaml
index f92c2830..0829b000 100644
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -81,6 +81,9 @@ parts:
       go mod download all
       go mod vendor
       go build -tags=withmsentraid -o ${GOBIN}/authd-msentraid ./cmd/authd-oidc
+
+      # Remove the libhimmelblau header, we don't need it anymore
+      rm ${CRAFT_STAGE}/include/himmelblau.h
   config:
     source: conf/
     source-type: local
```

However, that caused the build to fail with:
```
Failed to copy '/root/stage/include/himmelblau.h': no such file or directory.
```

Seems like snapcraft tries to copy all files which were copied into `$CRAFT_PART_INSTALL` in a previous stage (?) I don't feel like spending more time on this, since the size of the header is negligible.

snap/snapcraft.yaml
3v1n0, line 39: Not related to this, but just noticed I think we should now use [`type: dbus`](https://snapcraft.io/docs/services-and-daemons#daemons-and-d-bus) so that we have proper dbus-activation

snap/snapcraft.yaml
3v1n0: Ok, can also use the wildcard so that we don't have to maintain this, in case we add some other thing, as we don't care on include files in any case

snap/snapcraft.yaml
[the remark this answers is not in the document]
3v1n0: Sure, but we should need the `.so.0`, that is the actual file, while the `.so` should be a symlink that is used by the linker to figure out what is the so-version it's linked to (when you do `-lfoo` it can just look for `libfoo.so` and then internally links on `libfoo.so.X` where `X` is the target versioned library).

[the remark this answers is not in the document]
adombeck: ah, so with "likely exclude the `.so` file" you meant we should exclude the symlink? that makes more sense. I don't really see why we should care about that though, since a symlink has a file size of a few bytes.

PR #602, opened 2025-07-22
refactor: Split session.authInfo map

adombeck:
Use typed fields in the session struct instead of a map of type `map[string]any`.

Benefits:
* Compile-time type safety
* Better IDE support
* Avoids multiple uses of the term "auth info" for different things
(`session.authInfo["authInfo"]`)
* Also allows us to get rid of some tests which tested if the map
values have invalid types.

### DIFF
--- a/internal/broker/broker.go
+++ b/internal/broker/broker.go
@@ -575,6 +575,8 @@ func (b *Broker) handleIsAuthenticated(ctx context.Context, session *session, au
 			return AuthDenied, errorMessage{Message: "user not allowed in broker configuration"}
 		}
 
+		// Store the auth info in the session so that we can use it when handling the
+		// next IsAuthenticated call for the new password mode.
 		session.authInfo = authInfo
 		session.nextAuthModes = []string{authmodes.NewPassword}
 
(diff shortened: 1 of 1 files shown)


internal/broker/broker.go
```diff
@@ -70,12 +70,15 @@ type session struct {
 
 	oidcServer   *oidc.Provider
 	oauth2Config oauth2.Config
-	authInfo     map[string]any
```
denisonbarbosa, line 73: The idea behind this was to keep the session as dynamic as possible without having to define a lot of other fields in the type. That was, of course, before knowing that we would only have Device Authentication available as a provider authentication mode.

internal/broker/broker.go
```diff
@@ -70,12 +70,15 @@ type session struct {
...
 	isOffline    bool
 	userDataDir  string
 	passwordPath string
 	tokenPath    string
 
+	// Data to pass from one request to another.
+	deviceAuthResponse *oauth2.DeviceAuthResponse
+	authInfo           *token.AuthCachedInfo
+
```
denisonbarbosa, line 81: But I can understand why these changes make the code more readable and workable, so I agree with them.

denisonbarbosa (approved):

internal/broker/broker.go
adombeck: we can still add more fields to the session if at some point we support other authentication modes than device auth

PR #609, opened 2025-08-01
Add fuzz test to cover config file handling

denisonbarbosa:
This PR adds a fuzz test that covers the process of parsing the broker configuration files. This helps us make sure we cover edge cases and increase the safety of our code. More details are available in each commit.

UDENG-6525


adombeck:
I'll try the copilot review on this :)

adombeck:
Changes look good to me. I'll let the fuzzer run for a few minutes to see for how long it keeps finding interesting inputs (i.e. inputs which increase the number of reached code paths).

PR #615, opened 2025-08-05
Fix for changing local password

denisonbarbosa:
After #325, there was a regression where we stopped returning the NewPassword authentication mode as a next step for passwd sessions (to change the local password once it's already defined). This fixes that by returning the expected next authentication mode (i.e. NewPassword).

### DIFF
--- a/internal/broker/broker.go
+++ b/internal/broker/broker.go
@@ -605,6 +605,7 @@ func (b *Broker) handleIsAuthenticated(ctx context.Context, session *session, au
 			// Store the auth info in the session so that we can use it when handling the
 			// next IsAuthenticated call for the new password mode.
 			session.authInfo = authInfo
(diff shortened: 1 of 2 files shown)


3v1n0:
Looks good, is there a simple way we can add some tests for the layouts? So that we can have some check even before hitting the future e2e tests?

denisonbarbosa:
> Looks good, is there a simple way we can add some tests for the layouts? So that we can have some check even before hitting the future e2e tests?

I'm pretty sure we have tests for that already, but they were not checking for the expected next authentication modes.

adombeck:
I was able to reproduce the issue by running `passwd` as an authd user. With the current version of the authd-msentraid broker from the edge channel, it keeps asking for the local password. With the version built from this branch, after entering the correct local password, it asks for the new password.

The code change looks good to me. Thanks a lot for catching and fixing the issue!

PR #639, opened 2025-08-13
Print a message when the refresh token is expired

adombeck:
I encountered the case that the refresh token in one of my VMs was expired and was puzzled for a while why the PAM module auto-selects device authentication after I entered my local password. Let's print a message which makes it clear to the user why device authentication is required.

UDENG-7629

### DIFF
--- a/internal/broker/broker.go
+++ b/internal/broker/broker.go
@@ -618,9 +618,9 @@ func (b *Broker) handleIsAuthenticated(ctx context.Context, session *session, au
 			authInfo, err = b.refreshToken(ctx, session.oauth2Config, authInfo)
 			var retrieveErr *oauth2.RetrieveError
 			if errors.As(err, &retrieveErr) && b.provider.IsTokenExpiredError(*retrieveErr) {
-				// The refresh token is expired, so the user needs to authenticate via OIDC again.
+				log.Noticef(context.Background(), "Refresh token expired for user %q, new device authentication required", session.username)
 				session.nextAuthModes = []string{authmodes.Device, authmodes.DeviceQr}
-				return AuthNext, nil
+				return AuthNext, errorMessage{Message: "refresh token expired, please authenticate again using device authentication"}
 			}
 			if err != nil {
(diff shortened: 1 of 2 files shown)


internal/broker/broker.go
```diff
@@ -618,7 +618,7 @@ func (b *Broker) handleIsAuthenticated(ctx context.Context, session *session, au
 			authInfo, err = b.refreshToken(ctx, session.oauth2Config, authInfo)
 			var retrieveErr *oauth2.RetrieveError
 			if errors.As(err, &retrieveErr) && b.provider.IsTokenExpiredError(*retrieveErr) {
-				// The refresh token is expired, so the user needs to authenticate via OIDC again.
+				log.Infof(context.Background(), "Refresh token expired for user %q, requesting new device authentication", session.username)
```
3v1n0, line 621: Since we now support messages also in next case, can't we show the user a message too?

adombeck:
We could also consider to check the refresh token validity *before* asking the user for the local password. That would improve UX a bit more.

adombeck:
@3v1n0 waiting for your review

internal/broker/broker.go
3v1n0: Oh, sorry I missed the comment... But yes, it works in that way indeed... We need to double-check if GDM we've in stable supports it, though before releasing.

So add to your TO-CHECK before broker release todo list, please :)

adombeck: I already tested it on 24.04. I'm not sure if it makes sense to add this to the manual tests we do before release:
1. To trigger it (without patching the broker), we need an expired refresh token, which I don't think we can acquire unless we have a 3 month old token (I don't).
2. I don't expect that this GDM feature breaks for a specific Ubuntu release if it worked for that release before (in other words: now that I tested 24.04, I don't expect it to break in 24.04, but maybe in 25.10 or 26.04), so IMO we should test it before Ubuntu releases, but not before each broker release.

Ideally, we should have an e2e test which tests this, but again this requires an expired token which I don't have :/

PR #659, opened 2025-09-04
Add E2E tests for the msentraid broker

denisonbarbosa:
We've been needing those for a while now. Although they're not fully stable and automated yet, we should add the tests here to follow up with the other tasks and to discuss implementation details. 

More details are explained in the individual commits.

UDENG-7424


e2e-tests/common/utils.resource
```diff
@@ -0,0 +1,123 @@
...
+    BuiltIn.Sleep    2
+    PlatformVideoInput.Match    ${IMAGES_DIR}/circle-of-friends.png    60
+
+
+Log Out
+    Run Command    gnome-session-quit --logout --no-prompt
+
+
+Open GNOME Terminal
+    Start Application    /usr/bin/gnome-terminal
```
3v1n0, line 31: I'd simplify this by just doing `Ctrl+Alt+t`

e2e-tests/common/utils.resource
```diff
@@ -0,0 +1,123 @@
...
+    PlatformVideoInput.Match Text    @ubuntu:~$    15
+    PlatformHid.Keys Combo    Super_L    Up
```
3v1n0, line 33: Add a comment to clarify that we're maximizing it.

I know I suggested this, but on a 2nd though I think it would be actually better to instead make it full-screen by using `F11`.

So we are sure that no other terminal window is bothering us, but this also implies waiting for the "Window title" instead that for the content of the window (as the content may be show by another terminal anyways).

e2e-tests/common/utils.resource
```diff
@@ -0,0 +1,123 @@
...
+
+
+Open GNOME Terminal In Sudo Mode
+    Open GNOME Terminal
+    Enter Sudo Mode In GNOME Terminal
+
+
+Enter Sudo Mode In GNOME Terminal
+    PlatformHid.Type String    sudo -s
```
3v1n0, line 42: Add a clear first, just in case.

e2e-tests/common/utils.resource
```diff
@@ -0,0 +1,123 @@
...
+    PlatformHid.Type String    ubuntu
+    PlatformHid.Keys Combo    Return
+    PlatformVideoInput.Match Text    root@ubuntu:/home/ubuntu#    15
+    PlatformHid.Type String    clear
+    PlatformHid.Keys Combo    Return
+    PlatformVideoInput.Match Text    root@ubuntu:/home/ubuntu#    15
+
+
+Close GNOME Terminal
+    PlatformHid.Keys Combo    Control_L    Shift_L    W
```
3v1n0, line 54: Maybe check that it's still the focused app first?
Otherwise it's a very generical "Hit Cltr+Shift+W" on app.

So maybe I'd just rely on using `Alt+F4` and rename the function "Close Focused Window"

e2e-tests/common/utils.resource
```diff
@@ -0,0 +1,123 @@
...
+
+
+Close GNOME Terminal In Sudo Mode
+    PlatformHid.Type String    exit
```
3v1n0, line 58: Ensure we're in sudo mode first maybe (so clear + match)

e2e-tests/common/utils.resource
```diff
@@ -0,0 +1,123 @@
...
+    PlatformHid.Keys Combo    Return
+    BuiltIn.Sleep    2    # hacky
+    PlatformVideoInput.Match Text    cmd-finished    600
+
+
+Left Button Click
+    BuiltIn.Sleep    0.5
+    Click LEFT Button
+    BuiltIn.Sleep    0.5
+    Move Pointer To (100, 100)
```
3v1n0, line 99: I'd just move at the latest pixel of the screen, a it's less common to have something there.

e2e-tests/msentraid/common/broker.resource
```diff
@@ -0,0 +1,388 @@
...
+
+Continue Log In With Remote User: Log In On Browser
+    [Arguments]    ${username}    ${domain}
+# Wait until the verification URL and login code are displayed
+    PlatformVideoInput.Match Text    https://microsoft.com/devicelogin
+    # Read the user code.
+    ${text} =    PlatformVideoInput.Read Text
+    ${user_code} =    StringUtils.FirstMatch    (https://)?microsoft.com/devicelogin\n((Login code: )?([A-Z0-9]+))    ${text}
+
+    Start Process    ${ENTRAID_COMMON}/BrowserWindow.py    ${username}@${domain}    %{E2E_PASSWORD}    ${user_code}    alias=RemoteAuth
```
3v1n0, line 85: Do we want to run it with `RUNNING_OFFSCREEN` or we prefer to see the UI to record it, just in case?

denisonbarbosa: Sometimes it bugs out and types the password in the incorrect field (I couldn't understand why exactly), but it's something that we definitely do not want to record

e2e-tests/common/utils.resource
denisonbarbosa: I'm not sure I like the full-screen approach, and it shouldn't make a difference compared to maximizing the terminal window, right?

e2e-tests/common/utils.resource
```diff
@@ -0,0 +1,123 @@
...
+*** Keywords ***
+Log In
+    PlatformVideoInput.Match Text    Not listed    120
+    PlatformHid.Keys Combo    Return
+    BuiltIn.Sleep    2
+    PlatformHid.Type String    ubuntu
+    BuiltIn.Sleep    2
+    PlatformHid.Keys Combo    Return
+    BuiltIn.Sleep    2
+    PlatformVideoInput.Match    ${IMAGES_DIR}/circle-of-friends.png    60
```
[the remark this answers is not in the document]
3v1n0: Well, indeed both can change, I do just feel that the text mapping is safer than image checks in yarf, for example if we change the background or the [yaru dock background](https://github.com/ubuntu/yaru/issues/4322) this matching will break, so until we've enough fuzzing in the yarf image matching, I'd prefer to reduce it as much as we can to be more future proof

e2e-tests/common/utils.resource
3v1n0: It does, because there's no other text involved a part the one in the terminal itself, and there's no risk that something in the panel or in some popping-up application may break us

e2e-tests/msentraid/common/BrowserWindow.py
```diff
@@ -0,0 +1,199 @@
...
+    browser.send_key_taps(ascii_string_to_key_events(username) + [Gdk.KEY_Return])
+    browser.wait_for_stable_page()
+
+    browser.send_key_taps(ascii_string_to_key_events(password) + [Gdk.KEY_Return])
+    browser.wait_for_stable_page()
+
+    browser.send_key_taps([Gdk.KEY_Return])
+    browser.wait_for_stable_page()
+
+    sleep(5)
```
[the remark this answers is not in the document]
3v1n0: Ok, that's acceptable... But... We should never sleep this way in UI applications, but rely instead on the main loop, so can do instead something like:

```python

def wait_for_timeout(timeout):
    loop = GLib.MainLoop()

    def on_timeout():
        loop.quit()
        return False

    GLib.timeout_add(timeout, on_timeout)
    loop.run()
```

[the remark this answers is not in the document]
denisonbarbosa: I feel like that is what the `wait_for_stable_page` function was supposed to be doing. Maybe we have a logic flaw somewhere?

e2e-tests/common/utils.resource
```diff
@@ -0,0 +1,139 @@
...
+#     # Sleep a bit to avoid matching the text on the command itself
+#     BuiltIn.Sleep    2    # hacky
+#     Match Text    cmd-finished    600
+Run Command In GNOME Terminal
+    [Arguments]    ${command}
+    Hid.Type String    ${command}
+    Hid.Type String   clear && ${command} && clear && echo cmd-finished
+    Hid.Keys Combo    Return
+    BuiltIn.Sleep    2
+    Match Text    cmd-finished    600
```
3v1n0, line 108: Isn't sleep already part of matching (600)? Why sleeping more?

e2e-tests/msentraid/common/broker.resource
```diff
@@ -0,0 +1,305 @@
...
+    [Arguments]    ${username}    ${local_password}
+    Hid.Type String    ssh ${username}@localhost
+    Hid.Keys Combo    Return
+    Match Text    Enter your local password:    30
+    Hid.Type String    ${local_password}
+    Hid.Keys Combo    Return
+    Match Text    ${username}@ubuntu:~$    30
+
+
+Log In With Remote User Through GDM: QR Code
```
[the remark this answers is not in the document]
3v1n0: However, isn't this common to all brokers?

e2e-tests/common/images/circle-of-friends.png
3v1n0, line 1: Is this used anywhere?

3v1n0 (commented):
Great stuff, I hope it's quite stable but we have a great base now

e2e-tests/msentraid/common/broker.resource
3v1n0: Let's keep this open, and see if I can handle it :)

e2e-tests/common/utils.resource
```diff
@@ -0,0 +1,139 @@
...
+#     Hid.Keys Combo    ${ampersand}
+#     Hid.Keys Combo    ${ampersand}
+#     Hid.Keys Combo    space
+#     Hid.Type String    echo cmd-finished
+#     # Sleep a bit to ensure everything gets typed properly
+#     BuiltIn.Sleep    2    # hacky
+#     Hid.Keys Combo    Return
+#     # Sleep a bit to avoid matching the text on the command itself
+#     BuiltIn.Sleep    2    # hacky
+#     Match Text    cmd-finished    600
```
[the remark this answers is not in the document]
denisonbarbosa: oops, forgot to remove it

e2e-tests/common/utils.resource
denisonbarbosa: Not really... We need the sleep to make sure we don't match the "cmd-finished" of the command itself

e2e-tests/msentraid/common/broker.resource
```diff
@@ -0,0 +1,305 @@
...
+Enable Edge Broker
+    Open GNOME Terminal In Sudo Mode
+    Run Command In GNOME Terminal    snap refresh authd-msentraid --edge
+    Close GNOME Terminal In Sudo Mode
+
+
+Disable Broker And Purge Config
+    Open GNOME Terminal In Sudo Mode
+    Run Command In GNOME Terminal    snap stop authd-msentraid
+    Run Command In GNOME Terminal    rm ${AUTHD_BROKER_CFG}
```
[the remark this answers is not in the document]
denisonbarbosa: The syntax is for the YARF variable; it isn't going through to bash

e2e-tests/msentraid/common/broker.resource
```diff
@@ -0,0 +1,305 @@
...
+
+Continue Log In With Remote User: Log In On Browser
+    [Arguments]    ${username}
+    # Wait until the verification URL and login code are displayed
+    Match Text    https://microsoft.com/devicelogin
+    # Read the user code.
+    ${text} =    Read Text
+    ${user_code} =    StringUtils.FirstMatch    (https://)?microsoft.com/devicelogin\n((Login code: )?([A-Z0-9]+))    ${text}
+
+    Start Process    ${ENTRAID_COMMON}/BrowserWindow.py    ${username}    %{E2E_PASSWORD}    ${user_code}    alias=RemoteAuth
```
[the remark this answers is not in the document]
denisonbarbosa: I would then change the behavior and have a `RUN_ONSCREEN` variable that we can set to 1 and show the GTK window. Wdyt?

e2e-tests/msentraid/common/broker.resource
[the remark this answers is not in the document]
denisonbarbosa: Kind of. The broker name on the list of brokers will change, and if we ever release brokers with different authentication modes, it will complicate things (if we move this to authd.resource), so I think it's better to keep it this way for now

e2e-tests/msentraid/common/broker.resource
```diff
@@ -0,0 +1,305 @@
...
+    Match Text    ${username}@ubuntu:~$    30
+
+
+Log In With Remote User Through GDM: QR Code
+    [Arguments]    ${username}    ${local_password}
+    Start Log In With Remote User Through GDM: QR Code    ${username}
+    Select Broker Through GDM
+    Continue Log In With Remote User: Log In On Browser    ${username}
+    Continue Log In With Remote User Through GDM: Define Local Password    ${local_password}
+
```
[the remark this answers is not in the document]
denisonbarbosa: Oops, my bad

e2e-tests/run_tests.sh
```diff
@@ -0,0 +1,88 @@
...
+    ${VM_DIR}/spawn-vm.sh ${VM_IMG} &
+    sleep 5  # Wait a bit for the VM to boot up
+    VM_PID=$(pidof kvm)
+    echo "VM PID: ${VM_PID}"
+
+    E2E_USER="$E2E_USER" \
+    E2E_PASSWORD="$E2E_PASSWORD" \
+    yarf --outdir "output/${test_name}" --platform=Vnc . || true
+
+    test_result=$?
```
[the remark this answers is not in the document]
denisonbarbosa: It is always true, indeed. We have `set -e`, so we can't have a command failing, otherwise the script will exit. And we don't want that at this point.

e2e-tests/run_tests.sh
```diff
@@ -0,0 +1,88 @@
...
+
+    if [ -n "${KEEP_VM:-}" ]; then
```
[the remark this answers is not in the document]
denisonbarbosa: Does it work even with -u?

e2e-tests/run_tests.sh
```diff
@@ -0,0 +1,88 @@
...
+        echo "Stopping the test run and keeping VM running as requested: ${VM_PID}."
+        exit ${test_result}
+    else
+        echo "Stopping VM: ${VM_PID}"
+        kill -KILL "${VM_PID}"
+    fi
+
+    sleep 5  # Wait for the VM to stop
```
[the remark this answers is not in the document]
denisonbarbosa: Oh, nice... I didn't know about that

e2e-tests/run_tests.sh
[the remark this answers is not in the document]
3v1n0: No sure, but with the original syntax, you won't ever be able to catch the test results, reason the suggested one, that does not fail, but it also keeps track of the return value of `yarf`

e2e-tests/msentraid/common/broker.resource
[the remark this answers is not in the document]
3v1n0: As you like it better, for hacking on it is definitely better to have it onscreen, but otherwise we can avoid that

e2e-tests/common/utils.resource
3v1n0: But didn't we clear it? I'd say we should do something `command & clear && wait && echo cmd-finished`, so extra sleep should not be needed

e2e-tests/common/utils.resource
```diff
@@ -0,0 +1,139 @@
...
+Run Command In GNOME Terminal
+    [Arguments]    ${command}
+    Hid.Type String    ${command}
+    Hid.Type String   clear && ${command} && clear && echo cmd-finished
```
3v1n0, line 105:
```suggestion
    Hid.Type String   clear && ${command} & clear && wait && echo cmd-finished
```

e2e-tests/common/utils.resource
```diff
@@ -0,0 +1,139 @@
...
+#     Hid.Keys Combo    ${ampersand}
+#     Hid.Keys Combo    space
+#     Hid.Type String    echo cmd-finished
+#     # Sleep a bit to ensure everything gets typed properly
+#     BuiltIn.Sleep    2    # hacky
+#     Hid.Keys Combo    Return
+#     # Sleep a bit to avoid matching the text on the command itself
+#     BuiltIn.Sleep    2    # hacky
+#     Match Text    cmd-finished    600
+Run Command In GNOME Terminal
```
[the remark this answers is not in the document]
denisonbarbosa: Good point :joy:

I'll rename it to just "Run Command in Terminal"

e2e-tests/common/utils.resource
3v1n0: I see, BTW please check with the new syntax and let's see if we can get such race, in case we can add the sleep back (even less than a a second though)

e2e-tests/requirements.md
```diff
@@ -0,0 +1,46 @@
+# Requirements for running the authd tests
+
+<!-- TODO: Add instructions on how to build YARF from source and install the required dependencies -->
+
+- 2 VMs with Ubuntu 24.04 installed;
+
+- For the latest versions VM:
+  - Install authd and the authd-msentraid broker from their respective PPA and snap edge channels:
+    - `sudo add-apt-repository ppa:ubuntu-enterprise-desktop/authd-edge`;
+    - `sudo apt update && sudo apt install authd`;
```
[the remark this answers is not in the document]
denisonbarbosa: This is just a temporary file (my bad, I should've marked it as such). I will add a more detailed and updated documentation after we get this done (now I'm using `virsh` and `cloud-init` to configure the VM)

e2e-tests/common/BrowserWindow.py
```diff
@@ -0,0 +1,147 @@
+#!/usr/bin/env python3
```
3v1n0: You don't need this anymore if it's a library.

Also, maybe let's call the file lower_case as it's more pythonic.

e2e-tests/authd-msentraid/common/browser_login.py
```diff
@@ -0,0 +1,61 @@
...
+            [
+                "/usr/bin/env",
+                "RUNNING_OFFSCREEN=1",
+                "GDK_BACKEND=x11",
+                "xvfb-run",
+                "-a",
+                sys.executable,
+            ]
+            + sys.argv,
+        )
```
3v1n0: Can this be moved to an utility function too?

e2e-tests/authd-msentraid/common/browser_login.py
```diff
@@ -0,0 +1,61 @@
...
+
+    if len(sys.argv) < 4:
+        print("Usage: BrowserWindow.py <username> <password> <code>")
+        sys.exit(1)
+
+    username = sys.argv[1]
+    password = sys.argv[2]
+    device_code = sys.argv[3]
```
3v1n0: Not a big deal, but I feel you can have this with 4 lines of [`argparse`](https://docs.python.org/3/library/argparse.html)

e2e-tests/vm/runner-cloud-cfg.yaml
```diff
@@ -0,0 +1,66 @@
...
+runcmd:
+  # Disable apport
+  - sed -i 's/enabled=1/enabled=0/' /etc/default/apport
+  - systemctl mask --now apport.service
+  # Disable automatic updates and remove some unnecessary packages
+  - apt-get remove -y update-manager gnome-initial-setup
+  # We don't need Firefox for the tests, so we can remove it to save some space
+  - snap remove firefox --purge
+  # Configure SSH to allow interactive authentication
+  - sed -i 's/KbdInteractiveAuthentication no/KbdInteractiveAuthentication yes/' /etc/ssh/sshd_config
```
3v1n0: Maybe can use the exact code that is in https://github.com/ubuntu/authd/blob/main/docs/reference/cloud-init-deploy.md ?

e2e-tests/common/BrowserWindow.py
```diff
@@ -148,48 +144,4 @@ def ascii_string_to_key_events(string):
...
-    browser.wait_for_stable_page()
-
-    browser.send_key_taps(ascii_string_to_key_events(password) + [Gdk.KEY_Return])
-    browser.wait_for_stable_page()
-
-    browser.send_key_taps([Gdk.KEY_Return])
-    browser.wait_for_stable_page()
-
-    sys.exit(0)
+    raise TimeoutError("Operation timed out")
```
3v1n0: nit: add EOF

e2e-tests/common/utils.resource
```diff
@@ -7,111 +7,92 @@ Library             Process
...
 
 Log Out
     Run Command    gnome-session-quit --logout --no-prompt
 
 
-Open GNOME Terminal
-    Hid.Keys Combo    Control_L    Alt_L    t
-    Match Text    @ubuntu:~$    15
+Open Terminal
+    Run Command    /usr/bin/x-terminal-emulator
```
3v1n0: Can drop `/usr/bin`  and maybe call `xdg-terminal`, since we are not tergetting noble anyways here

e2e-tests/testing.md
```diff
@@ -0,0 +1,386 @@
...
+    Those are all part of the archive and can be installed on Ubuntu with:
+
+    ```bash
+    sudo apt-get install -y {virtualization dependencies}
+    ```
+
+- Test-run dependencies:
+
+    ```text
+    clang libxkbcommon-dev libcairo2-dev libgirepository-2.0-dev tesseract-ocr python3-tk python3-gi python3-cairo xvfb ffmpeg libwebkit2gtk-4.1-0 libwebkit2gtk-4.1-dev libgtk-3-0t64 libgtk-3-dev
```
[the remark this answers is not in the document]
denisonbarbosa: `libgirepository-2.0-dev`, `python3-tk`, `python3-cairo` are required to build the `pygobject` pip module (and they are not available through pip anymore).

`gir1.2-webkit2-4.1`: Yes, we talked about it yesterday and I forgot to commit it, my bad :sweat_smile: 

I'll remove the gtk libraries

e2e-tests/common/utils.resource
denisonbarbosa: Even though we are not targeting Noble (yet, at least), I think it's better to keep the tests as release-independent as possible. If we change that, we would need to have multiple versions of the resource files (and although this might eventually become the case in the future, we should avoid it for now)

e2e-tests/vm/runner-cloud-cfg.yaml
```diff
@@ -0,0 +1,66 @@
...
+      idle-delay=uint32 0
+
+runcmd:
+  # Disable apport
+  - sed -i 's/enabled=1/enabled=0/' /etc/default/apport
+  - systemctl mask --now apport.service
+  # Disable automatic updates and remove some unnecessary packages
+  - apt-get remove -y update-manager gnome-initial-setup
+  # We don't need Firefox for the tests, so we can remove it to save some space
+  - snap remove firefox --purge
```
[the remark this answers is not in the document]
denisonbarbosa: Not necessary, IMO. The cached image is already small enough. If we need more space in the future, we could that.

e2e-tests/authd-msentraid/common/browser_login.py
denisonbarbosa: I think it could, but the code spawns a subprocess on itself, so I think it's good to have a "in your face" approach with it, wdyt?

e2e-tests/authd-msentraid/common/browser_login.py
denisonbarbosa: I thought about it, but it's not worth adding the dependency for something as simple as this, especially for a script that is not intended to be run by people.

e2e-tests/run_tests.sh
```diff
@@ -45,36 +49,27 @@ for test_file in $TESTS_TO_RUN; do
...
 
-    echo "Spawning VM with snapshot: ${VM_IMG}"
-    ${VM_DIR}/spawn-vm.sh ${VM_IMG} &
-    sleep 5  # Wait a bit for the VM to boot up
-    VM_PID=$(pidof kvm)
-    echo "VM PID: ${VM_PID}"
+    virsh snapshot-revert e2e-runner "${SNAPSHOT_NAME}" || true
 
+    # Temporarily allow a command that could error out so we can grab the test result.
+    set +e
```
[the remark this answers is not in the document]
denisonbarbosa: We can, I just missed it :sweat_smile: I'll update it.

e2e-tests/run_tests.sh
```diff
@@ -0,0 +1,88 @@
...
+        kill -KILL "${VM_PID}"
+    fi
+
+    sleep 5  # Wait for the VM to stop
+    if [ ${test_result} -ne 0 ]; then
+        test_results+=("${test_name}: FAILED")
+    else
+        test_results+=("${test_name}: OK")
+    fi
+    rm "${test_name}"
```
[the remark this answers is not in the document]
denisonbarbosa: Ohhhh, now I understand what you meant... I'll update it

e2e-tests/testing.md
```diff
@@ -0,0 +1,386 @@
...
+    ```
+
+### 2. Setup the VM
+
+The tests need a VM to run. This can be easily setup by using the domain definition and cloud-init configuration provided in the repository.
+
+1. Download the latest Ubuntu Desktop image (and resize it):
+
+    ```bash
+    wget https://cloud-images.ubuntu.com/questing/current/questing-server-cloudimg-amd64.img -O {img_path}
```
adombeck: We can also omit the `-O {img_path}` part to make it easier to just copy/paste and run. By default, `wget` writes the output into a file of the same as the basename of the URL in the current directory, so `questing-server-cloudimg-amd64.img` in this case.

e2e-tests/testing.md
```diff
@@ -0,0 +1,386 @@
...
+
+    qemu-img resize {img_path} 10G
+    ```
+
+2. Create the cloud-init iso using the provided configuration in `e2e-tests/vm/runner-cloud-cfg.yaml`:
+   1. Make sure to update the YAML file and change the `E2E_VM_PASSWORD` placeholder with the desired password.
```
adombeck: can we just hardcode the password to make the setup steps simpler? I would use `ubuntu` as the hardcoded password.

e2e-tests/testing.md
```diff
@@ -0,0 +1,386 @@
...
+
+        3. Create a snapshot of this state, so we can revert to it later when we proceed to configure the brokers.
+
+            ```bash
+            virsh snapshot-create-as e2e-runner --name authd-edge-installed --reuse-external
+            ```
+
+        4. Repeat the above steps for the stable version of authd, just changing the PPA to `ppa:ubuntu-enterprise-desktop/authd` and naming the snapshot `authd-stable-installed`.
+
+        - Script to automate the above steps for both stable and edge versions of authd:
```
adombeck: Would be useful to mention above that there's a script here. I followed the instructions step by step, so at this point I already have the snapshot for the edge version, so I can't run the script as is

denisonbarbosa: The script is mentioned at line 135, but maybe it wasn't evident enough?

e2e-tests/vm/runner-cloud-cfg.yaml
```diff
@@ -0,0 +1,66 @@
...
+  expire: false
+  users:
+    - {name: ubuntu, password: E2E_VM_PASSWORD, type: text}
+
+package_update: true
+package_upgrade: true
+packages:
+  - build-essential
+  - openssh-server
+  - ubuntu-desktop-minimal
```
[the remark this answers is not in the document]
denisonbarbosa: I switched to it because Marco suggested, but it doesn't seem to support cloud-init.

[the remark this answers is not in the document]
adombeck: cloud-init did run for me, I was able to see its output with `virsh console e2e-runner`

e2e-tests/vm/runner-cloud-cfg.yaml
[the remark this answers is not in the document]
3v1n0: Check it out, since these are quite big eh:

```
-rw------- 2 root root 517M set 27 02:27 /var/lib/snapd/snaps/gnome-42-2204_226.snap
-rw------- 2 root root 619M set  9 10:23 /var/lib/snapd/snaps/gnome-46-2404_125.snap
```

e2e-tests/authd-msentraid/common/browser_login.py
3v1n0: FYI `argparse` is part of the standard library

e2e-tests/common/utils.resource
3v1n0: As you prefer, the `/us/bin/` drop still stands :)

e2e-tests/authd-msentraid/common/broker.resource
```diff
@@ -0,0 +1,298 @@
+*** Settings ***
```
adombeck: What's the purpose of the `common` directory in `authd-msentraid`? AFAICT nothing else uses the files in that directory. Can we remove that directory and move its contents to `authd-msentraid/`?

e2e-tests/authd-msentraid/common/broker.resource
```diff
@@ -0,0 +1,298 @@
...
+Documentation       MS Entra ID specific resources for the tests
+
+Resource            kvm.resource
+Resource            ${AUTHD_COMMON}/utils.resource
+
+Library            Process
+
+*** Variables ***
+${AUTHD_COMMON}    ${CURDIR}/../authd-common
+${ENTRAID_COMMON}    ${CURDIR}/../broker-common
```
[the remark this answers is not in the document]
adombeck: Ah, I now see that you're creating symlinks these as symlinks in `run_tests.sh`.

e2e-tests/resources/authd-msentraid/broker.resource
```diff
@@ -0,0 +1,298 @@
+*** Settings ***
```
denisonbarbosa: I reorganized the assets in the latest commit. What do you think?

e2e-tests/run_tests.sh
```diff
@@ -0,0 +1,80 @@
...
+fi
+
+# Run the YARF tests
+test_results=()
+for test_file in $TESTS_TO_RUN; do
+    test_name=$(basename "${test_file}")
+
+    ln -s "${test_file}" .
+    # Ensure the test file is removed on exit
+    trap "rm -f ${test_name}" EXIT
```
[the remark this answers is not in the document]
adombeck: Right. I don't like that such hacks are necessary to work around YARF limitations. RobotFramework supports running single files just by specifying the path to the file instead of a directory.

e2e-tests/resources/authd-msentraid/broker.resource
```diff
@@ -0,0 +1,305 @@
...
+    Match Text    https://microsoft.com/devicelogin    15
+
+
+Continue Log In With Remote User: Log In On Browser
+    [Arguments]    ${username}
+    # Wait until the verification URL and login code are displayed
+    Match Text    https://microsoft.com/devicelogin
+    # Read the user code.
+    ${text} =    Read Text
+    ${user_code} =    StringUtils.FirstMatch    (https://)?microsoft.com/devicelogin\n((Login code: )?([A-Z0-9]+))    ${text}
```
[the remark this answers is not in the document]
adombeck: my behave test prototype supported that :smile: 

https://github.com/ubuntu/authd/blob/5050484228da906aaabba6f04a24597274117d62/behave-tests/steps/steps.py#L339-L343

e2e-tests/tests/password_change.robot
```diff
@@ -0,0 +1,37 @@
...
+
+
+Log in with remote user with device authentication
+    Open Terminal
+    Log In With Remote User Through CLI: QR Code    ${username}    ${local_password}
+    Log Out From Terminal Session
+    Close Focused Window
+
+
+Change Local Password Of Remote User
```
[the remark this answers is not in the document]
adombeck: Ok, I'm also fine with title case, let's just be consistent. We're currently mixing the two styles.

e2e-tests/testing.md
```diff
@@ -0,0 +1,408 @@
+# End-to-end tests
+
+## Explanation
+
+The end-to-end tests are implemented using [YARF](https://github.com/canonical/yarf). They cover a wide range of scenarios,
+both for authd and the brokers.
+
+## Running the tests locally
+
+Running the tests locally requires a bit of setting up. This is a step-by-step guide to get you started.
```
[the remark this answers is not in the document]
adombeck: Ok, but let's please point to the script at the top, so that users see it before following the manual steps

e2e-tests/resources/authd/Command.py
```diff
@@ -0,0 +1,13 @@
+from robot.api.deco import keyword, library  # type: ignore
+from robot.libraries.Process import Process
+from robot.output import LOGGER
+
+@library
+class Command:
+    @keyword
+    async def run(self, command: str, *arguments):
+        result = Process().run_process(command, *arguments)
```
3v1n0: This is awaiting for us?

e2e-tests/vm/provision.sh
```diff
@@ -0,0 +1,264 @@
...
+    qemu-kvm \
+    qemu-system-x86 \
+    qemu-utils \
+    python3-cairo \
+    python3-gi \
+    python3-tk \
+    socat \
+    sshpass \
+    wget \
+    xvfb
```
3v1n0: I would use:

```bash
dependencies=(
    bsdutils
    clang
    cloud-image-utils
    ffmpeg
    gir1.2-webkit2-4.1
    libcairo2-dev
    libgirepository-2.0-dev
    libvirt0
    libvirt-clients
    libvirt-clients-qemu
    libvirt-daemon
    libvirt-daemon-driver-qemu
    libvirt-daemon-system
    libxkbcommon-dev
    qemu-kvm
    qemu-system-x86
    qemu-utils
    python3-cairo
    python3-gi
    python3-tk
    socat
    sshpass
    wget
    xvfb
)

sudo apt-get install -y "${dependencies[@]}"

Comment (3v1n0) [reply e2e-tests/vm/provision.sh]: BTW as per some previous discussion, some things like `python3-tk` are not needed at tests runtime, or `lib*-dev`, so can mark them as `auto` and then auto-remove them

File: e2e-tests/vm/provision.sh [hunk 45]
```diff
@@ -0,0 +1,264 @@
...
+
+# Download the image
+mkdir -p "${ARTIFACTS_DIR}"
+IMAGE_URL="https://cloud-images.ubuntu.com/questing/current/questing-server-cloudimg-amd64.img"
```
3v1n0: Use daily maybe

e2e-tests/setup_yarf.sh
```diff
@@ -0,0 +1,33 @@
...
+
+YARF_REPO_URL="https://github.com/canonical/yarf"
+
+SCRIPT_DIR=$(dirname "$(readlink -f "$0")")
+YARF_DIR="${SCRIPT_DIR}/.yarf"
+
+# Clone the YARF repository if it doesn't exist
+if [ ! -d "$YARF_DIR" ]; then
+    echo "Cloning YARF repository into $YARF_DIR..."
+    git clone "$YARF_REPO_URL" "$YARF_DIR"
```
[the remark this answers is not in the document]
adombeck: done

e2e-tests/vm/provision.sh
```diff
@@ -0,0 +1,264 @@
...
+
+        # Install broker, configure and restart services
+        $SSH "sudo snap install ${broker} --channel=${channel} && \
+              sudo mkdir -p /etc/authd/brokers.d && \
+              sudo cp /snap/${broker}/current/conf/authd/${broker_config} /etc/authd/brokers.d/ && \
+              sudo sed -i -e 's|<ISSUER_ID>|'${ISSUER_ID}'|g' \
+                          -e 's|<CLIENT_ID>|'${CLIENT_ID}'|g' \
+                          /var/snap/${broker}/current/broker.conf && \
+              sudo systemctl restart authd.service && \
+              sudo snap restart ${broker}"
```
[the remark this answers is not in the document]
adombeck: I went with the first proposal, I don't see why it would be better to not expand the variables directly.

I also added `-euo pipefail` to the bash command and used `<<-EOF` to be able to indent the commands we pass to bash.

adombeck:
> Given the amount of shell stuff we're adding I'd suggest to also include here a shellcheck action (https://github.com/marketplace/actions/shellcheck maybe?)

I added a job, lets see if it passes

e2e-tests/vm/provision.sh
3v1n0: Well, I understand but we do already have bashism everywhere, so I'd rather explicitly target that (and we can do it with shell check).

I don't expect bash to go anywhere in the next century at least, so if we can use cleaner code it's better IMHO.

And leading backslashes are a thing that is always easy to forget :)

e2e-tests/setup-yarf.sh
```diff
@@ -0,0 +1,33 @@
...
+    git pull
+fi
+
+# Install uv snap if not already installed
+if ! command -v uv &> /dev/null; then
+    echo "Installing uv snap..."
+    sudo snap install --classic astral-uv
+else
+    echo "uv snap already installed"
+fi
```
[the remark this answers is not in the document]
adombeck: there is no uv anymore after `uv sync`. I suspect that command replaces the `.venv` we created before

e2e-tests/vm/provision.sh
```diff
@@ -112,15 +112,18 @@ function install_brokers() {
...
+			sudo mkdir -p /etc/authd/brokers.d
+			sudo cp /snap/${broker}/current/conf/authd/${broker_config} /etc/authd/brokers.d/
+			sudo sed -i \
+		  		-e "s|<ISSUER_ID>|${ISSUER_ID}|g" \
+		  		-e "s|<CLIENT_ID>|${CLIENT_ID}|g" \
+		  		-e "s|#ssh_allowed_suffixes_first_auth =|ssh_allowed_suffixes_first_auth = ${AUTHD_USER}|g" \
+		  		/var/snap/${broker}/current/broker.conf
+			sudo systemctl restart authd.service
+			sudo snap restart "${broker}"
+		EOF
```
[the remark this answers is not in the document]
3v1n0: Well, it's just how scripts are written normally for verbatim stuff.

But really not a big deal, as you like it more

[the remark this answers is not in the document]
3v1n0: Well, it's just how scripts are written normally for verbatim stuff.

But really not a big deal, as you like it more

e2e-tests/resources/authd/utils.resource
```diff
@@ -0,0 +1,126 @@
...
+    Match Text    Show Apps    300
+    Update System Time
+
+
+Log Out
+    Run Command    gnome-session-quit --logout --no-prompt
+
+
+Update System Time
+    Run Command In Terminal    sudo service chronyd restart && sudo chronyc waitsync
```
adombeck: Is chrony even installed? It's not installed by default on Ubuntu AFAICT, and I don't see it being installed explicitly, so if it installed I wonder which package pulls it in as a dependency.

Can't we just use systemd-timesyncd?

```suggestion
    Run Command In Terminal    sudo systemctl restart systemd-timesyncd.service
```

adombeck: Ah ok, I didn't know that. That seems like the right approach then. Let's just use the systemd executable instead of the sys V init one to restart the service:

```suggestion
    Run Command In Terminal    sudo systemctl restart chronyd && sudo chronyc waitsync
```

e2e-tests/testing.md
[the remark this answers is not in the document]
3v1n0: Ok, right fair enough!

e2e-tests/resources/authd-msentraid/browser_login.py
```diff
@@ -0,0 +1,105 @@
...
+    browser.send_key_taps([Gdk.KEY_Return])
+
+    browser.wait_for_text_visible("You have signed in")
+    browser.wait_for_stable_page()
+    screenshot_window(browser, os.path.join(screenshot_dir, "05-device-login-success.png"))
+
+
+def screenshot_window(window: Gtk.Window, filename: str):
+    # Get widget allocation (size)
+    alloc = window.get_allocation()
```
3v1n0: Comment here is unneeded, it's more than clear.

However, maybe we can assert that window.is_drawable() and window.get_mapped()

e2e-tests/resources/authd/browser_window.py
```diff
@@ -0,0 +1,193 @@
...
+
+        overlay.destroy()
+        self.web_view.grab_focus()
+
+    def wait_for_text_visible(self, text, timeout_ms=5000,
+                              poll_interval_ms=100):
+        """Wait until `text` is present in the page's visible text or raise TimeoutError."""
+        loop = GLib.MainLoop()
+        poll_id = None
+        timeout_id = None
```
3v1n0: They're int values

```suggestion
        poll_id = 0
        timeout_id = 0
```

e2e-tests/resources/authd/browser_window.py
```diff
@@ -0,0 +1,193 @@
...
+                loop.quit()
+
+        def poll_fn():
+            # Use json.dumps to safely escape the text into a JS string literal
+            js = (
+                     "Boolean(document.body && document.body.innerText && "
+                     "document.body.innerText.indexOf(%s) !== -1)"
+                 ) % json.dumps(text)
+            self.web_view.run_javascript(js, None, on_js_finished, None)
+            return True  # keep polling until callback quits the loop
```
3v1n0: Instead of doing this here, I'd return `False` here and instead restart the polling (calling `poll_id = GLib.timeout_add(poll_interval_ms, poll_fn)` after the finish function has returned, so that there's no possibility that we're polling twice while the same async function has not returned yet.

e2e-tests/tests/block_login_with_different_username.robot
```diff
@@ -0,0 +1,28 @@
...
+Log in with local user
+    Log In
+
+
+Fail to login if usernames do not match
+    Open Terminal
+    Start Log In With Remote User Through CLI: QR Code   different_user
+    Select Provider
+    Continue Log In With Remote User: Log In On External Browser   ${username}
+    Check That User Information Can Not Be Fetched
```
adombeck: This step fails because the error message that's displayed is `Authentication failure: requested username "different_user" does not match the authenticated username`, not `authentication failure: could not fetch user info` (I think I changed that in a recent PR). 

I'll a push a commit to fix that.

e2e-tests/resources/authd-msentraid/broker.resource
```diff
@@ -0,0 +1,296 @@
...
+*** Keywords ***
+Enable Edge Broker
+    Open Terminal In Sudo Mode
+    Run Command In Terminal    snap refresh authd-msentraid --edge
+    Close Terminal In Sudo Mode
+
+
+Disable Broker And Purge Config
+    Open Terminal In Sudo Mode
+    Run Command In Terminal    snap stop authd-msentraid
```
adombeck: This step fails if I increase the OCR confidence threshold because it's expecting `'@ubuntu:~$'` but the string in the root terminal is `root@ubuntu: /home/ubuntu#`.

e2e-tests/resources/authd-msentraid/browser_login.py
```diff
@@ -0,0 +1,106 @@
...
+SNAPSHOT_INDEX = 0
+
+
+def main():
+    parser = argparse.ArgumentParser()
+
+    parser.add_argument("username")
+    parser.add_argument("password")
+    parser.add_argument("device_code")
+    parser.add_argument("totp_secret")
```
3v1n0: This is not the `totp_secret` but rather the `totp_code`, no?

Also can be optional or we require it all the times? In such case just call it `--totp-code`, and adapt the code depending whether is None or not.

adombeck: No, it's the base32 encoded secret key

e2e-tests/tests/block_login_with_different_username.robot
adombeck: done

e2e-tests/resources/authd-msentraid/broker.resource
adombeck: I fixed that

e2e-tests/tests/config_owner_auto_update.robot
```diff
@@ -0,0 +1,30 @@
...
+Try to log in with remote user
+    Open Terminal
+    Log In With Remote User Through CLI: QR Code    ${username}    ${local_password}
+    Log Out From Terminal Session
+    Close Focused Window
+
+
+Check that owner was updated in broker configuration
+    Open Terminal In Sudo Mode
+    Check Configuration Value    owner    %{username}
```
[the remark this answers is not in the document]
adombeck: I fixed that

e2e-tests/resources/authd-msentraid/broker.resource
```diff
@@ -0,0 +1,296 @@
...
+# Uses sed to change the broker configuration.
+# It should match both commented and uncommented lines.
+# The full commands looks like:
+#     sed -i 's/#*${config_key} = .*/${config_key} = ${config_value}/g' ${ENTRAID_BROKER_CFG}
+Change Broker Configuration
+    [Arguments]    ${config_key}    ${config_value}
+    Open Terminal In Sudo Mode
+    Hid.Type String    sed -i 's/#*${config_key} = .*/${config_key} = ${config_value}/g' ${ENTRAID_BROKER_CFG}
+    Hid.Keys Combo    Return
+    Run Command In Terminal    snap restart authd-msentraid
```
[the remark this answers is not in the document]
adombeck: I fixed that

e2e-tests/resources/authd-msentraid/browser_login.py
```diff
@@ -0,0 +1,106 @@
...
+        browser.destroy()
+
+
+def login(browser, username: str, password: str, device_code: str, totp_secret: str, screenshot_dir: str = "."):
+    browser.web_view.load_uri("https://microsoft.com/devicelogin")
+    browser.wait_for_stable_page()
+    browser.capture_snapshot(screenshot_dir, "page-loaded")
+
+    browser.wait_for_text_visible("Enter code to allow access")
+    browser.wait_for_stable_page()
```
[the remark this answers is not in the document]
adombeck: > In fact we should first wait for the page to stabilize then to check its contents.

I'm reverting that because I'm seeing it causing infinite hangs of the tests. Until we add a timeout to `wait_for_stable_page`, we should not call it before `wait_for_text_visible` (which has a timeout).

Arguably, calling `wait_for_stable_page` after `wait_for_text_visible` is not of much use, so we could also not call it at all.

e2e-tests/run-tests.sh
```diff
@@ -0,0 +1,157 @@
...
+if [ -z "${E2E_USER:-}" ] || [ -z "${E2E_PASSWORD:-}" ] || [ -z "${BROKER:-}" ]; then
+    echo "Error: E2E_USER, E2E_PASSWORD, and BROKER must be set either as environment variables or via command line arguments."
+    usage
+    exit 1
+fi
+
+if [ -z "${TESTS_TO_RUN}" ]; then
+    echo "Running all tests in ${TESTS_DIR}"
+    TESTS_TO_RUN=$(find "${TESTS_DIR}" -type f -name "*.robot")
+fi
```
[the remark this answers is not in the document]
denisonbarbosa: I'm not reverting the snapshot works... Won't YARF lose the VNC connection?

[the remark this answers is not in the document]
adombeck: There's no active VNC connection during the Suite Setup step. Later VNC connections work fine because the VNC port doesn't change.

e2e-tests/resources/authd/utils.resource
```diff
@@ -0,0 +1,143 @@
...
+    Hid.Keys Combo    Return
+    Match Text    root@ubuntu    30
+    Hid.Type String    exit
+    Hid.Keys Combo    Return
+    Close Focused Window
+
+
+Log Out From Terminal Session
+    Match Text    @ubuntu    30
+    Hid.Type String    exit
```
adombeck: This doesn't have any effect without a subsequent `Hid.Keys Combo    Return`. It doesn't seem to be needed though, the `]]]` is enough to exit the machinectl login session, so I would just remove it.

adombeck: done

adombeck (approved):
All the end to end tests pass on my system now.

The many new commits I pushed will need another review, but from my side we're good to merge now.

adombeck:
@denisonbarbosa reviewed my changes and approved them (except for a commit which used `robot` instead of `yarf` - I dropped that for now). Time to merge!

PR #673, opened 2025-09-10
Reduce device access token polling interval

adombeck:
The default interval is 5 seconds, which means the user has to wait up to 5 seconds after successful authentication. That's a quite long time to wait. Let's improve UX by reducing the interval to 1 second.

This becomes even more important with device registration, which itself takes about 5 seconds, adding to the time the user has to wait.

UDENG-7907

### DIFF
--- a/internal/broker/broker.go
+++ b/internal/broker/broker.go
@@ -227,7 +227,6 @@ func (b *Broker) GetAuthenticationModes(sessionID string, supportedUILayouts []m
 
 	for _, mode := range availableModes {
 		if !slices.Contains(modesSupportedByUI, mode) {
-			log.Debugf(context.Background(), "Authentication mode %q is not supported by the UI", mode)
 			continue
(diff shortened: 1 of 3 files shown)


internal/broker/broker.go
```diff
@@ -540,11 +541,18 @@ func (b *Broker) handleIsAuthenticated(ctx context.Context, session *session, au
 		}
 		expiryCtx, cancel := context.WithDeadline(ctx, response.Expiry)
 		defer cancel()
+
+		// The default interval is 5 seconds, which means the user has to wait up to 5 seconds after
+		// successful authentication. We're reducing the interval to 1 second to improve UX a bit.
+		response.Interval = 1
```
denisonbarbosa, line 547: This is quite a big reduction in the polling time. Did you test it manually to see if it doesn't break anything (particularly in the UI)?

PR #687, opened 2025-09-20
Add `vanilla` broker resource implementation for existing e2e-tests

BarcoMasile:
## Description
This PR builds on top of https://github.com/ubuntu/authd-oidc-brokers/pull/659.

This adds the broker resource implementation for the e2e tests present.
It implements the login behavior in the `vanilla/common/BrowserWindow.py` file, with the device flow.

An important thing to note is that the python script for the vanilla broker does two additional things compared to the original one used as a template from the msentraid folder:
- it adds the implementation (leveraging only the standard library) of the TOTP generation, so a new parameter is needed, which is the secret key for the TOTP
- it adds an assertion at the end of the sign-in process, since there was no assertion initially, so the script didn't change exit value based on the actual success of the login flow: now it does and it communicates success/failure through 0/1 exit codes

### DIFF
--- a/e2e-tests/resources/authd-google/browser_login.py
+++ b/e2e-tests/resources/authd-google/browser_login.py
@@ -94,7 +94,7 @@ def login(browser, uri: str, username: str, password: str, device_code: str, tot
     browser.send_key_taps(
         ascii_string_to_key_events(password) + [Gdk.KEY_Return])
 
-    browser.wait_for_pattern("Verify your identity")
+    browser.wait_for_pattern("Verify your identity", timeout_ms=20000)
     browser.wait_for_stable_page()
     browser.capture_snapshot(screenshot_dir, "device-login-enter-totp-code")
     enter_totp_code(browser, totp_secret)


e2e-tests/vanilla/common/BrowserWindow.py
```diff
@@ -0,0 +1,265 @@
...
+        return h
+
+
+def ascii_string_to_key_events(string):
+    if len(string) != len(string.encode()):
+        raise TypeError(f"{string} is not an ascii string")
+    return [ord(ch) for ch in list(string)]
+
+
+def timeout_handler(signum, frame):
```
3v1n0, line 174: Where is this used?

3v1n0: ah, it was there already... Maybe I had some idea for using it... But not actually done :)

e2e-tests/vanilla/common/BrowserWindow.py
```diff
@@ -0,0 +1,265 @@
...
+    raise TimeoutError("Operation timed out")
+
+
+def wait_for_successful_text(webview, selector="h1.p-heading--4", retries=10, delay=1.0):
+    js = f'''
```
3v1n0, line 179: maybe this can be part of the "generic" class and this can be another parameter

BarcoMasile: Yes it could be, even though from what I'm told this implementation is ok to keep implementation speicific (relatively to the OIDC broker target of this e2e test, which is the Canonical Identity Platform)

e2e-tests/vanilla/common/BrowserWindow.py
```diff
@@ -0,0 +1,265 @@
...
+import gi  # type: ignore
+
+gi.require_version("Gtk", "3.0")
+gi.require_version("Gdk", "3.0")
+gi.require_version("WebKit2", "4.1")
+
+from gi.repository import Gdk, Gtk, GLib, WebKit2 as WebKit  # type: ignore
+
+
+class BrowserWindow(Gtk.Window):
```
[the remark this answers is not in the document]
BarcoMasile: Nice, I'll coordinate with Denison for this

e2e-tests/vanilla/common/BrowserWindow.py
```diff
@@ -0,0 +1,261 @@
...
+            result_holder["text"] = js_result.to_string()
+
+        webview.evaluate_javascript(js, -1, None, None, None, on_js_finished, None)
+
+        while Gtk.events_pending():
+            Gtk.main_iteration()
+
+        if result_holder["text"]:
+            return result_holder["text"]
+
```
3v1n0, line 196: This is not something we should do, as it may break many things.

Instead the code should be looking something like:

```python
    js_result = None
    loop = GLib.MainLoop()

    def on_js_finished(webview, result):
        nonlocal js_result

        try:
            js_result = webview.evaluate_javascript_finish(result)
        except Exception as e: # TODO: filter exception...
            print(f"{e}")
            # We want to raise it too?
            # raise e
        loop.quit()

    webview.evaluate_javascript(js, -1, None, None, None, on_js_finished, None)
    loop.run()

    if js_result and js_result.to_string():
        return js_result.to_string()

    wait_for_timeout(delay)
```

Using: https://github.com/ubuntu/authd-oidc-brokers/pull/659#discussion_r2368294909

e2e-tests/vanilla/common/BrowserWindow.py
```diff
@@ -0,0 +1,261 @@
...
+        time.sleep(delay)
+
+    return None
+
+
```
3v1n0, line 201: Nope, no time sleep in UI components :)

See also https://github.com/ubuntu/authd-oidc-brokers/pull/659#discussion_r2368294909

e2e-tests/vanilla/common/BrowserWindow.py
```diff
@@ -0,0 +1,265 @@
...
+    def generate_totp(self) -> str:
+        intervals_number = int(time.time()) // 30
+        token = str(self.__get_hotp_token(self.otp_secret, intervals_no=intervals_number))
+        # Pad token up to 6 digits
+        while len(token) != 6:
+            token += '0'
+
+        return token
+
+    def __get_hotp_token(self, secret, intervals_no):
```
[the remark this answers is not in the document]
3v1n0: Well, we already have dependency on non-standard library already since we rely on `python3-gi` and some gir files, so I feel it's not a big deal one more.

e2e-tests/vanilla/common/BrowserWindow.py
3v1n0: If you use the code above should be good enough, if not we've other problems... :)

e2e-tests/vanilla/common/BrowserWindow.py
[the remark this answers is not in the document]
BarcoMasile: I've evaluated both those libraries when I went with this implementation
It is very simple as you can see, so IMO it would be good either way.
But I'll sync with Denis on this so we're aligned, thanks

PR #697, opened 2025-10-03
Make the broker service D-Bus activated

adombeck:
This has the effect that both after installation and during boot, the broker service is not started automatically but only the first time a user tries to authenticate via PAM.

For now there are a few small benefits:
* Currently, the service always fails to start after installation, because the user didn't have a chance to replace the placeholders in the configuration file yet. That results in errors in the system log, which is a bit unclean.
* In the case that the system is useful after boot without a user logging in (so usually when it's a server), we avoid using up resources for a service that's not needed - at least until the first user tries to log in.
* In case a broker was installed but authd was not configured to use it, that broker is not started anymore.

Having the system D-Bus activated will become even more useful if we at some point make it exit on idle, because then we can free up its resources (primarily the memory) after some time.

Closes ubuntu/authd#1101

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -11,12 +11,13 @@ license: GPL-3.0
 apps:
   authd-oidc:
     command: bin/authd-oidc
-    daemon: simple
+    daemon: dbus
     slots:
       - dbus-authd
     plugs:
       - network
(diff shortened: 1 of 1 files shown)


3v1n0 (approved):
Thanks! 👍

PR #712, opened 2025-10-22
Add workflow to provision and cache VM for E2E tests

denisonbarbosa:
This PR adds workflows to provision and cache an Ubuntu VM to be later used in the E2E tests.

UDENG-7426

### DIFF
--- a/.github/workflows/provision-e2e-vm.yml
+++ b/.github/workflows/provision-e2e-vm.yml
@@ -277,7 +277,7 @@ jobs:
           --user "${{ secrets.E2E_GOOGLE_USERNAME }}" \
           --password "${{ secrets.E2E_GOOGLE_PASSWORD }}" \
           --release "${{ matrix.release }}" \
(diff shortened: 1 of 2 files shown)


.github/workflows/provision-e2e-vm.yml
```diff
@@ -0,0 +1,102 @@
...
+          sudo virsh attach-disk --domain e2e-runner --source /tmp/vm/seed.iso --target sda --type cdrom --config
+
+          # Start the VM
+          sudo virsh start e2e-runner
+
+          sleep 300s
+
+          for i in {1..20}; do
+            domain_state=$(sudo virsh domstate e2e-runner || true)
+            if [[ -z "$domain_state" || "$domain_state" == "shut off" ]]; then
```
3v1n0: Not a big deal, but I think we can avoid `sudo` at all in `virsh` commands once you add the user to `libvirt` (and maybe `newgrp` to it)

PR #764, opened 2025-12-02
Set up E2E tests for authd-google

denisonbarbosa:
So far, we had only authd-msentraid covered. This PR adds the necessary modifications to also support running the tests for authd-google.

UDENG-8532

### DIFF
--- a/.shellcheckrc
+++ b/.shellcheckrc
@@ -1,0 +1,2 @@
+source-path=SCRIPTDIR
+external-sources=true


e2e-tests/vm/provision-authd.sh
```diff
@@ -0,0 +1,200 @@
...
+
+SCRIPT_DIR="$(dirname "$(readlink -f "$0")")"
+LIBVIRT_XML_TEMPLATE="${SCRIPT_DIR}/e2e-runner-template.xml"
+ARTIFACTS_DIR="${SCRIPT_DIR}/.artifacts/${RELEASE}"
+SSH="${SCRIPT_DIR}/ssh.sh"
+
+while [[ $# -gt 0 ]]; do
+    case "$1" in
+        -k|--ssh-public-key)
+            SSH_PUBLIC_KEY_FILE="$2"
```
adombeck: `SSH_PUBLIC_KEY_FILE` is unused in this file

e2e-tests/vm/provision-authd.sh
```diff
@@ -0,0 +1,200 @@
...
+wait_for_system_running
+
+# Create a snapshot of the initial setup
+force_create_snapshot "initial-setup"
+
+# Install authd stable and create a snapshot
+retry --times 3 --delay 1 -- timeout 30 -- "$SSH" -- \
+  "sudo add-apt-repository -y ppa:ubuntu-enterprise-desktop/authd"
+
+sleep 30
```
adombeck: why this long sleep?

e2e-tests/vm/provision-ubuntu.sh
```diff
@@ -0,0 +1,249 @@
...
+    libvirt-clients-qemu \
+    libvirt-daemon \
+    libvirt-daemon-driver-qemu \
+    libvirt-daemon-system \
+    qemu-kvm \
+    qemu-system-x86 \
+    qemu-utils \
+    sshpass \
+    wget \
+    xvfb
```
adombeck: Why is this only a subset of the packages installed in the other two `provision-` scripts? And why do we install packages in all three provision scripts? Aren't they executed one after the other, which means it would be enough to install the packages once here (or in `provision.sh`)?

e2e-tests/vm/provision-broker.sh
```diff
@@ -0,0 +1,219 @@
...
+    libvirt-daemon-system \
+    libxkbcommon-dev \
+    qemu-kvm \
+    qemu-system-x86 \
+    qemu-utils \
+    python3-cairo \
+    python3-gi \
+    python3-tk \
+    socat \
+    xvfb
```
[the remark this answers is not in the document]
adombeck: lets continue this discussion in https://github.com/ubuntu/authd-oidc-brokers/pull/764/files#r2582696740

e2e-tests/vm/provision-ubuntu.sh
denisonbarbosa: My intention was to make the `provision-` scripts a bit independent from what happens in `provision.sh` (it's going to be useful for CI especially) so we kind of need to (try to) install the dependencies on each step.

